PR #3, opened 2025-11-19
[PM-20624] Fix and improve failing GitHub workflows

iangregsondev:
# Description

There are a number of workflows that are failing. These should be fixed.

# Proposed Solution

Need to check for missing secrets and try and improve any workflows

# Testing

There are no source code changes. The only changes are on the GitHub workflows and GitHub secrets.
All workflows should function as expected.

### DIFF
--- a/.github/workflows/e2e-tests-local.yml
+++ b/.github/workflows/e2e-tests-local.yml
@@ -12,6 +12,10 @@ env:
   ALLURE_ADMIN_PASSWORD: ${{ secrets.ALLURE_ADMIN_PASSWORD }}
   NETWORK: "undeployed"
   DEPLOYMENT: "local"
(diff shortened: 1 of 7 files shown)


.github/workflows/e2e-tests-local.yml
```diff
@@ -12,6 +12,10 @@ env:
   ALLURE_ADMIN_PASSWORD: ${{ secrets.ALLURE_ADMIN_PASSWORD }}
   NETWORK: "undeployed"
   DEPLOYMENT: "local"
+  APP_INFRA_STORAGE_PASSWORD: ${{ secrets.APP_INFRA_STORAGE_PASSWORD }}
```
agronmurtezi: what are those needed for?

PR #6, opened 2025-11-20
[PM-20642] Add sample .env file and update docs

iangregsondev:
# Description

We are now using a .env file in the wallet repo(ignored by git), a `.env.example` is needed to enable developers to create their own.

Some documentation is also needed.

Also, it seems that some of the env variables we were using are actually not needed on the standalone version of the indexer. See [midnightntwrk/midnight-indexer](https://github.com/midnightntwrk/midnight-indexer?tab=readme-ov-file#running)

# Testing

All tests should still pass in the CI.
Locally, developers should be able to run tests.


README.md
```diff
@@ -99,6 +106,27 @@ turbo format
...
+### Environment Setup
+
+Tests that require environment variables (such as those using Docker Compose for local infrastructure) need to be
+configured. The repository includes a `.env.example` file that serves as a template showing all available configuration
+options. To configure your environment:
+
+1. Copy `.env.example` to `.env`:
+
+   ```shell
+   cp .env.example .env
```
jsidorenko: I think we need to add the `/.env` file to .gitignore

PR #7, opened 2025-11-21
[PM-20657] Extract test seeds for QA workflows

iangregsondev:
# Description

QA Workflows were using test seeds, which need to be rotated and extracted to the CI secrets

The default values for test seeds in the [Run Workflow] inputs were hard-coded; they should be removed and depend on the new CI secrets as default values.

# Testing

Healthcheck workflows should continue to work as before, and E2ETests should also continue to function as before.

Running the E2ETests provides a list of possible values or ways to override the seeds; these inputs no longer contain hardcoded seeds as their defaults, but a message stating to “leave them blank to use the defaults” has been added, which essentially will use the secrets on the CI.


iangregsondev:
@jtsang586 I beleive some of these inputs are deprecated or not correct, can you provide a correct list - p.s. these appear in multiple places. If you provide me with a correct list then I can ensure that these are updated.

```
      network:
        type: choice
        description: Midnight network to use
        required: true
        default: "devnet"
        options:
          - devnet
          - testnet
          - undeployed
      deployment:
        type: choice
        description: Midnight deployment to use
        required: true
        default: "qanet"
        options:
          - local
          - qanet
          - testnet
```

Take a note of the `options` and also the `default` value.

iangregsondev:
@jtsang586 Seeds are now extracted to the CI. I still need to get these physically added at the GitHub level. I have prefixed these.

```
      E2E_TESTS_SEED:
        required: true
      E2E_TESTS_SEED2:
        required: true
      E2E_TESTS_NT_SEED:
        required: true
      E2E_TESTS_NT_SEED2:
        required: true
```

I found two additional secrets in the e2etests workflow files,

```
  SEED_STABLE: ${{ secrets.SEED_STABLE }}
  HALO2_ADDRESSES: ${{ vars.HALO2_ADDRESSES }}
```

I have a feeling that we are NOT using HALO2_ADDRESSES, can you confirm ? Can we remove it?

And SEED_STABLE, is that specific to e2e tests?

Should we prefix this with E2E_TESTS_ ? Or some other prefix?

Thanks

iangregsondev:
> > SEED vs SEED2 and NT_SEED vs NT_SEED2
> > Although functionally it won't make a difference, might it be better to have more semantically correct names?
> 
> Yes these can be a bit clearer specifically NT_SEEDs are for native token tests...can we maybe leave it for now? I think we can possibly remove these actually but I need to spend some to run these tests.

Sure, i won't touch them at the moment.

.github/workflows/healthcheck-testnet.yml
jtsang586: I think we can remove the testnet.yml as those tests will run from the legacy branch and testnet will be decommissioned soon anyway.

PR #49, opened 2026-01-08
[PM-20889] Allow tx creation with no own outputs

iangregsondev:
# Description

Shielded wallet should allow creation of transactions with no own outputs (e.g. when spending the only shielded NFT)

# Testing

All existing tests should continue to pass


agronmurtezi:
@jtsang586 nice, merging this one as I'm doing a release

PR #78, opened 2026-01-22
Enable lint rule for type imports, so that type imports do not fail a…

kapke:
…t runtime when run with type stripping

# Description

On a different PR, I did run into an issue in a test, where node.js was running TS code with type stripping. It was a type import not marked with `type` keyword. This PR enables a lint rule, which forces to use type imports for type-only imports


packages/capabilities/src/balancer/Balancer.ts
```diff
@@ -10,8 +10,8 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-import { CounterOffer, TransactionCostModel } from './CounterOffer.js';
-import { CoinRecipe, Imbalance, Imbalances, TokenType, TokenValue } from './Imbalances.js';
+import { CounterOffer, type TransactionCostModel } from './CounterOffer.js';
+import { type CoinRecipe, type Imbalance, type Imbalances, type TokenType, type TokenValue } from './Imbalances.js';
```
jsidorenko: Can't this be rewritten as:
```suggestion
import type { CoinRecipe, Imbalance, Imbalances, TokenType, TokenValue } from './Imbalances.js';
```
?

iangregsondev:
Nice addition!

iangregsondev (dismissed):

packages/capabilities/src/balancer/Balancer.ts
kapke: It could, due to the number of files affected I was leaning on automated "fix lint" flow instead of doing the change manually

iangregsondev (dismissed):

agronmurtezi (dismissed):

agronmurtezi (approved):

PR #289, opened 2026-04-07
feat: add @midnight-ntwrk/wallet-sdk barrel package

kapke:
## Summary
- Introduces `@midnight-ntwrk/wallet-sdk`, a barrel package that re-exports all wallet SDK packages through a single dependency
- Provides 4 entry points: `.` (core wallets + hd + address-format), `./capabilities`, `./proving`, `./testing`
- Migrates `test-website` and `docs-snippets` from 6-7 individual `@midnight-ntwrk/wallet-sdk-*` dependencies to the single barrel package

## Test plan
- [x] `yarn dist --force` builds all packages successfully
- [x] `yarn typecheck --force` passes (32/32)
- [x] `yarn lint --force` passes (50/50)
- [x] `yarn test --force` passes (indexer-client failures are pre-existing infra tests requiring `APP_INFRA_SECRET`)


packages/wallet-sdk/package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "@midnight-ntwrk/wallet-sdk",
+  "description": "Barrel package re-exporting all Midnight Wallet SDK packages",
+  "version": "1.0.0-beta.0",
```
agronmurtezi: we should drop the beta version here and let changesets manage it as well

iangregsondev:
Out of interest, could we not now adopt this in the facade, so instead of the facade importing from many different packages it could now just use this barrel package ? @kapke

iangregsondev (dismissed):

kapke:
I considered this, but:
- `-facade` package is harder to find
- facade itself does not rely on address format and hd wallet, which need to be gathered too
- it would mix re-exports and actual code in the facade

I believe approach present in the PR is easier to maintain and allows for better separation where needed.

jsidorenko (dismissed):

packages/wallet-sdk/package.json
```diff
@@ -0,0 +1,84 @@
+{
+  "name": "@midnight-ntwrk/wallet-sdk",
+  "description": "Barrel package re-exporting all Midnight Wallet SDK packages",
+  "version": "0.0.0",
```
agronmurtezi, line 4: why `0.0.0`?

kapke: To let changesets tackle the versioning

agronmurtezi (approved):

PR #291, opened 2026-04-08
[PM-22421] Update release versions to major for some components for breaking changes; fix race condition in shielded wallet tx history

iangregsondev:
# Description

Some versions of the components in Wallet need a major version bump because of identified breaking changes and another has been identified as a minor bump.

Also, in the shielded txHistory there is a chance of a race condition when adding items to the history using the same “key”. 

# Proposed Solution

CleanShot 2026-04-08 at 05 51 14@2x

It is recommended to do the following :

Fix transaction history race condition by consolidating merge logic in the facade and delegating it to storage at construction time.

# Testing

NO QA Required.

Add new tests to test the race condition.


packages/abstractions/src/InMemoryTransactionHistoryStorage.ts
```diff
@@ -20,57 +20,67 @@ import {
...
-    const existing = this.entries.get(entry.hash);
-    this.entries.set(entry.hash, existing ? { ...existing, ...entry } : entry);
+    const existing = this.storage.get(entry.hash);
+    this.storage.set(entry.hash, existing ? this.merge(existing, entry) : entry);
     return Promise.resolve();
   }
 
   async *getAll(): AsyncIterableIterator<T> {
-    for (const entry of this.entries.values()) {
+    for (const entry of this.storage.values()) {
```
kapke: nitpick: `Array.fromAsync`? 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync

packages/e2e-tests/src/tests/utils.ts
```diff
@@ -441,6 +441,33 @@ export const waitForRegisteredTokens = (wallet: WalletFacade) =>
...
+      rx.mergeMap(async (state) => {
+        const txEntry = await wallet.queryTxHistoryByHash(txHash);
+        return { state, txEntry };
+      }),
+      rx.tap(({ txEntry }) => {
+        logger.info(
+          `Waiting for tx ${txHash} in history, found: ${txEntry !== undefined}, ready: ${txEntry !== undefined && isReady(txEntry)}`,
+        );
+      }),
+      rx.filter(({ state, txEntry }) => state.isSynced && txEntry !== undefined && isReady(txEntry)),
```
kapke: I think it's better to filter for synced state at the first step.
In such case the need for threading it further disappears entirely.

kapke (dismissed):

kapke:
There should be 2 tests at least:
1. showing that upserting arbitrary transaction history entries respects the merge function
2. showing that the merge function for wallet entries does not lose information

packages/abstractions/src/InMemoryTransactionHistoryStorage.ts
iangregsondev: Done

packages/e2e-tests/src/tests/utils.ts
iangregsondev: Agreed, thanks.

packages/facade/test/transactionHistory.test.ts
```diff
@@ -0,0 +1,206 @@
...
+});
+
+const baseEntry = (hash: string, overrides: Partial<WalletEntry> = {}): WalletEntry => ({
+  hash,
+  protocolVersion: 1,
+  status: 'SUCCESS',
+  ...overrides,
+});
+
+describe('InMemoryTransactionHistoryStorage upsert respects the merge function', () => {
```
kapke: this test should live in the `abstractions` package, correct?

packages/facade/test/transactionHistory.test.ts
```diff
@@ -0,0 +1,206 @@
...
+      shielded: { receivedCoins: [coin], spentCoins: [] },
+    });
+
+    const incoming = baseEntry('tx1', {
+      shielded: { receivedCoins: [coin], spentCoins: [] },
+    });
+
+    const merged = mergeWalletEntries(existing, incoming);
+
+    expect(merged.shielded?.receivedCoins).toHaveLength(1);
```
kapke: why checking length if the check is followed by `toEqual`?

packages/facade/test/transactionHistory.test.ts
iangregsondev: Yes, moved and created some additional const to suport the mergeEntries..

kapke (approved):

PR #293, opened 2026-04-08
fix(dust-wallet): fix fee sign bug and segment_id collision in balancer

adamreynolds-io:
## Summary

Fixes two bugs in `packages/dust-wallet/src/v1/Transacting.ts` that block consecutive dApp connector transactions.

### Bug 1: Fee sign convention mismatch in `computeBalancingRecipe` (infinite loop)

`getBalanceRecipe` uses a sign convention where negative = deficit (select coins) and positive = surplus (create change). The fee convergence loop at [L496](https://github.com/midnightntwrk/midnight-wallet/blob/main/packages/dust-wallet/src/v1/Transacting.ts#L496) stores `currentFee: newFee` where `newFee` from `dryRunFee` is always positive. On the next iteration at [L472](https://github.com/midnightntwrk/midnight-wallet/blob/main/packages/dust-wallet/src/v1/Transacting.ts#L472), this positive value is passed to `CapImbalances.fromEntry('dust', currentFee)` which interprets it as surplus — so `getBalanceRecipe` selects 0 coins. The loop repeats forever: `{currentFee: 1, newFee: 1, coverage: 0, inputs: 0}`.

This triggers when `initialFees = 0` (transaction's dust imbalance is already balanced). Contract call transactions from the dApp connector hit this; deploy transactions have `initialFees = -1` and converge in one iteration before the sign matters. The loop has no iteration limit, so it runs until the WASM module stack-overflows (~250s).

**Fix:** negate positive `currentFee` before passing to `getBalanceRecipe`.

### Bug 2: Segment ID collision in fee transaction creation

`dryRunFee` ([L380](https://github.com/midnightntwrk/midnight-wallet/blob/main/packages/dust-wallet/src/v1/Transacting.ts#L380)) and `balanceTransactions` ([L524](https://github.com/midnightntwrk/midnight-wallet/blob/main/packages/dust-wallet/src/v1/Transacting.ts#L524)) create fee transactions via `Transaction.fromPartsRandomized()` which assigns a random segment_id in [2, 65534]. These are merged with input transactions. If the random segment_id matches an input transaction's intent segment_id, `Transaction.merge()` throws `IntentSegmentIdCollision`. With the infinite loop from bug 1, enough iterations make collision inevitable.

**Fix:** collect input transaction segment_ids, retry `fromPartsRandomized` until non-colliding.

### Why this doesn't reproduce in Node.js

Transactions created internally by the SDK (`facade.transferTransaction()`, test helpers, CLI wallet) consistently produce `initialFees = -1`, so the loop converges on the first iteration before the positive `newFee` feeds back. The dApp connector path receives externally-constructed transactions (from `@midnight-ntwrk/midnight-js-contracts`) that produce `initialFees = 0`, triggering the sign mismatch.

## Test plan

- [x] New test: `balanceTransactions converges when transaction has zero dust imbalance` (5s timeout — fails fast if loop regresses)
- [x] New test: `balanceTransactions avoids segment_id collision with input transactions` (40 randomized txs, verifies merge succeeds)
- [ ] Existing dust-wallet tests pass
- [ ] Validated end-to-end: two consecutive mints via GSD Wallet dApp connector succeed

### DIFF
--- a/packages/dust-wallet/src/v1/Transacting.ts
+++ b/packages/dust-wallet/src/v1/Transacting.ts
@@ -469,7 +469,7 @@ export class TransactingCapabilityImplementation<TTransaction extends AnyTransac
                     value: coin.value,
                     token: coin.token,
                   })),
-                  initialImbalances: CapImbalances.fromEntry('dust', currentFee),
+                  initialImbalances: CapImbalances.fromEntry('dust', currentFee <= 0n ? currentFee : -currentFee),
                   feeTokenType: 'dust',
                   transactionCostModel: {
                     inputFeeOverhead: 0n,
--- a/packages/dust-wallet/src/v1/Transacting.ts
+++ b/packages/dust-wallet/src/v1/Transacting.ts
@@ -23,6 +23,7 @@ import {
   SignatureVerifyingKey,
   Transaction,
   UnshieldedOffer,
+  UnprovenIntent,
   UtxoOutput,
   UtxoSpend,
   FinalizedTransaction,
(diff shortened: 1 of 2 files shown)


adamreynolds-io:
## Code Review Summary

Ran a parallel 3-agent review (correctness, style/FP conventions, error handling).

### Findings by severity

| Severity | Count | Fixed | Dismissed |
|----------|-------|-------|-----------|
| P1 | 0 | — | — |
| P2 | 2 | 2 | 0 |
| P3 | 4 | 2 | 2 |
| P4 | 4 | 0 | — |

### P2 fixes applied

1. **Test flake in collision test** — Merging 40 `fromPartsRandomized` txs had ~1.2% birthday-problem collision rate between *input* txs (unrelated to the fix under test). Replaced with a direct assertion that the balancing tx's segment IDs don't overlap with input segment IDs.

2. **`let` + `for` loop in collision check** — Replaced `let collision` + inner `for` loop with `[...txIntents.keys()].some(id => usedSegmentIds.has(id))`.

### P3 fixes applied

3. **`collectIntentSegmentIds` rewritten functionally** — Imperative `for` loops + `Set.add()` mutation → `new Set(transactions.flatMap(...))` one-liner.

4. **Improved error message** — `createNonCollidingFeeTx` failure now includes occupied ID count.

### P3 dismissed

5. **`throw` vs typed `Either` in `createNonCollidingFeeTx`** — Both call sites catch thrown errors (`Effect.try` wraps as `OtherWalletError`, `LedgerOps.ledgerTry` wraps as `LedgerError`). The condition is probabilistically unreachable (~(0.0006)^100). Adding `Either` return type would complicate callers without meaningful benefit. Consistent with existing imperative patterns in the same file (lines 358-370).

6. **Fee sign expression simplification** — `currentFee <= 0n ? currentFee : -currentFee` is equivalent to `-abs(currentFee)` for bigints. Could extract a helper, but the expression is clear enough inline and only used once.

### Quality pipeline

- Build: passing
- Tests: 14/14 passing
- Typecheck: passing
- Formatting: passing
- Effect diagnostics: 0 errors, 0 warnings

All fixes committed as `8c8bac1` and pushed.

agronmurtezi:
closing as it's addressed on #334

PR #307, opened 2026-04-11
test(dust-wallet): external transaction shapes and fee sign bug repro

adamreynolds-io:
Tests for the dust balancer's fee convergence loop with transaction structures from external dApps (via midnight-js / dApp connector). Existing tests only cover wallet-originated transactions with `guaranteedUnshieldedOffer`.

### Bug under test

A wallet-sdk bug in `computeBalancingRecipe` (`packages/dust-wallet/src/v1/Transacting.ts`). The convergence loop passes `currentFee` (always positive from `dryRunFee`) directly to `getBalanceRecipe`, which interprets positive values as surplus and selects 0 coins. The loop never converges. The WASM behaves correctly — the stack overflow is a symptom of the infinite loop, not the cause. Fix is in #293: negate positive `currentFee` before passing to `getBalanceRecipe`.

### Tests added

- **`getBalanceRecipe` sign convention** — positive imbalance selects 0 coins (surplus), negative selects coins (deficit)
- **Fallible-only transaction** — no guaranteed content, `fromPartsRandomized`
- **Mixed guaranteed + fallible** — both sections present
- **`feeImbalance` for bare intent** — fee always appears as dust deficit regardless of content
- **Serialization round-trip** — `serialize` → `deserialize` → `balanceTransactions`
- **Fee sign bug repro** — forces `initialFees = 0` via `calculateFee` stub; triggers infinite convergence loop on unfixed code (WASM stack overflow)

### Expected results

All tests pass except the sign bug repro, which **fails on `main`** (infinite loop). Apply the fix from #293 to make it pass.


agronmurtezi:
closing as it's addressed on #334

PR #334, opened 2026-04-20
Deterministically set segment ID during balancing

agronmurtezi:
Replaces `Transaction.fromPartsRandomized` with deterministic segment placement when building balancing transactions in both the **dust-wallet** and **unshielded-wallet** capabilities. Previously the balancing intent was assigned to a random fallible segment id, which made transaction layouts non-reproducible and harder to debug. This PR picks the lowest unused fallible segment id (range `[1, 65535]`) that does not collide with any intents in the transactions the balancer is working against.


iangregsondev (approved):

agronmurtezi:
closes #337

PR #365, opened 2026-04-28
feat: integrate pending tx service with txHistory

### LINKED ISSUE #364: Save tx identifiers in txHistory on submit, clear on sync confirmation
## Describe the proposed feature:

When a transaction is submitted, a pending entry is added to the txHistory storage keyed by transaction identifier. When sync later confirms the transaction, the pending entry is cleared and the confirmed entry is written via the existing sync flow.

iangregsondev:
# Description

When a transaction is submitted, a pending entry is added to the txHistory storage keyed by transaction identifier. When sync later confirms the transaction, the pending entry is cleared and the confirmed entry is written via the existing sync flow.

# Testing

All existing and new tests should pass


packages/abstractions/src/InMemoryTransactionHistoryStorage.ts
```diff
@@ -30,13 +31,16 @@ import {
 export class InMemoryTransactionHistoryStorage<
   T extends { hash: TransactionHash } = TransactionHistoryCommon,
   I = T,
-> implements TransactionHistoryStorage<T> {
+  P extends { hash: TransactionHash; identifiers: readonly string[] } = PendingTransactionHistoryCommon,
+> implements TransactionHistoryStorage<T, P> {
   private storage: Map<TransactionHash, T>;
+  private pendingStorage: Map<TransactionHash, P>;
```
agronmurtezi: why not use the `status` field in the storage instead? The flow would be something like this:

tx submitted -> add entry to tx history storage with status 'pending' -> once the tx is resolved -> update the status of the entry

iangregsondev: This has been addressed in the latest PR update.

packages/abstractions/src/test/inMemoryTransactionHistoryStorage.test.ts
```diff
@@ -11,13 +11,22 @@
...
 
-const baseEntry = (hash: string, overrides: Partial<TransactionHistoryCommon> = {}): TransactionHistoryCommon => ({
+const finalizedInput = (
+  hash: string,
+  overrides: Partial<FinalizedEntryInput<FinalizedTransactionHistoryCommon>> = {},
+): FinalizedEntryInput<FinalizedTransactionHistoryCommon> => ({
   hash,
   protocolVersion: 1,
   status: 'SUCCESS',
+  finalizedAt: new Date(0),
```
kapke: If anything, it should be a block hash or simplified block information, including its hash.

packages/abstractions/src/TransactionHistoryStorage.ts
```diff
@@ -40,13 +89,78 @@ export type SerializedTransactionHistory = string;
...
+ * `TransactionHistoryWriter<Wider>` is assignable to `TransactionHistoryWriter<Narrower>`.
+ */
+export interface TransactionHistoryWriter<
+  T extends FinalizedTransactionHistoryCommon = FinalizedTransactionHistoryCommon,
+> {
+  /**
+   * Record that a tx has been submitted and is awaiting confirmation. Keyed by the tx's first identifier (the only
+   * stable id available before the tx appears on-chain).
+   */
+  gotPending(hash: TransactionHash, identifiers: readonly string[], submittedAt: Date): Promise<void>;
```
kapke: Why not taking `T` instance here?

packages/facade/src/index.ts
```diff
@@ -59,34 +59,59 @@ import {
...
 } from '@midnight-ntwrk/wallet-sdk-address-format';
 
-/**
- * Full entry schema for transaction history — common fields + all wallet sections. Pass this to
- * `InMemoryTransactionHistoryStorage` to enable serialize/restore.
- */
-export const WalletEntrySchema = Schema.Struct({
-  ...TransactionHistoryStorage.TransactionHistoryCommonSchema.fields,
+/** Schema for a finalized wallet entry — common finalized fields + wallet sections. */
+export const FinalizedWalletEntrySchema = Schema.Struct({
```
kapke: Why not exposing a function from the capability to create a schema with the lifecycle given an extension schema? 
That is:
```
export const ExtendedSchema = <T>(extensionSchema: Schema.Schema<T>): Schema.Schema<T & Base> => Schema.Intersection(base, extensionSchema)
```

packages/facade/src/index.ts
```diff
@@ -945,7 +990,22 @@ export class WalletFacade {
       this.shielded.revertTransaction(tx),
       this.unshielded.revertTransaction(tx),
       this.dust.revertTransaction(tx),
```
kapke, line 1359: why not reporting rejection here?

packages/facade/src/index.ts
```diff
@@ -945,7 +990,22 @@ export class WalletFacade {
...
-    ]).then(() => this.pendingTransactionsService.clear(tx as unknown as ledger.FinalizedTransaction));
+    ]).then(async () => {
+      await this.pendingTransactionsService.clear(tx as unknown as ledger.FinalizedTransaction);
+      // Mirror submitTransaction's keying: prefer the locally-computed tx hash, fall back to firstIdentifier. This
+      // ensures gotRejected hits the same key the original gotPending used, so the lifecycle transition is in-place.
+      const identifiers = tx.identifiers();
```
kapke: Those are things that should not be a concern here. This code should just call proper lifecycle method with the tx it has at hand and call it a day

packages/facade/src/index.ts
```diff
@@ -59,34 +59,59 @@ import {
...
 
+export type FinalizedWalletEntry = Schema.Schema.Type<typeof FinalizedWalletEntrySchema>;
+
+/**
+ * Full entry schema for transaction history — covers all three lifecycle variants. Pass this to
+ * `InMemoryTransactionHistoryStorage` to enable serialize/restore of pending, finalized and rejected entries together.
+ */
+export const WalletEntrySchema = Schema.Union(
+  FinalizedWalletEntrySchema,
+  TransactionHistoryStorage.PendingTransactionHistoryCommonSchema,
```
kapke: Why do the pending and rejected have different shape?

packages/facade/src/index.ts
```diff
@@ -945,7 +990,22 @@ export class WalletFacade {
...
+      const identifiers = tx.identifiers();
+      const rejectedHash = ((): TransactionHistoryStorage.TransactionHash | undefined => {
+        try {
+          return (tx as ledger.FinalizedTransaction).transactionHash().toString();
+        } catch {
+          return identifiers[0];
+        }
+      })();
+      if (rejectedHash !== undefined) {
+        await this.#txHistoryStorage.gotRejected(rejectedHash, identifiers, this.clock.now());
```
kapke: Is it really rejection here?

kapke (changes requested):

packages/facade/src/index.ts
iangregsondev: This has been changed and made this a concern of the lifecycle method

packages/facade/src/index.ts
iangregsondev: @kapke Pending vs Rejected are different based upon the discriminator of lifecycle, can you give me more context on what you want to see happening here ?

```
PendingTransactionHistoryCommonSchema  = { hash, identifiers, lifecycle: {status:'pending',  submittedAt} }
RejectedTransactionHistoryCommonSchema = { hash, identifiers, lifecycle: {status:'rejected', rejectedAt, reason?} **}**
```

```
export const PendingLifecycleSchema = Schema.Struct({
  status: Schema.Literal('pending'),
  submittedAt: Schema.Date,
});

export const RejectedLifecycleSchema = Schema.Struct({
  status: Schema.Literal('rejected'),
  rejectedAt: Schema.Date,
  reason: Schema.optional(Schema.String),
});
```


SO they only differ in the lifecycle as per above

```
export const PendingTransactionHistoryCommonSchema = Schema.Struct({
  hash: TransactionHashSchema,
  identifiers: Schema.Array(Schema.String),
  lifecycle: PendingLifecycleSchema,   <<<<<<<<< CHANGE HERE 
});

export const RejectedTransactionHistoryCommonSchema = Schema.Struct({
  hash: TransactionHashSchema,
  identifiers: Schema.Array(Schema.String),
  lifecycle: RejectedLifecycleSchema,   <<<<<<<< CHANGE HERE
});
```

packages/facade/src/index.ts
iangregsondev: I think this is now addressed, the logic has moved into the lifecycle method

@kapke Can you confirm that is what you were referring to?

Or give me a little more context. Thanks

packages/facade/src/index.ts
iangregsondev: As far as I can see I am reporting the rejection here with lifecycle method gotRejected, can you explain what you are referring to ?

```
    await Promise.all([
      this.shielded.revertTransaction(tx),
      this.unshielded.revertTransaction(tx),
      this.dust.revertTransaction(tx),
    ]).then(async () => {
      await this.pendingTransactionsService.clear(tx as unknown as ledger.FinalizedTransaction);
      await this.#txHistoryStorage.gotRejected(tx, this.clock.now());
    });
```

packages/abstractions/src/test/inMemoryTransactionHistoryStorage.test.ts
iangregsondev: @kapke Are you refereing to just the test or the signature ?

kapke: Signature

packages/facade/src/index.ts
kapke: NVM. I thought I saw a slightly different code.

packages/facade/src/index.ts
iangregsondev: Done

packages/abstractions/src/test/inMemoryTransactionHistoryStorage.test.ts
iangregsondev: Done

packages/abstractions/src/InMemoryTransactionHistoryStorage.ts
```diff
@@ -14,68 +14,120 @@ import { Schema } from 'effect';
...
+    const { submittedAt, ...rest } = input;
+    const entry = { ...rest, lifecycle: { status: 'pending', submittedAt } } as unknown as T;
+    await this.#upsert(entry);
+  }
+
+  async gotFinalized(input: FinalizedEntryInput<T>): Promise<void> {
+    const { finalizedBlock, ...rest } = input;
+    const entry = { ...rest, lifecycle: { status: 'finalized', finalizedBlock } } as unknown as T;
+    await this.#upsert(entry);
+    this.#clearPendingByIdentifiers((rest as { identifiers?: readonly string[] }).identifiers ?? [], entry.hash);
```
kapke: why type casting?

packages/abstractions/src/InMemoryTransactionHistoryStorage.ts
```diff
@@ -14,68 +14,120 @@ import { Schema } from 'effect';
...
+  #upsert(entry: T): Promise<void> {
+    const existing = this.#storage.get(entry.hash);
+    this.#storage.set(entry.hash, existing ? this.#merge(existing, entry) : entry);
+    return Promise.resolve();
+  }
+
+  #clearPendingByIdentifiers(identifiers: readonly string[], newHash: TransactionHash): void {
+    if (identifiers.length === 0) return;
+    const provided = new Set<string>(identifiers);
+    const match = [...this.#storage.values()].find((entry) => {
```
kapke: IIRC there's a `some` method which returns boolean

packages/abstractions/src/InMemoryTransactionHistoryStorage.ts
```diff
@@ -14,68 +14,120 @@ import { Schema } from 'effect';
...
+      const ids = (entry as { identifiers?: readonly string[] }).identifiers;
+      const lifecycle = (entry as { lifecycle?: { status?: string } }).lifecycle;
+      return (
+        Array.isArray(ids) &&
+        ids.length > 0 &&
+        ids.every((id: string) => provided.has(id)) &&
+        lifecycle?.status === 'pending' &&
```
kapke: why only 'pending'?

packages/facade/src/index.ts
```diff
@@ -1002,13 +1101,15 @@ export class WalletFacade {
...
     const raw = await this.#txHistoryStorage.get(hash);
-    return raw && isWalletEntry(raw) ? raw : undefined;
+    return raw && isFinalizedWalletEntry(raw) ? raw : undefined;
   }
 
   async getAllFromTxHistory(): Promise<WalletEntry[]> {
-    const allEntries = await this.#txHistoryStorage.getAll();
-    return [...allEntries];
+    const all = await this.#txHistoryStorage.getAll();
+    return all.filter((entry): entry is WalletEntry => isFinalizedWalletEntry(entry) || isPendingWalletEntry(entry));
```
kapke: why this filter?

packages/facade/src/index.ts
```diff
@@ -1002,13 +1101,15 @@ export class WalletFacade {
     ]);
   }
 
-  async queryTxHistoryByHash(hash: TransactionHistoryStorage.TransactionHash): Promise<WalletEntry | undefined> {
+  async queryTxHistoryByHash(
+    hash: TransactionHistoryStorage.TransactionHash,
+  ): Promise<FinalizedWalletEntry | undefined> {
     const raw = await this.#txHistoryStorage.get(hash);
-    return raw && isWalletEntry(raw) ? raw : undefined;
+    return raw && isFinalizedWalletEntry(raw) ? raw : undefined;
```
kapke: why checking if is finalized?

packages/facade/src/index.ts
```diff
@@ -60,33 +64,115 @@ import {
...
+ */
+const revertTxHistoryKey = (
+  tx: AnyTransaction,
+): { readonly hash: string; readonly identifiers: readonly string[] } | undefined => {
+  const identifiers = tx.identifiers();
+  if (identifiers.length === 0) return undefined;
+  try {
+    return { hash: tx.transactionHash().toString(), identifiers };
+  } catch {
+    return { hash: identifiers[0], identifiers };
```
kapke: First identifier is not a hash.

packages/abstractions/src/InMemoryTransactionHistoryStorage.ts
iangregsondev: Yes, correct, this was not necessary. Updated and removed.

packages/abstractions/src/InMemoryTransactionHistoryStorage.ts
iangregsondev: Correct there is, but `some` as you suggested gives me a boolean, I still need the return value because I am using the hash

```
this.#storage.delete(match.hash);
```

@kapke

packages/abstractions/src/InMemoryTransactionHistoryStorage.ts
iangregsondev: We are using txHistory to store the pending as well as finalized/rejected etc... 

So finalized/rejected are valid items in the txHistory - they are not in process - they are end states..

Pending is an in process state.

@kapke I would like to know if you have another idea here ?

packages/facade/src/index.ts
iangregsondev: You bring up an important fact here, yes, so there are 3 lifecycle states, pending, finalized and rejected.

Current code returns pending + finalized - maybe we just don't filter here at all, and the caller has the ability to filter using Lifecycle.Status.

I am going to remove these filters.

@kapke

packages/facade/src/index.ts
iangregsondev: I am now using a simple .serialize() whcih does not throw.

packages/facade/src/index.ts
iangregsondev: Yes, this has now been reverted, and it keeps the API compatible with main - returns WalletEntry

kapke (dismissed):

agronmurtezi (approved):

PR #374, opened 2026-04-29
Add dust balancing sign fix

JosephDenman:
PR #293 fixed two unrelated bugs in computeBalancingRecipe: a sign-convention mismatch that hangs when initialFees = 0, and random segment-id collisions between the balancing tx and the input txs. It got closed with "closing as it's addressed on #334" but #334 only carried over the segment-id fix (deterministic placement instead of random + retry). The sign fix never landed. main and wallet-sdk-dust-wallet@4.0.0 still hang on the original triggering case, so this PR re-applies just the missing piece.

### DIFF
--- a/.changeset/spicy-mangos-wish.md
+++ b/.changeset/spicy-mangos-wish.md
@@ -1,0 +1,6 @@
+---
+'@midnight-ntwrk/wallet-sdk-dust-wallet': patch
+---
+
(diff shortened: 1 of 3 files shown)


agronmurtezi:
closing as non issue, @JosephDenman in case you run into this again I'd be happy to take another look

PR #404, opened 2026-05-15
fix(shielded-wallet): handle empty transactions array from indexer to enable proper retry (#401)

### LINKED ISSUE #401: [Wallet] (shielded-wallet, dust-wallet): tx history silently drops shielded/dust sections when the indexer lags events
## Summary

`ShieldedWallet`'s sync path writes a tx-history entry only after a successful `getTransactionDetails(change.source)` call against the indexer's HTTP endpoint. When the wallet's WS `ZswapEvents` stream delivers a relevant event before the indexer's HTTP `transactions(...)` endpoint has indexed the same hash (a normal race on any deployed network), the lookup throws an unhandled `TypeError`, the configured retry schedule does not engage (because the throw is a `Cause.Die` defect, not a typed `Cause.Fail`), and the resulting error is swallowed by a `catchAllCause → Console.error`. The local change is never re-processed because `applyUpdate` has already advanced `appliedIndex`.

**Net effect:** the shielded section for that tx is permanently absent from `WalletFacade.queryTxHistoryByHash` / `getAllFromTxHistory`, while balances and coin sets still update correctly.

`dust-wallet` has the **identical** defect in its own `TransactionHistory.ts` / `RunningV1Variant.ts` (confirmed during #532 review) — dust sections are dropped under the same indexer-lag conditions. Both wallets are fixed together in #532.

## Impact

- Shielded **and dust** tx-history sections for in-flight tx submissions can be silently lost on any network with non-zero indexer ingest lag.
- e2e: `tokenTransfer.remote.test.ts` 'can perform a combined transaction' hangs in `waitForTxInHistory` because the predicate `(e) => e.shielded !== undefined && e.unshielded !== undefined` is unsatisfiable for the affected tx. (That predicate only checks `shielded`/`unshielded`, so it doesn't surface the analogous dust-section loss.)

## Observed behaviour

From a failing remote run on qanet:

```
Waiting for tx c0a46613b653a5c6f14a369f3799cdb122a57c3ec83a9f1358717fa8a2221204
  in history: found, status=SUCCESS, ready=false
…repeats every 500 ms…
```

Storage snapshot shows the entry exists but with `sections: "unshielded,dust"` only. (The logs below are from a shielded case; dust exhibits the analogous symptom by code inspection — the entry's `dust` section missing and the stack trace in `dust-wallet`.)

Earlier in the same run:

```
Error processing tx history metadata {
  _id: 'Cause',
  _tag: 'Die',
  defect: TypeError: Cannot read properties of undefined (reading '__typename')
      at packages/shielded-wallet/src/v1/TransactionHistory.ts:127:30
  …
}
```

The sender wallet's balances and available shielded coins update correctly — only the shielded slice of the history entry is missing.

## Root cause

`packages/shielded-wallet/src/v1/TransactionHistory.ts:120-147`

```ts
getTransactionDetails: (hash) =>
  Effect.gen(function* () {
    const statusQuery = yield* TransactionHistoryDetail;
    const result = yield* statusQuery({ transactionHash: hash });
    const tx = result.transactions[0];                      // (1) unchecked
    const rawStatus = tx.__typename === 'RegularTransaction' // (2) throws TypeError on undefined
      ? tx.transactionResult.status
      : undefined;
    …
  }).pipe(
    …,
    Effect.retry(Schedule.exponential(Duration.seconds(1)).pipe(Schedule.compose(Schedule.recurs(3)))),
    Effect.mapError((cause) => new TransactionHistoryError({ … })),
  ),
```

1. `result.transactions` is empty when the indexer hasn't yet ingested the tx → `tx` is `undefined`.
2. Reading `.__typename` throws synchronously inside `Effect.gen` → routed to `Cause.Die`.
3. `Effect.retry` on the schedule above retries typed failures only; defects bypass it. Zero retries occur.
4. `Effect.mapError` only maps typed failures, so the defect stays a defect.

The defect then surfaces at `packages/shielded-wallet/src/v1/RunningV1Variant.ts:148-163`:

```ts
Effect.flatMap(({ changes, protocolVersion }) =>
  pipe(
    Effect.forEach(
      changes,
      (change) =>
        pipe(
          this.#v1Context.transactionHistoryService.getTransactionDetails(change.source),
          Effect.flatMap((metadata) =>
            this.#v1Context.transactionHistoryService.put(change, metadata, protocolVersion),
          ),
          Effect.catchAllCause((cause) =>
            Console.error('Error processing tx history metadata', cause),  // swallowed, no telemetry
          ),
        ),
      { discard: true, concurrency: 'unbounded' },
    ),
    Effect.forkScoped,
  ),
)
```

`change` is not remembered after the catch. `applyUpdate` has already moved `appliedIndex` past the event that produced it (`Sync.ts:268-280`), so there is no later retry of `put` for that `change.source`.

`dust-wallet` has the identical structure at `packages/dust-wallet/src/v1/TransactionHistory.ts` / `RunningV1Variant.ts` (same `getTransactionDetails` empty-array dereference and same `catchAllCause → Console.error` swallow).

## Reproduction

1. Submit a combined shielded + unshielded transfer on a network where the indexer's HTTP ingest lags its WS event delivery (testnet, or local stack with the indexer's HTTP indexer paused for ~5 s after WS events fire).
2. Poll `wallet.queryTxHistoryByHash(txHash)`. Observe an entry exists with `unshielded` set, but `shielded` (and, under the same race, `dust`) is permanently `undefined`.
3. Stderr contains `Error processing tx history metadata` with `_tag: 'Die'` and the `__typename` TypeError.

A unit-level repro: inject a fake `TransactionHistoryDetail` that returns `{ transactions: [] }` once, then a populated response. The current code drops the entry on the first call; expected behaviour is to retry until populated (or to write the section without the metadata).

## Acceptance criteria

- Given an indexer that returns an empty `transactions` array on the first `getTransactionDetails` call for a hash and a populated response on a later call (within the retry window), the shielded/dust section MUST eventually be written to tx-history storage.
- Given a sustained indexer outage longer than the retry window, the failure MUST surface as a structured error (with `change.source`), not a silent `Console.error` defect.

(Both criteria apply to `shielded-wallet` and `dust-wallet`.)

BossChaos:
### Root Cause
`getTransactionDetails` accesses `result.transactions[0].__typename` without checking if the array is empty. When the indexer's WS event delivers before HTTP indexing completes, `transactions` is `[]` and `tx` is `undefined`. Accessing `undefined.__typename` throws a `TypeError` that routes to `Cause.Die`, bypassing the retry schedule entirely.

### Fix
Added explicit `undefined` check at line 126-129 that throws a regular `Error` when the indexer returns an empty transactions array. This converts the defect into a typed failure, which the existing `Effect.retry(Schedule.exponential(...).pipe(Schedule.compose(Schedule.recurs(3))))` can now catch and retry up to 3 times.

```ts
const tx = result.transactions[0];
if (tx === undefined) {
  // Indexer hasn't indexed this tx yet — retry will handle it
  throw new Error(`Transaction ${hash} not yet indexed`);
}
```

### Acceptance Criteria Met
- ✅ Empty `transactions` array now triggers retry (typed failure instead of defect)
- ✅ Retry schedule: exponential backoff × 3 attempts covers normal indexer ingest lag

### Testing
This fix enables the existing e2e test `tokenTransfer.remote.test.ts 'can perform a combined transaction'` to pass, as `waitForTxInHistory` will no longer encounter permanently-undefined `shielded` sections.

Wallet: 0xdaE5d307339074A24F579dB48e7c639359D94904


iangregsondev (approved):

iangregsondev (approved):

iangregsondev:
Thanks for the fix @BossChaos.

Heads up: the two failing checks (Build and check-changeset) are not caused by your change. They fail because GitHub doesn't pass repository secrets to workflows from forked PRs (the runner reports Secret source: None). 

There's nothing you can do from your fork to make these pass — it's on us. I have created a shadow PR to enable the running of the CI actions.

Here > https://github.com/midnightntwrk/midnight-wallet/pull/473

Once thing worth noting is that there is a missing changeset.

Could you create the missing changset (yarn changeset add) on your PR ?

Once everything passes then I can merge the new PR.

Does this work @BossChaos  ?

Thanks.

iangregsondev:
I am closing this PR, I have tested it again and it doesn't seem to address the original issue.

The codebase has also moved in, I will open up a new PR very shortly.

PR #416, opened 2026-05-19
fix(dust): book Night UTxOs at build time during registration

### LINKED ISSUE #302: [Wallet SDK] Dust registration does not book Night UTxOs
## Describe the proposed feature:
The Dust sub-wallet handles the entire registration flow internally, so the unshielded wallet never gets a chance to book the Night UTxOs used. This allows double-spending those UTxOs, with failures only surfacing at submission time.

## User Need:
The SDK must prevent double-use of Night UTxOs when a Dust registration transaction is created concurrently with a Night transfer using the same UTxOs.

## Expected Benefit:
Wallet fails fast at transaction creation (step 2) rather than silently allowing conflicting transactions to be submitted.

## Details:
**Scenario demonstrating the problem:**

1. Create a Dust registration transaction using some Night UTxOs
2. Create a Night transfer requiring overlapping Night UTxOs (before submitting the registration)
3. Submit both — only one fails at step 3 or later

**Expected:** Wallet should fail at step 2.

**Required process change:**

1. Dust wallet is presented with Night UTxOs and selects the one to use in the guaranteed section
2. Unshielded wallet creates a transaction intent moving the provided Night UTxOs (selected one in the guaranteed offer; rest in the fallible one)
3. Unshielded wallet books the coins used
4. Dust wallet adds a dust registration action to the intent
5. Standard signing/proving/binding/submission flow continues from that point

---
_Migrated from [PM-21646](https://shielded.atlassian.net/browse/PM-21646)_

[PM-21646]: https://shielded.atlassian.net/browse/PM-21646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

iangregsondev:
# Description

The Dust sub-wallet handles the entire registration flow internally, so the unshielded wallet never gets a chance to book the Night UTxOs used. This allows double-spending those UTxOs, with failures only surfacing at submission time.

# Proposed Solution

The SDK must prevent double-use of Night UTxOs when a Dust registration transaction is created concurrently with a Night transfer using the same UTxOs.

# Testing

All existing tests should pass as well as additional tests added.


packages/dust-wallet/src/v1/RunningV1Variant.ts
```diff
@@ -200,19 +196,56 @@ export class RunningV1Variant<TSerialized, TSyncUpdate, TTransaction, TStartAux>
...
+        resolvedTime,
+      );
+      return yield* this.#v1Context.transactingCapability
+        .createDustGenerationTransaction(resolvedTime, ttl, utxosWithDustValue, nightVerifyingKey, dustReceiverAddress)
+        .pipe(EitherOps.toEffect);
+    });
+  }
+
+  splitNightUtxosForDustAction(
+    currentTime: Date | undefined,
```
kapke: Why `| undefined`?

packages/dust-wallet/src/v1/RunningV1Variant.ts
```diff
@@ -200,19 +196,56 @@ export class RunningV1Variant<TSerialized, TSyncUpdate, TTransaction, TStartAux>
...
+        nightUtxos,
+        resolvedTime,
+      );
+      return yield* this.#v1Context.transactingCapability
+        .createDustGenerationTransaction(resolvedTime, ttl, utxosWithDustValue, nightVerifyingKey, dustReceiverAddress)
+        .pipe(EitherOps.toEffect);
+    });
+  }
+
+  splitNightUtxosForDustAction(
```
kapke: The split is only relevant for registration, correct?

packages/dust-wallet/src/v1/RunningV1Variant.ts
```diff
@@ -200,19 +196,56 @@ export class RunningV1Variant<TSerialized, TSyncUpdate, TTransaction, TStartAux>
...
+    currentTime: Date | undefined,
+    nightUtxos: ReadonlyArray<UtxoWithMeta>,
+    isRegistration: boolean,
+  ): Effect.Effect<NightUtxoSplitForDustActionWithCurrentTime, WalletError> {
+    if (nightUtxos.some((utxo) => utxo.type !== nativeToken().raw)) {
+      return Effect.fail(new OtherWalletError({ message: 'Token of a non-Night type received' }));
+    }
+    return Effect.gen(this, function* () {
+      const currentState = yield* SubscriptionRef.get(this.#context.stateRef);
+      const blockData = yield* this.#v1Context.syncService.blockData();
```
kapke: This implies a fetch. Do we need to make a remote request in order to learn time?

packages/facade/src/index.ts
```diff
@@ -458,28 +458,78 @@ export class WalletFacade {
...
+        registeredForDustGeneration: u.utxo.registeredForDustGeneration,
+      },
+    });
+    const guaranteedForUnshielded = split.guaranteedUtxos.map(toUnshieldedUtxoWithMeta);
+    const fallibleForUnshielded = split.fallibleUtxos.map(toUnshieldedUtxoWithMeta);
+
+    // Step 2 — Unshielded books the Night UTxOs (move available -> pending) and builds the intent
+    // with the two offers. After this point, a concurrent build call that wants any of these UTxOs
+    // will fail fast.
+    const txWithOffers = await this.unshielded.createDustActionBookingTransaction(
```
kapke: I think a name like "rotateUtxos" or sth along this lines would be better. I can imagine how sometimes one just wants to reorganize their utxos using this method.

packages/facade/src/index.ts
```diff
@@ -458,28 +458,78 @@ export class WalletFacade {
...
     }
 
-    const signatureData = intent.signatureData(1);
-    const signature = await Promise.resolve(signDustRegistration(signatureData));
-
-    return await this.dust.addDustGenerationSignature(transaction, signature);
+    try {
+      const signatureData = intent.signatureData(1);
+      const signature = await Promise.resolve(signDustRegistration(signatureData));
+      return await this.dust.addDustGenerationSignature(txWithDustActions, signature);
```
kapke: IMO this should now go via `signRecipe`

packages/dust-wallet/src/v1/RunningV1Variant.ts
iangregsondev: Yes, but also for deregistration.

packages/dust-wallet/src/v1/RunningV1Variant.ts
iangregsondev: I am able to cache the timestamp from the last sync. Would this suffice?

packages/dust-wallet/src/v1/RunningV1Variant.ts
iangregsondev: This gave us the opportunity to pass in a currentTime but if we did NOT , then we would use the blockTime (from sync service - see your other comment about fetch below).

Probably using the blockTime (or at least a cached version of it (see comment on fetch below)) is probably safer.,.

So we could just drop this parameter and always SOURCE it.

ideas?

kapke: Always require it. The caller has access to some notion of clock, so will provide the time.

packages/dust-wallet/src/v1/RunningV1Variant.ts
kapke: But `DustSpend` is also a Dust action, the split is not needed there. Hence - I still prefer to call it that it is for "registration". Especially that deregistration is registration, but with no address

packages/facade/src/index.ts
iangregsondev: This has been done

packages/dust-wallet/src/v1/RunningV1Variant.ts
iangregsondev: I have updated the name here to include ForRegistration which indicates more what its going to do 

splitNightUtxosForDustRegistration

packages/facade/src/index.ts
iangregsondev: I haver renamed this to 

bookNightUtxosForDustRegistration

Reasoning : Renaming to bookNightUtxosForDustRegistration to match the sibling splitNightUtxosForDustRegistration we just renamed

If oyu disagree, let me name and I will rename it to your specific name.

packages/dust-wallet/src/v1/RunningV1Variant.ts
iangregsondev: No longer relevant as we are now always going to pass the timestamp.

packages/dust-wallet/src/v1/RunningV1Variant.ts
iangregsondev: This has now been fixed, it uses the `this.clock....` which is used in other places also.

packages/facade/src/index.ts
kapke: The thing is - the purpose of this method is to:
1. spend an utxo at all to issue the registration
2. spend all the rest of utxos to trigger the generation

Hence - the method is suitable to work stand-alone, not only for the purpose of booking the coins.

iangregsondev: Agreed, renaming to rotateUtxos

packages/facade/src/index.ts
```diff
@@ -455,31 +474,82 @@ export class WalletFacade {
...
     }
 
-    const signatureData = intent.signatureData(1);
-    const signature = await Promise.resolve(signDustRegistration(signatureData));
-
-    return await this.dust.addDustGenerationSignature(transaction, signature);
+    // Step 4 — Sign via the standard signRecipe pathway, which now stamps both the unshielded
+    // offers and the dust registration. Signing failures also need to release the booking.
+    try {
+      const signedRecipe = await this.signRecipe(
```
kapke, line 522: When I asked about signing going through `signRecipe` I meant `should this method return a recipe now, so that signing could be performed in that unifrom way`?

packages/facade/src/index.ts
```diff
@@ -455,31 +474,82 @@ export class WalletFacade {
     action: { type: 'registration'; dustReceiverAddress: DustAddress } | { type: 'deregistration' },
     nightUtxos: readonly UtxoWithMeta[],
     nightVerifyingKey: ledger.SignatureVerifyingKey,
-    signDustRegistration: (payload: Uint8Array) => Promise<ledger.Signature> | ledger.Signature,
```
kapke, line 458: It needs to allow `Promise<ledger.Signature>`

packages/facade/src/index.ts
```diff
@@ -107,6 +107,25 @@ const TokenKindsToBalance = new (class {
   };
 })();
 
+// Memoize so signSegment runs at most once per unique payload. The same segment-1 signature can
+// need to land on both unshielded offers and a DustRegistration; without this, hardware-wallet
+// signers would prompt the user twice.
+const memoizeSignSegment = (
```
kapke: No need for this, it can be handled differently and should cover `signRecipe` too.

iangregsondev: Removed.

packages/facade/src/index.ts
iangregsondev: The promise needs to be removed until we change the public api.

kapke (approved):

PR #437, opened 2026-05-28
fix: support waiting for dust generation

### LINKED ISSUE #415: [Wallet] (dust-wallet): registerNightUtxosForDustGeneration submits an under-funded tx when called before the registered UTXOs have generated enough Dust        to cover their own fee
### Context & versions:

`WalletFacade.registerNightUtxosForDustGeneration(...)` constructs a `DustRegistration` whose `allow_fee_payment` is taken from a synthetic projection of "Dust that *would* have been generated so far by the unregistered Night UTXOs" (`coin.dust.generatedNow`, computed at the wallet's latest indexed block timestamp). It never compares that budget against the fee the registration tx will actually pay, and it never waits for that budget to be sufficient.

Immediately after a Night UTXO is received (so its `ctime` is one or two blocks ago), the projected `generatedNow` is on the order of 10¹⁴ stroke for a 1 NIGHT UTXO — right at the edge of the registration's own Dust fee. The chain rejects the submission with `MalformedTransaction::BalanceCheckOverspend` (`balancing_check` in the ledger) → RPC `1010: Invalid Transaction: Custom error: 138`.

On retry, several seconds of real wall-clock have elapsed, the same UTXO has back-credited materially more Dust, and the chain accepts the identical-shape transaction. That is why the healthcheck test flips between **fails-first-run / passes-on-retry** deterministically.

### Impact

- Any DApp/wallet caller that registers a Night UTXO promptly after receiving it can hit a silent under-funding failure surfaced only as a generic `SubmissionError`.
- E2E: `packages/e2e-tests/src/tests/dust.undeployed.test.ts:144` ("Able to register Night tokens for Dust generation after receiving unshielded tokens @healthcheck") races with block production and is intermittently red on the first run.
- The simulation suite already documents the same hazard and works around it with `simulator.fastForward(10_000n)` — see `packages/facade/test/simulation-mode.test.ts:124-127`. No equivalent exists for real networks.

### Observed behaviour

From a local undeployed run on `chore/disable-healthcheck-cron-schedules`:

```
RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus
  1010: Invalid Transaction: Custom error: 138

[13:27:32.942] ERROR (17479): Transaction submission error
[13:27:32.942] ERROR (17479): SubmissionError: Transaction submission error
    at packages/capabilities/src/submission/submissionService.ts:108:32
```

Wire payload (single intent, segment 1):

```
DustActions {
  spends: [],
  registrations: [
    DustRegistration {
      night_key: <verifying key>,
      dust_address: Some(DustPublicKey(ae63…d034)),
      allow_fee_payment: 148_806_000_000_000,   //  ≈ 1.49 × 10^14 stroke
      signature: Some(<sig>),
    },
  ],
  ctime: Timestamp(1779106704),
},
```

`spends: []` means the chain has only one source for the registration's Dust fee: back-credited Dust generated by the Night UTXOs being registered, capped by `allow_fee_payment`. At the block the tx lands in, the Dust the chain has actually accrued for those UTXOs is less than the fee, so `balance(...).bal < 0` for the Dust token in segment 1 → `BalanceCheckOverspend`.

### Root cause

`packages/dust-wallet/src/v1/Transacting.ts:231-238`

```ts
const feePayment = receiver
  ? pipe(
      splitResult.guaranteed,
      IterableOps.filter((coin) => !coin.utxo.registeredForDustGeneration),
      IterableOps.map((coin) => coin.dust.generatedNow),
      BigIntOps.sumAll,
    )
  : 0n;
```

`coin.dust.generatedNow` comes from `packages/dust-wallet/src/v1/CoinsAndBalances.ts:135-149` via a `fakeGenerationInfo` (`:169-176`):

```ts
const generatedValue = ledger.updatedValue(coin.ctime, coin.initialValue, genInfo, currentTime, parameters);
```

`currentTime` defaults to `blockData.timestamp` of the latest block the wallet's sync service has indexed — see `packages/dust-wallet/src/v1/RunningV1Variant.ts:206`:

```ts
Effect.let('currentTime', ({ blockData }): Date => currentTime ?? blockData.timestamp),
```

The facade always passes `undefined` for `currentTime` (`packages/facade/src/index.ts:462-463`), so this default is what's used in practice.

Failure modes in the wallet:

1. `feePayment` (the value sent as `allow_fee_payment`) is never compared against the tx's estimated Dust fee. The relevant fee estimator is in the same file (`Transacting.ts:370` — `transaction.feesWithMargin(ledgerParams, feeBlocksMargin)`); its result is not used to guard the registration.
2. The wallet silently submits a transaction the caller could not have known is doomed.
3. There is no public API to wait until `Σ generatedNow(unregistered Night UTXOs)` covers a given threshold, so callers cannot defer registration deterministically. Existing helpers (`waitForDustBalance`, `waitForRegisteredTokens` in `packages/e2e-tests/src/tests/helpers/stateWaiters.ts`) cover the post-registration balance, not the pre-registration budget.

### Reproduction

1. On any network with non-zero block time (local devnet is sufficient), fund receiver wallet A with shielded + Dust as usual.
2. From sender wallet, send wallet A a Night UTXO via an unshielded transfer.
3. As soon as wallet A reports the new UTXO (e.g. via `waitForUnshieldedCoinUpdate`), call `wallet.registerNightUtxosForDustGeneration(nightUtxos, …)`, finalize, submit.
4. Submission fails with `1010: Custom error: 138`. The wire-level `DustRegistration.allow_fee_payment` will be small (low-10¹⁴ stroke range for a 1 NIGHT UTXO).
5. Retry the same `registerNightUtxosForDustGeneration → submit` flow without changing inputs. After ~5 blocks of elapsed wall-clock, the identical tx-shape is accepted.

E2E reference: `packages/e2e-tests/src/tests/dust.undeployed.test.ts:99-130`. The current workaround (a single `waitForBlockAdvancement` on line 110) is not sufficient to clear the race.

### Acceptance criteria

1. **Fail fast with a typed error.** `WalletFacade.registerNightUtxosForDustGeneration(...)` MUST validate that the registration's `allow_fee_payment` covers the registration's own fee (with the configured `feeBlocksMargin`) *before* the tx leaves the wallet. When it does not, the call MUST reject with a typed error in the `WalletError` union (proposed: `InsufficientDustForFeeError`) carrying:
   - the shortfall in stroke,
   - the estimated number of seconds (or blocks) until generation crosses the threshold given current generation rate and `nightDustRatio`.
2. **Provide a wait helper.** A facade-level API MUST exist so callers can defer until the budget is sufficient. Proposed shape:
   ```ts
   wallet.waitForGeneratedDust(
     utxos: readonly UtxoWithMeta[],
     requiredStroke: bigint,
     opts?: { timeoutMs?: number },
   ): Promise<void>
   ```
   Implementation should reuse `estimateDustGeneration` and resolve on the next state tick where `Σ generatedNow(utxos) ≥ requiredStroke`.
3. **E2E coverage.** `dust.undeployed.test.ts` and `dust.remote.test.ts` healthcheck tests MUST use the new wait helper (and remove the single `waitForBlockAdvancement` on line 110 of the undeployed test). After this change, the registration test MUST pass on the first run with `retry: 0`.
4. **No regression to simulation tests.** `packages/facade/test/simulation-mode.test.ts` continues to pass; the `simulator.fastForward(10_000n)` step there documents the same hazard but uses a different time source and should not be affected.

iangregsondev:
# Description

registerNightUtxosForDustGeneration submits an under-funded tx when called before the registered UTXOs have generated enough Dust to cover their own fee.

# Proposed Solution

Support the ability to wait on the dust generation.

# Testing

All existing tests and new tests should pass.


packages/facade/src/index.ts
```diff
@@ -1034,6 +1051,23 @@ export class WalletFacade {
     };
   }
 
+  /** See `DustWalletAPI.waitForGeneratedDust`. */
```
kapke: Provide more documentation here, as it is the entrypoint for users

packages/dust-wallet/src/DustWallet.ts
```diff
@@ -397,6 +416,41 @@ export function CustomDustWallet<
...
+      requiredAmount: bigint,
+      opts?: { now?: () => Date; timeoutMs?: number },
+    ): Promise<void> {
+      if (nightUtxos.length === 0) {
+        throw Error('At least one Night UTXO is required.');
+      }
+      if (requiredAmount <= 0n) {
+        return;
+      }
+      const now = opts?.now ?? (() => new Date());
```
kapke: The source of current time can't be optional here.

packages/facade/src/index.ts
iangregsondev: Done

packages/dust-wallet/src/DustWallet.ts
iangregsondev: Removed it from opts and added a parameter.

packages/dust-wallet/src/DustWallet.ts
```diff
@@ -170,6 +170,34 @@ export type DustWalletAPI<TStartAux = DustSecretKey, TSerialized = string> = {
...
+   * `WalletFacade.estimateRegistration` to pick `requiredAmount` guarantees the subsequent
+   * `registerNightUtxosForDustGeneration` will pass its fee-coverage guard.
+   *
+   * @param nightUtxos - UTxOs to project generation for; same set passed to `registerNightUtxosForDustGeneration`.
+   *   Already-registered UTxOs are ignored. Must be non-empty.
+   * @param requiredAmount - Threshold to wait for, as a Dust amount. Resolves immediately if `<= 0n`.
+   * @param getCurrentTime - Source of current time, called on every tick. Required, and a function rather than a
+   *   snapshot `Date` like the other methods' `currentTime`: the projection only advances because this is re-read each
+   *   tick, and callers must inject their own clock so simulator-driven tests respect simulator time. Facade callers
+   *   pass `() => this.clock.now()`.
```
kapke: Why not make it compatible with clock, so that facade just provides `this.clock`?

iangregsondev: Yes,  I was able to do this and moved the clock into its own file in utils so that we are able to share it.

packages/dust-wallet/src/DustWallet.ts
```diff
@@ -23,6 +23,7 @@ import { type ProtocolState, ProtocolVersion, type SyncProgress } from '@midnigh
 import { type DustAddress } from '@midnight-ntwrk/wallet-sdk-address-format';
 import { type Runtime, WalletBuilder } from '@midnight-ntwrk/wallet-sdk-runtime';
 import { type Variant, type VariantBuilder, type WalletLike } from '@midnight-ntwrk/wallet-sdk-runtime/abstractions';
+import { type ClockOps } from '@midnight-ntwrk/wallet-sdk-utilities';
```
kapke: Why moving it to a shared package? There's nothing to share other than the types being welcome to match.

packages/utilities/src/Clock.ts
```diff
@@ -0,0 +1,23 @@
+// This file is part of MIDNIGHT-WALLET-SDK.
```
kapke, line 1: This file should be called just "Clock" and exported as "Clock". The `Ops` suffix  is added if the type is not owned by this package

iangregsondev: Agreed, renamed and exported as Clock

kapke (approved):

PR #442, opened 2026-06-02
fix: change stream implementation from async to async push

agronmurtezi:
# Description
 - Switch the WS subscription stream from `Stream.async` to `Stream.asyncPush`. At the indexer's sustained ~1k msg/sec push rate, `Stream.async` was forking a top-level fiber per `emit.single` call (via `Runtime.runPromiseExit`), which accumulated in Effect's `Global.roots` and showed up as a
  fiber/memory leak. `Stream.asyncPush` writes directly into an internal queue and tears down with the surrounding scope.
  - Move subscription teardown into `Effect.acquireRelease` so the underlying `graphql-ws` `dispose()` is now tied to scope finalization rather than fiber termination.


GRID-SERVICES-GIT (commented):
Nice fix — switching to `Stream.asyncPush` and tying teardown to the scope via `Effect.acquireRelease` is the right shape for this leak, and pulling the transition rules into a pure, exhaustively-unit-tested state machine (`decideItem`/`decideResume`/`decideComplete`/`decideTerminate`) is really clean. A few things I'd want to understand before signing off — mostly "help me see the bounds," not blockers:

**1. The `asyncPush` queue is `{ bufferSize: 'unbounded' }`, so the memory bound looks "soft."**
`decideItem` doesn't consider `paused`, and the pause/dispose happens *after* `emit.single(item)`. Because `dispose()` isn't instantaneous, any `onItem` callbacks that arrive in the dispose-latency window (same `generation`, `key > lastKey`) still pass `decideItem` and get emitted into the unbounded queue, pushing `inFlight` past `bufferSize`. Steady-state is bounded, but the overshoot per pause-cycle is roughly (source dispose latency × push rate) — at the ~1k msg/sec this targets, even a few ms of latency adds items each cycle. Is `graphql-ws` `dispose()` synchronous enough that this is negligible in practice? If not, would *dropping* items once `paused` (they'd be re-delivered on resume, since `variables(lastKey)` re-opens inclusively) hard-cap the queue instead of relying on dispose latency?

**2. `resumeThreshold < bufferSize` (and positivity/integer-ness) isn't enforced at runtime.**
The invariant is documented on `BackpressureOptions`, but `withBackpressure` uses the values directly with no guard. A caller passing `resumeThreshold >= bufferSize` (or a non-positive value) silently breaks the hysteresis — it can pause and immediately resume (thrash), or never resume. A small validation (throw or clamp) at construction would make the contract self-enforcing.

**3. The subscription wiring is duplicated verbatim between `dust-wallet` and `shielded-wallet`.**
The added lines in `dust-wallet/src/v1/RunningV1Variant.ts` and `shielded-wallet/src/v1/RunningV1Variant.ts` are byte-for-byte identical. Since `Backpressure.ts` already lives in the shared `indexer-client`, could this wiring be factored there too? Two identical copies tend to drift over time.

**4. The pure state machine is thoroughly tested, but I didn't spot integration coverage for the wiring.**
The `decide*` functions are covered exhaustively (nice) — but `withBackpressure` itself (the asyncPush queue + `Stream.tap` resume + the dispose-latency / late-callback path from #1) doesn't appear to have a test. A test driving a fake `Source` that keeps emitting briefly after `dispose()` would pin down the overshoot behavior and lock in the resume hysteresis end-to-end.

Solid work overall — happy to dig into any of these.

CjDabrow (commented):
Nice fix switching to `Stream.asyncPush` and tying teardown to the scope via `Effect.acquireRelease` is the right shape for this leak, and pulling the transition rules into a pure, exhaustively-unit-tested state machine (`decideItem`/`decideResume`/`decideComplete`/`decideTerminate`) is really clean. A few things I'd want to understand before signing off  mostly "help me see the bounds," not blockers:

**1. The `asyncPush` queue is `{ bufferSize: 'unbounded' }`, so the memory bound looks "soft."**
`decideItem` doesn't consider `paused`, and the pause/dispose happens *after* `emit.single(item)`. Because `dispose()` isn't instantaneous, any `onItem` callbacks that arrive in the dispose-latency window (same `generation`, `key > lastKey`) still pass `decideItem` and get emitted into the unbounded queue, pushing `inFlight` past `bufferSize`. Steady-state is bounded, but the overshoot per pause-cycle is roughly (source dispose latency × push rate) at the ~1k msg/sec this targets, even a few ms of latency adds items each cycle. Is `graphql-ws` `dispose()` synchronous enough that this is negligible in practice? If not, would *dropping* items once `paused` (they'd be re-delivered on resume, since `variables(lastKey)` re-opens inclusively) hard-cap the queue instead of relying on dispose latency?

**2. `resumeThreshold < bufferSize` (and positivity/integer-ness) isn't enforced at runtime.**
The invariant is documented on `BackpressureOptions`, but `withBackpressure` uses the values directly with no guard. A caller passing `resumeThreshold >= bufferSize` (or a non-positive value) silently breaks the hysteresis it can pause and immediately resume (thrash), or never resume. A small validation (throw or clamp) at construction would make the contract self-enforcing.

**3. The subscription wiring is duplicated verbatim between `dust-wallet` and `shielded-wallet`.**
The added lines in `dust-wallet/src/v1/RunningV1Variant.ts` and `shielded-wallet/src/v1/RunningV1Variant.ts` are byte-for-byte identical. Since `Backpressure.ts` already lives in the shared `indexer-client`, could this wiring be factored there too? Two identical copies tend to drift over time.

**4. The pure state machine is thoroughly tested, but I didn't spot integration coverage for the wiring.**
The `decide*` functions are covered exhaustively (nice)  but `withBackpressure` itself (the asyncPush queue + `Stream.tap` resume + the dispose-latency / late-callback path from #1) doesn't appear to have a test. A test driving a fake `Source` that keeps emitting briefly after `dispose()` would pin down the overshoot behavior and lock in the resume hysteresis end-to-end.

Solid work overall  happy to dig into any of these.

Craphtex (commented):
I've loaded this with Lace and while it has the effect of keeping the memory pressure down Lace is still eventually toppling over by an OOM kill when syncing preprod. It lasts longer but the memory usage keeps piling.

Given that it's hard to deterministically reproduce a sync even on the same machine these figures are a bit arbitrary but indicative.

For preprod sync the shared effects observed are
- crashes by OOM kill
- reloading resets the progress a bit before resuming
- each restart takes the sync further
- eventually the sync completes

and the different ones are
- Latest Lace (no fix): Crashes fairly early on (~50%) and requires many reloads (7+).
- Latest Lace + [83be88b](https://github.com/midnightntwrk/midnight-wallet/pull/442/commits/83be88bce63a1aac1f87767fc632b6249e1d02df) (async push): Crashes relatively later (~90%) and requires few reloads (~2).
- Latest Lace + [fed5006](https://github.com/midnightntwrk/midnight-wallet/pull/442/commits/fed50061e8c6cb8b2d432644bbd55f306d41619e) (async push + backpressure): Same as [83be88b](https://github.com/midnightntwrk/midnight-wallet/pull/442/commits/83be88bce63a1aac1f87767fc632b6249e1d02df) but much slower sync.

It looks like the GC doesn't feel the need to clean until it's too late, and this fix help patching the woulds but doesn't seem to resolve the underlying issue. I'm trying to collect some measurements in the hope that could guide us further.

agronmurtezi:
@Craphtex I have discovered another issue with the wasm memory usage which I fixed in this PR - https://github.com/midnightntwrk/midnight-wallet/pull/450

Can you apply both those fixes (#442 and #450) and see if that resolved it?

Craphtex:
@agronmurtezi I applied both PR fixes to Lace and the result is unfortunately the same in the sense that it crashes with an OOM kill around 95% synced (of preprod).

I pulled some measurements and this is the heap and footprint at different stages of the sync.
image

Do you have data on the amount of content there is at different blocks of the chain? I'm worried there's a correlation between the chain usage and this large memory footprint, which then would be a bit foreboding regarding mainnet.

agronmurtezi:
@Craphtex thanks, I'll take a closer look - can you share a development build of lace so I can do some testing?

agronmurtezi:
@Craphtex sharing here the progress from the private chat.

I created a test page in browser to monitor the sync memory and it looks pretty reasonable, I'm going to merge this PR and share a build with you so you can verify it without monkey patching.

Screenshot 2026-06-12 at 15 04 56

kapke (approved):

PR #450, opened 2026-06-09
fix(runtime): keep only the latest state in the state-changes stream

agronmurtezi:
## Summary

  Fixes unbounded memory growth during sync: the runtime state-changes stream kept every
  past wallet state alive, including the wasm instances those states hold.

  ## Why

  The stream used a shared buffer with replay. The replay buffer (in Effect, up to and
  including 3.21.3) keeps every published value in a linked list, and a long-lived
  subscriber holds on to its starting position in that list forever — so every state
  published during its lifetime stays in memory. In a syncing wallet this meant thousands
  of retained states and steadily growing wasm memory.

  ## Changes

  - `Runtime.stateChanges` no longer uses a shared buffer with replay. Each subscriber now
  reads the state ref's changes directly, through a sliding buffer of capacity 1.
  - Subscribers still receive the current state immediately on subscribe and always end up
  with the latest state, but may skip intermediate states if they can't keep up. Memory is
  bounded to the current state plus at most one buffered state per subscriber.
  - Runtime tests updated to this contract: received states must arrive in order and end at
  the final state, but may skip intermediate ones.


iangregsondev (dismissed):

packages/runtime/src/testing/utils.ts
```diff
@@ -46,6 +46,40 @@ export const toProtocolStateArray = <T>(
...
+ * This matches the runtime's latest-value state stream contract: a subscriber always converges on the latest state, but
+ * may skip intermediate states when it lags behind the producer. Received states must therefore never be reordered,
+ * fabricated, or repeated out of order — but any prefix of intermediate states may be missing.
+ *
+ * @internal
+ */
+export const isOrderedSubsequenceOf = <T>(
+  received: readonly T[],
+  expected: readonly T[],
+  equals: (a: T, b: T) => boolean,
```
kapke: This is very effect-heavy code. I think taking `Eq` from effect instead would be better

packages/runtime/src/testing/utils.ts
```diff
@@ -46,6 +46,40 @@ export const toProtocolStateArray = <T>(
...
+  return searchEnd >= 0;
+};
+
+/**
+ * Equality of {@link ProtocolState.ProtocolState} values over primitive states, for use with
+ * {@link isOrderedSubsequenceOf}.
+ *
+ * @internal
+ */
+export const protocolStateEquals = <T>(a: ProtocolState.ProtocolState<T>, b: ProtocolState.ProtocolState<T>): boolean =>
```
kapke: Define as `Eq` instance next to `ProtocolState`

kapke (dismissed):

packages/runtime/src/testing/utils.ts
```diff
@@ -46,6 +46,42 @@ export const toProtocolStateArray = <T>(
...
+  return searchEnd >= 0;
+};
+
+/**
+ * Equality of {@link ProtocolState.ProtocolState} values over primitive states, for use with
+ * {@link isOrderedSubsequenceOf}.
+ *
+ * @internal
+ */
+export const protocolStateEquals: Equivalence.Equivalence<ProtocolState.ProtocolState<unknown>> = Equivalence.struct({
```
kapke: This still should be defined at the `ProtocolState` in the abstractions package

kapke (dismissed):

kapke (approved):

PR #458, opened 2026-06-10
fix(hd): return keyOutOfBounds for invalid BIP32 path components

### LINKED ISSUE #454: @midnight-ntwrk/wallet-sdk-hd: deriveKeyAt() throws on out-of-range/invalid index instead of returning { type: 'keyOutOfBounds' }
## Summary
`RoleKey.deriveKeyAt(index)` (and `CompositeRoleKey.deriveKeysAt`) is typed to return a discriminated union `DerivationResult = { type: 'keyDerived', key } | { type: 'keyOutOfBounds' }`, so callers branch on `result.type`. But `index` is never validated, and the underlying `@scure/bip32` `derive(path)` **throws** for out-of-range (≥ 2³¹), negative, or non-integer indices. So instead of the documented `{ type: 'keyOutOfBounds' }`, callers get an uncaught exception. The `keyOutOfBounds` branch is effectively unreachable for in-range inputs (the genuine BIP32 invalid-scalar case is ~1-in-2¹²⁷), so the only realistic "bad index" path throws rather than returning the typed result.

## Environment
- `@midnight-ntwrk/wallet-sdk-hd` (midnight-wallet, `packages/hd`), current `main`
- `@scure/bip32` (resolved ^2.x)

## Affected code
`packages/hd/src/HDWallet.ts` — `RoleKey.deriveKeyAt` (~L123) and `CompositeRoleKey.deriveKeysAt` (~L141):
```ts
public deriveKeyAt(index: number): DerivationResult {
  const path = `m/${PURPOSE}'/${COIN_TYPE}'/${this.account}'/${this.role}/${index}`;
  const derivedKey = this.rootKey.derive(path);   // throws on a bad index — not wrapped
  return derivedKey.privateKey ? { type: 'keyDerived', key: derivedKey.privateKey } : { type: 'keyOutOfBounds' };
}
```

## Steps to reproduce
Given a `RoleKey` (or `CompositeRoleKey`) built from a seed:
```ts
roleKey.deriveKeyAt(2 ** 31); // expected: { type: 'keyOutOfBounds' } — actual: throws "Invalid index"
roleKey.deriveKeyAt(-1);      // throws "invalid child index: -1"
roleKey.deriveKeyAt(1.5);     // throws "invalid child index: 1.5"
roleKey.deriveKeyAt(NaN);     // throws "invalid child index: NaN"
```
(The existing property test only fuzzes in-range indices via `fc.nat(2147483647)`, so the out-of-range / negative / non-integer paths are never exercised and the `keyOutOfBounds` branch never actually fires.)

## Expected
Out-of-range / invalid indices return `{ type: 'keyOutOfBounds' }` — the documented contract — so callers that `switch` on `result.type` can handle it (e.g. loop-based key scanning that increments `index` toward/past 2³¹).

## Actual
An uncaught exception is thrown; the discriminated-union contract is violated, crashing callers that follow the typed API.

## Suggested fix
Either:
- validate up front — `Number.isInteger(index) && index >= 0 && index < 2 ** 31` (and the same for `account` / hardened segments), returning `{ type: 'keyOutOfBounds' }` otherwise; or
- wrap `this.rootKey.derive(path)` in try/catch and map a thrown range error to `{ type: 'keyOutOfBounds' }`.

Also worth tightening the property test to include out-of-range / negative / non-integer indices. Happy to open a PR.

agronmurtezi:
Fixes #454

## Summary

`RoleKey.deriveKeyAt` interpolates `account`/`role`/`index` straight into the BIP32 derivation path, so any value outside the non-hardened range (non-integer, negative, or `>= 2^31`) escaped as `@scure/bip32`'s thrown `Error: invalid child index: ...` instead of the typed `keyOutOfBounds` result the API models. `CompositeRoleKey.deriveKeysAt` inherited the same behavior. All four repro cases from the issue (`2 ** 31`, `-1`, `1.5`, `NaN`) now return `{ type: 'keyOutOfBounds' }`.

This takes the issue's first suggested approach: guard all three path components against the BIP32 non-hardened child range (integers in `[0, 2^31)`, using `HARDENED_OFFSET` exported by `@scure/bip32`) before deriving, making `deriveKeyAt` total. An explicit guard was chosen over wrapping `derive()` in try/catch so that genuine defects (e.g. deriving after `clear()` wiped the private key) still throw instead of being misreported as out-of-bounds.

## Changes

- `packages/hd/src/HDWallet.ts` — validate `account`, `role`, and `index` in `deriveKeyAt`; return `keyOutOfBounds` on any invalid component
- `packages/hd/test/tests.test.ts` — property-based tests for out-of-range/negative/non-integer indices and accounts (as the issue suggested, these paths were previously never exercised), an invalid role smuggled past the type system, and `deriveKeysAt` reporting all requested roles
- `.changeset/hd-out-of-bounds-derivation.md` — patch changeset

## Testing

- `yarn test --filter=@midnight-ntwrk/wallet-sdk-hd` — 9/9 passing; the 4 new tests were written first and confirmed failing with the `@scure/bip32` throw before the guard was added

### DIFF
--- a/.changeset/hd-out-of-bounds-derivation.md
+++ b/.changeset/hd-out-of-bounds-derivation.md
@@ -1,0 +1,7 @@
+---
+'@midnight-ntwrk/wallet-sdk-hd': patch
+---
+
+`deriveKeyAt`/`deriveKeysAt` now return `keyOutOfBounds` for invalid BIP32 path components (non-integer, negative, or
+`>= 2^31` account/role/index values) instead of leaking the underlying `invalid child index` error thrown by
+`@scure/bip32`.
--- a/packages/hd/src/HDWallet.ts
+++ b/packages/hd/src/HDWallet.ts
@@ -10,7 +10,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-import { HDKey } from '@scure/bip32';
+import { HARDENED_OFFSET, HDKey } from '@scure/bip32';
 
 type ValueOf<T> = T[keyof T];
 
@@ -34,6 +34,11 @@ type HDWalletResult =
 
 const PURPOSE = 44;
 const COIN_TYPE = 2400;
+
+// BIP32 path components (account, role, index) are integers in [0, 2^31);
+// @scure/bip32 inlines this check inside derive() and throws, so guard here
+const isValidChildIndex = (value: number): boolean =>
+  Number.isSafeInteger(value) && value >= 0 && value < HARDENED_OFFSET;
 
 const CompositeDerivationResult = {
   fromResults: <T extends readonly Role[]>(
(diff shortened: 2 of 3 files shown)


jtsang586 (approved):
Lgtm

PR #473, opened 2026-06-15
[CI mirror] #404 tx-history indexer race

iangregsondev:
Mirror of #404 to run CI with secrets. Do not merge — validating for #404 by @BossChaos.


packages/shielded-wallet/src/v1/TransactionHistory.ts
```diff
@@ -124,6 +124,10 @@ export const makeDefaultTransactionHistoryService = (
         const statusQuery = yield* TransactionHistoryDetail;
         const result = yield* statusQuery({ transactionHash: hash });
         const tx = result.transactions[0];
+        if (tx === undefined) {
+          // Indexer hasn't indexed this tx yet — retry will handle it
+          throw new Error(`Transaction ${hash} not yet indexed`);
```
agronmurtezi, line 129: isn't it better if we don't throw and instead wait some seconds and try again?

iangregsondev:
This PR is going to be closed, it needs to be re-tested.

PR #508, opened 2026-06-25
Move wallet specification into the wallet repository

kapke:
# Description

Wallet spec living in a separate repository has become quite a pain; this PR moves it here from the midnight-architecture one; including reference, unsafe implementation of address format and key derivation


agronmurtezi (dismissed):
changeset is missing

iangregsondev (approved):

Issue #504, opened 2026-06-23
Issue: wallet signing callback is synchronous, blocking async signer backends (MPC/HSM)


jtsang586:
Every wallet signing entry point takes a synchronous signer callback of the shape (data: Uint8Array) => ledger.Signature. MPC (threshold protocols with inter-party network round-trips) and HSM (PKCS#11 device calls) signers are inherently asynchronous — they return Promise<Signature>. An async backend therefore cannot be invoked from inside the current callback, so it cannot be plugged into the normal signing path without a workaround.

- The whole point of MPC/HSM is that the secret never materializes in-process and signing happens over the network/on-device — fundamentally async. A synchronous callback assumes the key (or a local keystore that signs instantly) is available in-process, which is exactly what these backends avoid.
- Callers reaching for deasync/blocking hacks to satisfy the sync signature would block the event loop and reintroduce the very coupling MPC/HSM remove.

Issue created as discussed.

PR #517, opened 2026-06-30
feat: support async signers (MPC/HSM) via SigningService and keystore.signDataAsync (#504)

agronmurtezi:
## Summary

Implements #504 — every wallet signing entry point now accepts an **asynchronous** signer callback
(`(data: Uint8Array) => Promise<ledger.Signature>`, exported as `SignSegment`), so MPC (threshold protocols with
network round-trips) and HSM (PKCS#11) backends plug into the normal signing path. Previously the callback was
synchronous (`(data) => ledger.Signature`), which couldn't invoke an out-of-process signer without event-loop-blocking
hacks — the gap `ecdsaMpc.test.ts` explicitly documented.

## Approach

- **Signing is now a service.** A new `SigningService` (the Effect/imperative-shell layer, alongside the proving and
  submission services) drives the async signer. The pure transformations — `collectSignableData`, `attachSignatures`
  (incl. the scheme-mismatch check) — stay in `TransactionOps`, so the `Transacting` capability stays pure `Either` and
  no longer carries `signUnproven/UnboundTransaction`.
- **Errors stay typed.** A signer rejection surfaces as a `SignError`; a signature-scheme mismatch is still rejected
  before anything is attached (no partially-signed transaction escapes). Segments are signed sequentially.
- **Keystore ergonomics.** `UnshieldedKeystore` keeps its synchronous `signData(data): Signature` primitive and gains
  a `signDataAsync` adapter conforming to the async callback shape, so callers pass `keystore.signDataAsync` directly
  instead of wrapping at each call site.
- **Gap closed.** `ecdsaMpc.test.ts` / `ecdsaHsm.test.ts` now drive `FakeMpcCoordinator` / `FakeHsm` through the real
  `SigningService` end-to-end (threshold/timeout/availability/auth failures included), plus a new `SigningService` unit
  test and a retargeted `TransactionOps.attachSignatures` scheme-mismatch test.

## Breaking change

Signer callbacks must now return a `Promise`. **Major** bump for `@midnightntwrk/wallet-sdk-facade` and
`@midnightntwrk/wallet-sdk-unshielded-wallet`. Migration:

```ts
// before
wallet.signRecipe(recipe, (data) => keystore.signData(data));
// after
wallet.signRecipe(recipe, keystore.signDataAsync);
```

## Verification

- `yarn typecheck` (34/34), `yarn test:unit` (49/49), `yarn lint` (53/53), `yarn dist` (16/16) all green.
- The integration and `*.undeployed`/`*.remote` e2e tiers require Docker/live infra (not run here) but typecheck and
  compile; they run in CI.


kapke (approved):

jtsang586 (approved):
Lgtm

PR #532, opened 2026-07-03
fix(shielded-wallet, dust-wallet): keep tx-history shielded/dust sections when the indexer lags events (#401)

iangregsondev:
Relates to #401 (kept open for QA; this PR does not auto-close it).

## Problem

The wallet sync path writes a tx-history entry only after a successful `getTransactionDetails(change.source)` call against the indexer's HTTP endpoint. When the WS events stream delivers a relevant event **before** the indexer's HTTP `transactions(...)` endpoint has indexed the same hash (a normal race on any deployed network), the lookup dereferenced an empty result array:

```ts
const tx = result.transactions[0];                       // undefined during the race
const rawStatus = tx.__typename === 'RegularTransaction' // → TypeError → Cause.Die defect
```

The synchronous `TypeError` became a `Cause.Die` defect, so:
- `Effect.retry(...)` never engaged (it retries typed failures, not defects),
- `Effect.mapError(...)` never wrapped it, and
- the defect was swallowed by `catchAllCause → Console.error`.

`applyUpdate` had already advanced `appliedIndex`, so the change was never re-processed. **Net effect:** the shielded/dust section of that tx-history entry was permanently lost, while balances and coin/utxo sets still updated correctly.

This affected **both** `shielded-wallet` and `dust-wallet` — the two have the same `getTransactionDetails` / sync-fork shape, so the same defect was present in each.

### User-facing symptom

The missing section is permanently absent from `WalletFacade.queryTxHistoryByHash` / `getAllFromTxHistory` — the entry shows `sections: "unshielded,dust"` only (shielded missing), or the dust section missing in the dust case. This is what surfaced the bug: the e2e `tokenTransfer.remote.test.ts` → *'can perform a combined transaction'* **hangs** in `waitForTxInHistory`, because its predicate `(e) => e.shielded !== undefined && e.unshielded !== undefined` is never satisfiable for the affected tx. With the section now written after retry, that predicate becomes satisfiable — **QA should confirm the hang is gone on remote (qanet)**.

## Fix

Applied to both `shielded-wallet` and `dust-wallet` (`TransactionHistory.ts` + `RunningV1Variant.ts`):

1. **Typed failure instead of a defect** — the empty-array case now fails with a typed `TransactionHistoryError` via `EArray.head(...).pipe(Effect.orElseFail(...))` (the `UnshieldedState` idiom) rather than dereferencing `undefined`. Because it's a typed failure, the retry schedule actually engages.
2. **Longer, configurable, jittered retry window** — the retry is now a jittered exponential bounded by `Schedule.upTo(retryWindow)`, where `retryWindow` is a new optional config field `transactionDetailsRetryWindow` (**default 2 minutes**, up from the old ~7 s). Jitter stops a batch of concurrent lookups retrying in lockstep against an already-behind indexer.
3. **Bounded sync fan-out** — the tx-history `Effect.forEach` goes from `concurrency: 'unbounded'` to `concurrency: 8`, so a burst of N changes can't peak at ~4N simultaneous queries against a lagging indexer.
4. **No double-wrapping** — the final `mapError` lets our own `TransactionHistoryError` through untouched (keeping the useful "not yet indexed" message at the top level) and only wraps the indexer query's `ClientError`/`ServerError`.
5. **Loud, not silent** — the terminal `catchAllCause` now emits a structured `Effect.logError` carrying the affected `txHash` instead of a silent `Console.error` defect.

`dust-wallet` additionally preserves its `fees` extraction (dust is the canonical source for fees).

### Honest limitation

This narrows the race but is **not a durability guarantee**: if the indexer lags beyond the retry window, the section is still lost — the change is not re-processed, even across restarts — but the failure is now logged (with the hash) rather than swallowed.

## Acceptance criteria (from #401)

- ✅ Empty first response then populated within the retry window → shielded/dust section is eventually written.
- ✅ Sustained outage beyond the retry window → structured error carrying `change.source`, not a silent defect.

## Tests

New unit files (written test-first, confirmed **red** before the fix), injecting a fake `TransactionHistoryDetail` query via its service tag and driving the retry schedule with `TestClock` (a short configured window keeps them fast/deterministic):

- `packages/shielded-wallet/src/v1/test/transactionHistory.test.ts`
- `packages/dust-wallet/src/v1/test/transactionHistory.test.ts` (also asserts `fees` is parsed)

Each covers:
- empty-then-populated → `getTransactionDetails` eventually succeeds with parsed details;
- always-empty → fails with a typed `TransactionHistoryError` (asserted **not** a `Cause.Die`).

## Verification

- New tests pass; full unit suites green (shielded: 51 passed; dust: 33 passed)
- `typecheck`, `lint`, and Effect LSP diagnostics all clean


agronmurtezi:
The same bug exists in dust-wallet: `packages/dust-wallet/src/v1/TransactionHistory.ts:161` still reads `result.transactions[0]` with no empty-result handling, and `RunningV1Variant.ts:169` still swallows the error. Dust history entries get dropped under the same conditions this PR fixes for shielded. Can we fix it here too, or open a follow-up?

agronmurtezi:
Retries stop after ~7 seconds. If the indexer lags longer than that, the entry is still lost for good — it's never re-processed, even after a restart — just with a warning now. Consider a longer or configurable retry window, and softening the changeset's "re-queries until the indexer catches up".

agronmurtezi:
Two nits: (1) the final `Effect.mapError` wraps this error in a second `TransactionHistoryError`, hiding the useful "not yet indexed" message inside `cause` — let it pass through instead. (2) This block can be one line: `yield* EArray.head(result.transactions).pipe(Effect.orElseFail(() => new TransactionHistoryError({ message: ... })))`, same pattern as `UnshieldedState.ts:82`.

agronmurtezi:
With unbounded concurrency and no jitter, a burst of N changes retries in lockstep (1s/2s/4s) — up to 4N simultaneous queries against an indexer that's already behind. `Schedule.jittered` and/or a concurrency limit would spread them out.

agronmurtezi:
Minor: this is permanent data loss, so `logError` fits better than `logWarning`.

iangregsondev:
The following issues were also found

1. The `concurrency: 8` only limits one batch: each sync update forks its own fiber with its own limit of 8, and those fibers pile up while their lookups retry (now up to 2 minutes each). Total in-flight queries against the indexer are 8 × the number of live batches — still unbounded under sustained lag. A single semaphore shared across batches (held in the variant) would make the cap real. Same in the dust twin..

2. The retry has no error filter, so failures that can never succeed (wrong URL, 4xx, schema mismatch) now retry for the full 2-minute window while holding a concurrency slot — before this PR they gave up after ~7s. `Effect.retry({ schedule, while: (e) => e instanceof TransactionHistoryError })` keeps the retry for the not-yet-indexed case and fails fast on the rest. Same in the dust twin..

3. The delay doubles without a cap, so with a 120s window the attempts land at 1, 3, 7, 15, 31, 63, 127s (measured with the real schedule): most attempts happen in the first 15s, then there's a 64s gap, and the last attempt overshoots the window — up to ~3.4 minutes with unlucky jitter. If the indexer recovers at t≈70s, the next query is ~1 minute away. Capping the delay fixes both: `Schedule.exponential(Duration.seconds(1)).pipe(Schedule.either(Schedule.spaced(Duration.seconds(10))), Schedule.jittered, Schedule.upTo(retryWindow))` probes every ≤10s and ends at ~124s (verified). Same in the dust twin..

4. This comment still describes the old schedule ("retrying up to 4× … ~4N simultaneous queries") — the retry is now a time window. The dust twin already has the corrected wording; can you mirror it here?.

These are being addressed.

iangregsondev:
> 4. This comment still describes the old schedule ("retrying up to 4× … ~4N simultaneous queries") — the retry is now a time window. The dust twin already has the corrected wording; can you mirror it here?.
> 
> These are being addressed.

Fixing number 1 also fixed 4.

agronmurtezi (approved):

PR #571, opened 2026-07-14
fix: read persisted data written by older SDKs, and gate the formats against silent breaks

### LINKED ISSUE #559: [DRAFT] Serialization compatibility: migrations, per-train fixtures, and PR gate
> [!WARNING]
> **DRAFT / WIP** — this issue is being reviewed and its scope may still change. Do not pick it up yet.

## Problem

The SDK persists user data across app restarts through several public serialize/restore APIs:

- Wallet state snapshots (`ShieldedWallet` / `UnshieldedWallet` / `DustWallet` `.serialize()` / `.restore()`)
- Transaction history (`TransactionHistoryStorage` in `abstractions`, introduced with abstractions 2.1.0)
- Pending transactions (`PendingTransactions` in `capabilities`)

None of these formats carry a version envelope, and there is no migration path when a schema changes. Nothing today stops a PR from changing a persisted format in a way that makes the SDK unable to read data it previously wrote — the failure only surfaces later, in apps, as a hard `ParseError` on `restore()`.

A concrete case is already in front of us: the 3.0 line of `abstractions` evolves the tx-history entry schema with new required fields (`lifecycle`; `identifiers` also becomes required). Before that ships stable, `restore()` needs a migration so payloads written by 2.x apps keep loading.

This issue covers **adding that migration** and **putting guardrails in place so persisted-format changes are always caught and always migrated**.

## 1. Migrate tx-history payloads to the new (`lifecycle`) schema

`InMemoryTransactionHistoryStorage.restore()` must accept payloads written by abstractions 2.1.0:

- **Migrate on read**: when an entry has no `lifecycle`, synthesize it (and synthesize `identifiers` when absent). The wallet-specific sections (shielded/unshielded/dust) have not changed shape, so migration only needs to fill in the two missing fields — no section rewriting.
- **Add a version envelope to `serialize()`** going forward (e.g. `{version: 1, entries: [...]}`), with the reader accepting both enveloped and legacy bare-array payloads. Every future schema change then has an explicit version to migrate from, instead of guessing by shape.

## 2. Serialization compatibility test package (fixtures vs current code)

Add a private workspace package (e.g. `packages/serialization-tests`) containing:

- **Committed, frozen fixtures**: serialized wallet snapshots, tx-history payloads, and pending-transactions payloads as written by **previously published SDK versions** — real content (synced coins, generated dust, pending spends, populated history), not hand-written JSON.
- **Tests that restore every fixture through the current workspace code's public APIs** and assert the _content_ survives (balances, coin values, UTXO flags, history entries) — not merely that decoding doesn't throw, since silent data loss is also a failure mode.

## 3. Release trains: definition and how to create one

Wallet packages are published in lockstep — a **release train** is one coordinated publish of shielded/unshielded/dust/facade/abstractions/capabilities plus the exact ledger version apps resolved at that date. Fixtures are captured **once per train** and never regenerated: published versions are immutable, so a train's fixtures are frozen forever.

Creating a new train must be a documented, repeatable procedure:

- A standalone fixture generator (not a yarn workspace member) installs the train's published versions via npm aliases, with the train's exact ledger pinned via `overrides`.
- Wallet content is produced by driving that train's own ledger through a deterministic on-chain scenario (mints, dust registration, night claim, transfers) and replaying the events through the train's own sync path.
- Every fixture is **self-checked at generation**: the version that wrote it must read it back, or generation fails.
- The generator must only add the new train's fixtures — existing trains' files must never be rewritten.
- Document per-train boundaries (what changed in each persisted format between trains) alongside the fixtures.

## 4. PR gate: run all fixtures against the current code on every PR

- A root-level script (e.g. `yarn test:serialization`) runs the fixture-restore tests against the checked-out code, building current workspace packages first.
- Wire it into PR CI as part of the required checks: a red run means the PR changed a persisted format without a migration.
- Tests must be pure unit tests (no Docker/network) so the gate is fast; cache inputs must include the fixture files so new fixtures always trigger a run.

## 5. Detect when a release changes a persisted format, and capture a new train when it does (needs investigation)

Not every release needs a new train — most releases don't touch persisted formats, and their fixtures would be redundant. What we need is a way to **notice** when a release did change what the SDK writes, and capture a new train at that point.

Investigate the simplest reliable detection, e.g.: serialize the standard scenario with the current code and compare against the latest train's fixtures — if the output shape changed, a new train is due. Whether that check runs on release, on a schedule, or as part of the PR gate is part of the investigation; the outcome should be a clear signal ("this release needs a new train") followed by the documented manual procedure from section 3.

## Acceptance criteria

- [ ] `InMemoryTransactionHistoryStorage.restore()` reads non-empty payloads written by abstractions 2.1.0 (lifecycle/identifiers synthesized on read)
- [ ] `serialize()` emits a versioned envelope; `restore()` accepts both enveloped and legacy payloads
- [ ] Fixture package exists with fixtures for all historically published trains (back to 1.0.0), covering wallet snapshots, tx history, and pending transactions
- [ ] `yarn test:serialization` restores every fixture through current public APIs and asserts content — including the 2.1.0-era tx-history fixtures passing post-migration
- [ ] The gate runs on every PR in CI as a required check
- [ ] "Creating a new train" is documented and only ever adds fixtures
- [ ] A documented mechanism exists to detect that a release changed a persisted format and that a new train should be captured

iangregsondev:
Related issue: #559

# Description

The SDK hands an application five strings to store and hand back later — the shielded, unshielded and dust wallet snapshots, the transaction history, and the pending transactions. An application may hand one back long after the version that wrote it stopped shipping.

Two of those had already broken, and neither was found by a test.

**A transaction history written by abstractions 2.1.0 — npm `latest` — cannot be read.** Making `lifecycle` and `identifiers` required means every non-empty history throws a `ParseError` on restore, so an application that upgrades cannot open the history its users already have. It is not in a stable release yet, but it is live on the `beta` tag.

**A shielded 1.0.0 snapshot silently loses its transaction history.** That release embedded the history in the snapshot as `txHistory`. When the field was removed from the schema, Effect Schema's default of ignoring unknown keys meant those snapshots restored *without any error* and the history was discarded on the next save. Nothing failed, so nothing could notice.

This fixes both, versions all five surfaces, and adds gates so the next one cannot merge unnoticed.

# Proposed Solution

Every surface now carries `{ version: 'vN', ... }`, a string literal, as pending transactions already did. A payload with no `version` is the first version of its surface — the envelope did not exist when it was written, so its absence is the marker.

The transaction history moves to `v2`, because its shape genuinely changes. The three snapshots are labelled `v1` on the V1 variant, which is what they have always been: the shape does not change, only the label, so a snapshot written by an earlier release restores exactly as before; the only thing it gains on the way back out is the `version` label itself. The label earns its keep the day a snapshot's shape *does* change — that becomes `v2` and one upgrade step is written, instead of a reader having to guess which of two undifferentiated shapes it is holding. That guessing is precisely the position the transaction history was in.

On `v2` that day has already come for one surface. The V2 unshielded variant, on ledger-v9, stores the verifying key as `{ tag, value }` where the V1 variant stores a bare string, because ledger-v9 signs with more than one scheme. A retyped field is a new format version, so the V2 unshielded writer writes `version: 'v2'`, and one pure step — a bare-string key becomes `{ tag: 'schnorr', value }` on the JSON before the schema runs — reads a `v1` snapshot. The tagged shape only ever shipped in the `4.0.0` pre-releases, which carry no promise, so those snapshots simply read as `v2`. Shielded and dust stay on `v1` for both variants: the V2 shielded writer adds only optional fields, and dust is identical.

Upgrades are chained one version at a time and never skip, so each step only has to know the version immediately before it and can be written once and left alone. They are called **format upgrades**, not migrations — the runtime already uses that word for moving live wallet state across a hard fork.

The transaction-history upgrade runs on the encoded JSON before any entry schema sees it, so wallet packages that extend the entry shape need no change of their own. Its rule is **fill in what is missing, never overwrite what is there**, which is what makes it safe to apply to a payload it has already touched.

Entries written in the first format all become `finalized`, whatever their `status`. The only writer at the time ran from the sync path, after the indexer had returned the transaction inside a block; there was no pending-entry writer, so no entry could have been anything else. `status` answers "what happened when it ran", `lifecycle` answers "did it reach a block" — a `FAILURE` reached a block too, it just failed once it ran.

Nothing is invented that was not recorded. An upgraded entry has no `finalizedBlock`, so that field becomes optional and a reader that needs the block fetches it from the indexer by transaction hash. A payload that cannot be read comes back as a `Left` carrying a tagged error that names the surface and the version the payload was read from, rather than being handed back as an empty store — losing a history quietly is worse than failing to open it.

**Public API.** `finalizedBlock` on a finalized lifecycle is now optional. `InMemoryTransactionHistoryStorage.restore` returns an `Either` rather than throwing, so an unreadable payload is a value the caller has to handle rather than an exception it has to know about; `upgradeV1ToV2` takes an array and `upgradeToCurrentFormat` returns an `Either` for the same reason. A reader that meets a snapshot version it does not know now says which surface it was reading and which version it found, instead of surfacing a bare schema tree. `finalizedTransactionTraits` — the versioned registry pending transactions are read with — is now exported from the facade, so the compatibility tests restore a stored pending-transaction payload exactly as the wallet does rather than freezing a copy of the trait. `Serialization` is now exported from dust's `v1` entry point, matching shielded and unshielded, which already exported it. Both wallet variants write the snapshot `version`; each package keeps its format versions, and any upgrade step, in a ledger-free `src/SnapshotFormat.ts` that both twins import, so the two cannot drift and the V2 variant no longer loads ledger-v8 to read a constant.

**Both variants are writers, and both are gated.** The V2 shielded variant now keeps the transaction history a 1.0.0 snapshot embedded, and the cross-ledger migration carries it across `forks.v9`, so the history is not lost on the first save after the crossing. The V2 dust deserializer names the refused version in `error.message`, as V1 does, and dust's `v2` entry point exports `Serialization`. The four gates below run per writer: drift baselines live under `_baseline/<writer>/`, coverage checks each writer's current version, information preservation and compatibility feed every frozen fixture through both the V1 and the V2 readers, because a build that registers only the V2 variant still has to open what a V1 wallet stored.

**Fixtures are filed by format version, not by release.** A fixture pins a shape, and shapes change far more rarely than releases, so filing by release produces mostly duplicates and ties the corpus to the release process — which would put the obligation to add one inside the changesets release pull request, a bot-generated branch rewritten on every push that cannot safely be hand-edited. Filed by version, the obligation lands on the branch that creates the version, for the person creating it.

**Four gates, all firing in the pull request that does the work, once per writer.** Compatibility asks whether real stored data still loads with the right content. Drift asks whether what we *write* has changed since it was last recorded. Coverage asks whether every version the code declares is backed by a payload someone can point at. Preservation asks whether every key a published payload carried is still there after this build reads it and writes it back.

That last gate is the one that earns its place. Dropping a field does not stop old data decoding, so compatibility stays green; drift does go red, but its own documented remedy is to re-record the baseline, which makes the loss permanent and green. Preservation compares against the frozen fixture, which re-recording cannot touch and CI will not let anyone edit. It reproduces both of the breaks above.

**Production is unaffected by `finalizedBlock` becoming optional.** The only implementations that read it from storage belong to the simulator service, in both wallet variants; production always queries the indexer for transaction details and never reads that field from storage at all. In the simulator there is no indexer to ask, so an entry with no recorded block fails — but with its own message, distinct from the one for a transaction storage has never heard of. Those are different facts and a caller should not have to guess which it hit. `TransactionDetails` is unchanged.

**A pending-transaction TTL bug, found while reviewing this work.** The shielded-offer check in `hasTTLExpired` was inverted: a transaction counted as carrying a fallible offer when it had none. A pending transaction with no shielded offers and no dust spends was given the dust grace period as its expiry and dropped from the pending set once that period elapsed, even with its intent deadline still far in the future. Both ledger traits carried the same inverted check; both are fixed, and one test runs every case against each, at its own ledger's grace period. Unrelated to the format work, but it sits in the code this pull request touches.

**Unshielded gained its first unit test file.** Its snapshot previously had no test of any kind, which is how it went unguarded.

**The unshielded fixtures carry a real key.** The fixtures are synthesised by driving each published release with fixed seeds — the generator that does it ships in `packages/serialization-tests/fixture-generator/`, a standalone npm project outside the yarn workspace because it installs many versions of the same packages side by side, with its README saying when it may and may not be run — and the first cut gave the unshielded wallet a placeholder key and address that no wallet could have produced. This branch checks at the deserialization trust boundary that an unshielded address derives from its key, so those fixtures were regenerated by each release's own keystore from the seed the chain driver already used. Only the key, the address and its hex changed; every UTXO, owner, value, date and sync point is byte-identical, and each regenerated payload was read back by the release that wrote it before being committed. Pending-transaction fixtures are untouched: they are read through the same fork schedule the wallet uses, so an unstamped payload written under ledger-v8 is read by the ledger-v8 trait.

# Testing

`yarn test:unit --filter=@midnight/wallet-serialization-tests` runs the compatibility, drift, coverage and preservation checks, for the V1 and the V2 writers alike. They are pure unit tests — no Docker, no network — and also run as part of `yarn test:unit`. The unshielded `v1 → v2` step has its own unit tests in `unshielded-wallet/src/test/snapshotFormat.test.ts`, and each V2 variant's serialization tests mirror the V1 format-version cases.

Each gate was mutation-tested rather than assumed:

- bumping a surface's version constant without adding a fixture fails coverage, naming the missing version
- editing one field in a baseline fails drift, naming the surface
- removing a field from a schema *without* bumping the version, then re-recording the baseline to silence drift, still fails preservation, naming the lost field

Worth reviewing closely: whether `finalized` is the right lifecycle for every upgraded entry (the reasoning is in the changeset and ADR 0008), whether `restore` returning an `Either` is the right call for a public API that every consumer has to update, and whether the simulator failing an entry with no block is the behaviour you want.

Commits are one per concern and worth reading in order; the fixture folder is large and can be skipped.


jtsang586:
Spent some time testing this branch and wanted to share what came out of it. Short version: the design holds up
well — the three things you flagged for close review all check out — but I found one real defect and two places
where the gate is weaker than it reads. Everything below is measured rather than reasoned, and there are commits
for all of it on `test/pr571-wallet-layer-restore` if you want to cherry-pick.

## 1. The V1 shielded reader silently drops `coinHashesPending`

The V2 shielded writer emits `coinHashesPending`; the V1 schema never declared it; unknown keys are ignored on
decode. So the V1 reader **accepts a V2-written snapshot and discards the field on the next write** — the same
mechanism that lost the embedded `txHistory`.

**This pre-dates the PR** (it's the same on `v2`: the field is on the V2 schema and absent from V1), and no
application reaches it — routing sends such a snapshot to V2, and no single-variant composition is built from V1.
But it's squarely the failure class this PR closes, in the same file, and your new gate is what would catch it the
moment a v9 fixture exists.

Fix is the precedent you already set for `txHistory`: declare it on the V1 schema and carry it through untouched
(`foreignCoinHashesPending` on the V1 `CoreWallet`, named so it's clear the variant never sets or reads it). A V1
wallet then writes byte-for-byte what it wrote before. Dust is unaffected — its twins are field-identical — and
unshielded is already protected by its `v2` literal.

## 2. `informationPreservation` compares key paths, so a value-level loss survives the whole gate

The docstring is explicit that values may legitimately change on re-encode, and that's a fair call. But it leaves a
reachable hole: a field whose *value* is quietly rewritten keeps its path. Drift does fire — and drift's own
documented remedy is `yarn capture`, which makes the loss permanent and green.

Measured on two surfaces, each with the baseline re-recorded first (i.e. the gate at its weakest):

| Mutation | `informationPreservation` | after `yarn capture` |
| --- | --- | --- |
| `registeredForDustGeneration` always decodes `false` | green | **all 343 tests green** |
| a shielded coin-hash nullifier decodes to a constant | green | **all 343 tests green** |

Worth knowing because the docstring reads as if this case is covered. It's covered for *dropped fields*, not for
*corrupted values*.

I added a `valuePreservation.test.ts` that asks the same question of leaf values against the frozen fixture. It
needs no allowlist of "fields that may drift" — measured across every fixture and both writers, every leaf value
round-trips byte-identically, with exactly one exception (the V2 unshielded key retyping, which is asserted
separately to carry the stored string inside the tag).

## 3. The preservation gates pair readers with payloads routing forbids

`informationPreservation` and `valuePreservation` pair every writer with every fixture. The justification is
one-directional —

> Both writers read every frozen fixture, because a build that registers only the V2 variant still has to open what
> a V1 wallet stored.

— but the code runs the other direction too. That's harmless today only because every frozen payload is V1-written.
**The first captured ledger-v9 fixture turns the gates red**, and it'll look like data loss while actually being a
combination no build produces. Since ADR 0008 and the generator README both plan for that capture, it's a scheduled
breakage rather than a hypothetical.

I added an `isHandedTo` predicate that pairs a reader with what routing would hand it, applied in the two
preservation gates and `snapshotCompat`. It filters nothing today, so there's a small test pinning the logic now
rather than leaving it dead until the day it matters.

## 4. Two things still owed

- **`restore` returning an `Either` has no docs coverage.** `packages/docs-snippets` has no example of persisting or
  restoring the transaction history at all, and the repo convention is that public API is exemplified there. Every
  consumer has to update for this one.
- **The downgrade door is closed twice over, and nothing says so.** I'd assumed only the `v2` envelope blocked
  2.1.0; a test against the actually-published 2.1.0 proved otherwise — it also *required* `protocolVersion` and
  `status`, both optional now. So unwrapping the envelope is not a downgrade path. Probably worth a line in the
  release note.

## The three things you asked reviewers to check

All three hold, and here's the evidence rather than just agreement:

<details>
<summary><b>Is <code>finalized</code> the right lifecycle for every upgraded entry? — yes</b></summary>

Stronger than the reasoning in the PR body, I think. Only one stable `@midnightntwrk/wallet-sdk-abstractions` was
ever published — 2.1.0 (2026-06-19); everything else is canary or `3.0.0-beta.*`. Unpacking it:

- `TransactionHistoryCommonSchema` has **no `lifecycle` field**, and `status` is a required literal.
- `TransactionHistoryStorage` is `upsert / getAll / get / serialize` — there is no pending or rejected writer to
  call.
- In the 4.1.0 train the facade only *reads* the store; every write is `upsert` from a wallet sync path, downstream
  of a `getTransactionDetails` that resolves the containing block.

So every entry in a v1 payload reached a block. `gotPending`/`gotRejected` first acquire callers in the 3.0.0-beta
line, which has never been stable.
</details>

<details>
<summary><b>Is the simulator failing a blockless entry right? — the premise is accurate</b></summary>

Swept every `finalizedBlock` reference: exactly four sites read it from storage, all in
`makeSimulatorTransactionHistoryService` (shielded + dust × v1/v2), all now guarded. Unshielded has no simulator
history service. The facade discriminates on `lifecycle.status` but never dereferences `finalizedBlock`. So
"production never reads it from storage" holds.
</details>

<details>
<summary><b>The TTL fix — sound, and the stub is faithful</b></summary>

I was suspicious of a structural stub standing in for a `FinalizedTransaction`, but both ledgers declare exactly
what the stub provides:

```
fallibleOffer:    Map<number, ZswapOffer<P>> | undefined     (v8 and v9)
guaranteedOffer:  ZswapOffer<P> | undefined
intents:          Map<number, Intent<S, P, B>> | undefined
Intent.ttl:       Date
```

So `.size` is the right accessor and the inverted-check fix is right.
</details>

One more that isn't on your list but I went looking for: whether a snapshot written just after the crossing could
carry a sub-fork `protocolVersion` and get routed back to the V1 reader, which would now refuse it on the `version`
literal. It can't — sync folds the version into the state, the variant emits `State(current)` *before*
`VersionChange(Version(current.protocolVersion))`, and the runtime hands that same `lastState` to `migrateState`.
Mutating the migration to carry `ProtocolVersion(1n)` fails several cases in `forkSimulation.test.ts`, so it's
guarded already.

packages/shielded-wallet/src/v2/CoreWallet.ts
```diff
@@ -286,6 +305,7 @@ export const CoreWallet = {
       coinHashesPending: true,
       progress: SyncProgress.createSyncProgress({ ...previous.progress, isConnected: false }),
       protocolVersion: ProtocolVersion.ProtocolVersion(previous.protocolVersion),
+      ...(previous.legacyTxHistory === undefined ? {} : { legacyTxHistory: previous.legacyTxHistory }),
```
agronmurtezi: **This carries `legacyTxHistory` across the crossing, but the first V2 sync update drops it.** `resolveCoinHashes` (line 351, outside this diff) rebuilds the wallet from a fixed field list (`state`, `publicKeys`, `networkId`, `coinHashes`, `progress`, `protocolVersion`) with no `legacyTxHistory`. `completeCrossing` calls it at the start of every V2 sync update (`v2/Sync.ts:658` and `:783`).

How it fails: restore a shielded 1.0.0 snapshot with `txHistory: [h1, h2]`, let the chain reach `forks.v9`, and apply one V2 sync update (an empty batch is enough). The next `serializeState()` has no `txHistory`, and the history is gone for good. Reproduced against the built packages with the real `facade-1.0.0` fixture: 2 entries after migration, `undefined` after the first sync update. A mid-crossing snapshot restored through `restoreWithPendingCoinHashes` loses it the same way.

The new migration and serialization tests all assert before any sync update, so none of them catch this.

packages/abstractions/src/InMemoryTransactionHistoryStorage.ts
```diff
@@ -91,21 +124,56 @@ export class InMemoryTransactionHistoryStorage<
...
-    const storage = new InMemoryTransactionHistoryStorage<T, Encoded>(schema, merge);
-    for (const entry of decoded) {
-      storage.#storage.set(entry.hash, entry);
-    }
-    return storage;
+  ): Either.Either<InMemoryTransactionHistoryStorage<T, Encoded>, TransactionHistoryRestoreError> {
+    return parseJson(serialized).pipe(
+      Either.flatMap(upgradeToCurrentFormat),
+      Either.flatMap(({ version, entries }) =>
+        Schema.decodeUnknownEither(Schema.Array(schema))(entries).pipe(
```
agronmurtezi: **Readers still drop unknown fields without a word.** This decode, all six snapshot `Serialization.ts` decoders, and `upgradeToCurrentFormat` (which forwards only `entries`) use Effect's default `onExcessProperty: 'ignore'`. `persisted-formats.md` says adding an optional field needs no version bump, so a payload from a newer SDK keeps a version label this build accepts, and the new field is gone on the next save.

How it fails: release N+1 adds optional field `f` with no bump. The app rolls back to this build, or two builds share storage. The payload decodes, is re-serialized without `f`, and N+1 later finds `f` gone with nothing reported. Checked against the built packages: V1 and V2 shielded, dust and unshielded snapshots with an extra key under their current label decode `Right` and re-serialize without it, and a v2 history loses extra envelope, entry and `finalizedBlock` keys. The "refuse unknown versions" rule never fires, because the version is known.

packages/facade/src/transaction.ts
```diff
@@ -72,7 +72,7 @@ const v9FinalizedTransactionTrait: PendingTransactions.TransactionTrait<ledgerV9
       Arr.flatMap((i) => i.dustActions?.spends ?? []),
       Arr.isNonEmptyArray,
     );
-    const hasShieldedOffers = tx.guaranteedOffer != null || (tx.fallibleOffer?.size ?? 0) == 0;
+    const hasShieldedOffers = tx.guaranteedOffer != null || (tx.fallibleOffer?.size ?? 0) > 0;
```
agronmurtezi: **Lower confidence, since no SDK path submits these today.** After the `> 0` change, a transaction with no intents and no zswap offers gets no TTL at all, so `hasTTLExpired` stays false forever; before, it expired after the grace period. A rewards or cNIGHT-bridge claim from `Transaction.fromRewards` has exactly that shape.

How it fails: a signed claim (`ClaimRewardsTransaction.new(..., 'CardanoBridge')` → `Transaction.fromRewards`) has `intents`, `guaranteedOffer` and `fallibleOffer` all undefined. Both traits return false even at creation + 10,000 days (the old code expired it at +3h). `queryForStatus` returns none for any `__typename !== 'RegularTransaction'`, so TTL is the only way out: a claim submitted through `facade.submitTransaction` stays in the persisted pending set, and "pending" in history, forever.

Separately, the new tests pin "shielded offers expire at creation + grace", which isn't a ledger rule: the grace period bounds dust `ctime`, and zswap roots are kept for `global_ttl` (1h).

packages/unshielded-wallet/src/SnapshotFormat.ts
```diff
@@ -0,0 +1,76 @@
...
+ * @param json The snapshot as decoded from its JSON string.
+ * @returns The same snapshot in the current format, or the input unchanged when the step does not apply.
+ */
+export const upgradeSnapshotV1ToV2 = (json: unknown): unknown => {
+  if (!isRecord(json)) return json;
+  if (isV1Keyed(json)) {
+    const { publicKey, ...rest } = json;
+    return {
+      ...rest,
+      version: SNAPSHOT_FORMAT_VERSION,
```
agronmurtezi: **This frozen v1→v2 step should stamp the literal `'v2'`, not `SNAPSHOT_FORMAT_VERSION`.** The constant is "the current version of the surface", so it moves with the next bump. Same at line 74, and `TransactionHistoryFormat.upgradeV1ToV2` stamps `CURRENT_FORMAT_VERSION` at `TransactionHistoryFormat.ts:145`.

How it fails: bump to v3 as `persisted-formats.md` describes, setting the constant to `'v3'` and adding `upgradeSnapshotV2ToV3` keyed on `version === 'v2'`. A stored v1 or unlabelled snapshot now leaves this step labelled `'v3'` but still v2-shaped, v2→v3 never runs, and the v3 schema refuses or mis-reads a snapshot the compatibility promise says must load. Reproduced by bumping only the constants.

packages/serialization-tests/test/snapshotCompat.test.ts
```diff
@@ -0,0 +1,183 @@
...
+      const { rewritten, reread } = reader.roundTrip(fixture);
+
+      expect(reread).toEqual(rewritten);
+    });
+  });
+});
+
+const unshieldedV1 = UnshieldedV1Serialization.makeDefaultV1SerializationCapability();
+const unshieldedV2 = UnshieldedV2Serialization.makeDefaultV2SerializationCapability();
+const unshieldedReaders = [
```
agronmurtezi: **The first frozen `unshielded/v2` fixture will turn this gate permanently red.** Every frozen fixture of a surface goes to both variants' readers and writers (here, in `informationPreservation.test.ts:94`, and in format drift through `sourceForBaseline` in `rewrite.ts`), but the V1 unshielded reader refuses `version: 'v2'` by design.

How it fails: adding the V2 writer's own output as `fixtures/unshielded/v2/facade-5.0.0.json` to a copy of the harness takes it from 343/343 passing to 5 failures, all on the V1 side (snapshotCompat ×3, informationPreservation, formatDrift). `yarn capture` fails the same way.

The documented generator can't produce such a fixture either: `generate.mjs` drives only `v1/Serialization.js` (lines 92, 195 and 279) and calls pre-5.0 APIs: `storage.upsert` (:370), a non-`Either` `restore().getAll()` (:374), and a 4-argument `addPendingTransaction` (:432).

packages/serialization-tests/test/snapshotCompat.test.ts
```diff
@@ -0,0 +1,183 @@
...
+});
+
+/**
+ * Assert a value against the fixture's own record of what it should be, when the fixture records it.
+ *
+ * Fixture kinds describe different things — a deep shielded snapshot records a coin count, a receiver snapshot records
+ * a balance — so a key absent from one is not a gap in that fixture, it is a key that kind does not carry. A key that
+ * _is_ present is always asserted; this never quietly skips one that exists.
+ */
+const expectRecorded = (fixture: Fixture, key: string, actual: () => unknown): void => {
```
agronmurtezi: **No test asserts the balances, coins or UTXOs the fixtures record.** `expectRecorded` is only called for `networkId`, `embeddedTxHistoryCount`, `address`, `appliedId` and `publicKey`, and returns silently for any other key. `persisted-formats.md` says to assert content ("balances, entries…").

How it fails: a regression restores a frozen `ZswapLocalState`, `DustLocalState` or UTXO set with a coin missing, or with `registeredForDustGeneration` flipped. `networkId` still matches, the round trip compares the writer's output with itself, and informationPreservation only sees key paths (the state is one hex string), so everything stays green. Only drift fails, and its remedy, `yarn capture`, re-records the loss.

Recorded keys that nothing reads: `coinValuesA`, `balanceA`, `balanceB`, `coinCount`, `availableValues`, `registeredFlags`, `dustUtxoCount`.

packages/abstractions/src/TransactionHistoryStorage.ts
```diff
@@ -35,9 +35,14 @@ export const FinalizedBlockSchema = Schema.Struct({
...
 
+/**
+ * A transaction that reached a block. `finalizedBlock` is optional because an entry restored from a history written
+ * before the lifecycle field existed is known to have been finalized, but carries no record of which block it landed
+ * in. No block is invented for those; a reader that needs one fetches it from the indexer by transaction hash.
+ */
 export const FinalizedLifecycleSchema = Schema.Struct({
   status: Schema.Literal('finalized'),
-  finalizedBlock: FinalizedBlockSchema,
+  finalizedBlock: Schema.optional(FinalizedBlockSchema),
```
agronmurtezi: **This makes `finalizedBlock` optional for every format and writer, not just for upgraded v1 entries.** Blockless finalized entries from any source are now accepted and persisted, look the same as upgraded ones, and nothing backfills them.

How it fails: before this PR, a JS or cast `gotFinalized` call without `finalizedBlock` made `serialize()` throw. Now it persists, and the simulator reports it as "restored from a pre-lifecycle history". A v2 payload with `{status: 'finalized'}` restores `Right`. Upgraded entries are never repaired either, because all three syncs resume after the snapshot's applied index or id, so for consumers `FinalizedWalletEntry.lifecycle.finalizedBlock` stays `| undefined` for good.

packages/abstractions/src/TransactionHistoryFormat.ts
```diff
@@ -0,0 +1,199 @@
...
+/** Names this persisted surface on a {@link TransactionHistoryRestoreError}, so one error type can serve all five. */
+export const TRANSACTION_HISTORY_SURFACE = 'transaction-history';
+
+/**
+ * Raised when a stored transaction history cannot be brought up to {@link CURRENT_FORMAT_VERSION} — a version this build
+ * does not know (written by a newer SDK), or a payload that fails to decode once upgraded.
+ *
+ * Never swallowed into an empty store: losing a user's history silently is worse than failing to open it.
+ */
+export class TransactionHistoryRestoreError extends Data.TaggedError('TransactionHistoryRestoreError')<{
```
agronmurtezi: **`TransactionHistoryRestoreError` has no `message`, and one tag covers four failure modes.** `.message` is `''` and `String(err)` is just the tag, which is less to go on than the `ParseError`/`SyntaxError` that `restore` used to throw.

How it fails, checked against the built package: `restore('{"version":"v3","entries":[]}')`, `'not json'` and a bad v1 entry each give a `Left` with `message === ""`. Logging `err.message` or `String(err)` shows nothing, `Effect.fail` prints "An error has occurred", and `EitherOps.getOrThrowLeft` throws "Unexpected left value" with the reason two causes deep. That's the opposite of what this PR does for dust.

The snapshot surfaces return a generic `WalletError.other`/`OtherWalletError` with the version only in the message text, so ADR 0008's "tagged error carrying the surface, the version detected and the cause" holds nowhere, and an app can't tell "written by a newer SDK" from "corrupt".

packages/serialization-tests/test/informationPreservation.test.ts
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Every key path in a payload, as dotted strings, with array elements collapsed to one path per position-independent
+ * shape. Comparing paths rather than values is deliberate: values legitimately change when a payload is re-encoded (a
+ * sync offset is re-derived, an optional is re-ordered), but a key that was there and is now gone is data loss.
+ */
+const keyPaths = (value: unknown, prefix = ''): readonly string[] => {
+  if (Array.isArray(value)) {
+    // One entry's shape stands for the array's: an empty array carries no shape to lose, and a ragged array would
+    // otherwise report a path as missing merely because a later element never had it.
+    return value.flatMap((item) => keyPaths(item, `${prefix}[]`));
```
agronmurtezi: **`keyPaths` merges every array element onto one `[]` path, so a field or element dropped from one entry goes unnoticed while another entry of the same array still has it.**

How it fails: on the real unshielded fixtures (two `state.availableUtxos`), dropping the second UTXO, or just its `meta.ctime`, gives `lost = []`. The same goes for dropping `transactions[1]` or its `creationTime` from a pending-transactions fixture. No other test catches either; only drift notices, and its remedy re-records the loss.

packages/shielded-wallet/src/v2/Serialization.ts
```diff
@@ -63,8 +66,26 @@ const StateFromUInt8Array = (): Schema.Schema<ledger.ZswapLocalState, Uint8Array
...
+ * the reader having to parse a schema tree.
+ *
+ * The version names the snapshot's shape, not the variant that wrote it, and the constant lives in
+ * `../SnapshotFormat.ts` so the twins cannot drift apart. Both variants write this shape — the V2 variant adds only
+ * optional fields to it, which is not a new version. A V1 reader never meets a V2 snapshot anyway: `../Restore.ts`
+ * routes each snapshot to the variant that owns its `protocolVersion`.
+ */
+const SnapshotVersionSchema = Schema.Literal(SNAPSHOT_FORMAT_VERSION).annotations({
+  message: (issue) =>
+    `Refusing a shielded snapshot written in format version ${JSON.stringify(issue.actual)}: this build reads ${SNAPSHOT_FORMAT_VERSION} and does not downgrade.`,
```
agronmurtezi: **This refusal never reaches callers of `ShieldedWallet.restore()`.** `restore` is `Either.getOrThrow(tryRestore(...))` (`ShieldedWallet.ts:555`, outside this diff), which throws a bare `Error('getOrThrow called on a Left')` with no cause. The same holds in `UnshieldedWallet.ts:460`, `DustWallet.ts:637` and the single-variant wallets, and dust's `restore` docs promise `@throws UnsupportedSnapshotVersionError`.

How it fails: a `version: 'v2'` shielded snapshot gives a `Left` carrying this refusal, but `Either.getOrThrow` throws "getOrThrow called on a Left" with `cause` undefined. An app following the README (`ShieldedWallet(configuration).restore(serializedState)`) learns neither the surface nor the version.

The `restore` code predates this PR; `Either.getOrThrowWith((e) => e)` would fix it.

packages/unshielded-wallet/src/v2/Serialization.ts
```diff
@@ -93,7 +105,11 @@ export const makeDefaultV2SerializationCapability = (): SerializationCapability<
     deserialize: (serialized): Either.Either<CoreWallet, WalletError> =>
       pipe(
         serialized,
-        Schema.decodeUnknownEither(Schema.parseJson(SnapshotSchema)),
+        // Parse, upgrade, then decode: the upgrade step is a pure function on the JSON and runs before any schema, so
+        // the schema describes exactly one shape and a `v1` payload arrives at it already in `v2`.
+        Schema.decodeUnknownEither(Schema.parseJson(Schema.Unknown)),
+        Either.map(upgradeSnapshotV1ToV2),
```
agronmurtezi: **Reading a V1-written snapshot straight into the V2 variant skips the cross-ledger migration.** Treating it as a pure format upgrade misses what the migration does: releasing booked UTXOs, setting `registeredForDustGeneration: false`, and, for shielded, releasing stranded spends through `coinHashesPending`.

How it fails, reproduced in the default two-variant build: V1 sync annotates `protocolVersion ≥ forks.v9`, the app serializes before the hand-over, and `tryRestore` routes the snapshot straight to V2's deserializer. The UTXO booked for a ledger-v8 transaction stays pending forever, the Night UTXO keeps `registeredForDustGeneration: true`, and shielded stranded spends are never released. The same happens in any V2-only build.

This was already possible through the old `Schema.Union`; this PR makes it the documented design.

.github/workflows/ci.yml
```diff
@@ -407,22 +407,42 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+
+      # No toolchain setup and no install: the check is a plain node script over `git diff`, using only the standard
+      # library, so the runner's own node runs it.
+      - name: Check published fixtures are unmodified
+        run: node packages/serialization-tests/check-frozen.mjs origin/${{ github.base_ref || 'main' }}
```
agronmurtezi: **The frozen-fixture gate does nothing outside `pull_request` runs, and a PR can get around it.**

- On a push to `main`, the base is `origin/main`, which is `HEAD`, so the diff is always empty.
- It runs the PR's own copy of `check-frozen.mjs`, so a PR that also loosens `FROZEN` passes. CODEOWNERS gives both files to the general wallet owners.
- `git diff --name-status` without `-z` C-quotes unusual paths, so the anchored `FROZEN` regex misses them.

Reproduced in a scratch repo: a push to `main` that edits `fixtures/shielded/v1/*.json` exits 0 (it exits 1 against `github.event.before`), and edits to `café.json` or `q"uote.json` exit 0. The job also does a full-history clone and waits on `general-checks` without needing either.

packages/shielded-wallet/src/v2/TransactionHistory.ts
```diff
@@ -233,21 +233,28 @@ export const makeSimulatorTransactionHistoryService = (
...
-                  height: entry.lifecycle.finalizedBlock.height,
-                  timestamp: entry.lifecycle.finalizedBlock.timestamp.getTime(),
-                },
-                status: entry.status ?? 'SUCCESS',
-                identifiers: entry.identifiers,
-              })
-            : Effect.fail(
-                new TransactionHistoryError({ message: `No transaction found in storage for hash: ${hash}` }),
-              ),
+          !isFinalized(entry)
```
agronmurtezi: **The new "finalized but records no block" branch is only tested in the V1 twins.** The identical V2 branches (here, and in dust's `v2/TransactionHistory.ts:254`) have no test, so a later edit to either V2 copy, such as swapped branches or a reworded message, passes CI.

The two failure modes also differ only by English text on the catch-all `TransactionHistoryError`, so a dApp told to fetch the block when it's missing has to string-match "records no block".

packages/unshielded-wallet/src/test/snapshotFormat.test.ts
```diff
@@ -0,0 +1,92 @@
+// The one upgrade step an unshielded snapshot has: `v1`, whose verifying key is a bare string, to `v2`, whose key
```
agronmurtezi: **This file is missing the Apache license header** that `.licenserc.yaml` requires for `**/*.ts`; its siblings start with `// This file is part of MIDNIGHT-WALLET-SDK.`. `license-headers.yml` only runs on pushes and PRs to `main`/`develop`, so this PR's CI can't catch it. It will fail the v2 → main merge PR, and `main` after that, until the header is added.

packages/serialization-tests/fixture-generator/generate.mjs
```diff
@@ -0,0 +1,583 @@
...
+    sh: 'sh-3.0.0',
+    un: 'un-3.0.0',
+    du: 'du-4.0.0',
+    shieldedFamily: 'withChanges',
+    abs: 'abs-2.1.0',
+    fa: 'fa-4.0.0',
+    cap: 'cap-3.3.0',
+  },
+  {
+    train: 'facade-4.1.0',
```
agronmurtezi: **This duplicates the cross-release snapshot corpus #740 already added to `v2`** (`scripts/cross-release-corpus/`, `packages/*/test/fixtures/cross-release/from-1.2.0/` and `packages/*/test/crossReleaseCorpus.test.ts`) for the same facade 4.1.0 release, without reusing or mentioning it.

The two already disagree on the recorded ledger (8.1.2 vs 8.1.0). The new gates only cover `packages/serialization-tests/fixtures/`, so the #740 fixtures stay unguarded, and ADR 0008 tells reviewers that nothing pinned released output before, which isn't so.

packages/dust-wallet/src/v2/Serialization.ts
```diff
@@ -88,7 +110,11 @@ export const makeDefaultV2SerializationCapability = (): SerializationCapability<
       return pipe(
         serialized,
         Schema.decodeUnknownEither(Schema.parseJson(SnapshotSchema)),
-        Either.mapLeft((err) => new OtherWalletError({ message: 'Error while deserializing snapshot', cause: err })),
+        // The parse detail is carried in the message, not only in the cause: a caller that logs `error.message` must
+        // still learn why the snapshot was refused, including the format version it was written in.
+        Either.mapLeft(
+          (err) => new OtherWalletError({ message: `Error while deserializing snapshot: ${err.message}`, cause: err }),
```
agronmurtezi: The dust deserializer now puts the whole parse-error tree in `message`. Shielded and unshielded already do this, so it's a nit.

packages/serialization-tests/fixture-generator/README.md
```diff
@@ -0,0 +1,73 @@
...
+  address against the key refuses.
+
+Never run it to make a failing compatibility test pass. If a frozen fixture fails to restore, the code is wrong, not the
+record — the CI frozen-file check exists precisely so the record cannot be quietly rewritten.
+
+## How to run it
+
+```bash
+cd packages/serialization-tests/fixture-generator
+npm ci
```
agronmurtezi: The fixture generator's lockfile resolves `@midnight-ntwrk/*` from `npm.pkg.github.com`, so `npm ci` fails without a token. The README doesn't mention it.

packages/serialization-tests/fixture-generator/generate.mjs
```diff
@@ -0,0 +1,583 @@
...
+  TOKEN_B,
+  MINT_A_VALUES,
+  MINT_B_VALUE,
+  TRANSFER_VALUE,
+  NIGHT_VALUE,
+  CUSTOM_UNSHIELDED,
+  CUSTOM_UNSHIELDED_VALUE,
+} from './chainDriver.mjs';
+
+const HERE = path.dirname(new URL(import.meta.url).pathname);
```
agronmurtezi: `new URL(import.meta.url).pathname` breaks on paths with spaces. The same line is in `chainDriver.mjs:24` and `probe.mjs:8`.

packages/serialization-tests/fixture-generator/probe.mjs
```diff
@@ -0,0 +1,52 @@
...
+        }
+      }
+    } catch (e) {
+      console.log(`  ${f}: FAILED TO IMPORT (${String(e.message).slice(0, 120)})`);
+    }
+  }
+};
+
+const wallets = process.argv[2] ? [process.argv[2]] : ['sh', 'un', 'du'];
+const trains = process.argv[3] ? [process.argv[3]] : ['t1', 't2', 't3', 't4', 't6'];
```
agronmurtezi: `probe.mjs` crashes with its default arguments. These train names build aliases like `sh-t1`, but the generator's aliases are version-named (`sh-1.0.0`), so reading the first `package.json` throws.

packages/serialization-tests/test/txHistoryCompat.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      expect(entries.map((entry) => entry.status)).toEqual(fixture.expected['statuses']);
+    });
+
+    it('should preserve the fees of every entry', async () => {
+      const entries = await restore().getAll();
+
+      const asWritten = (fixture.expected['fees'] as readonly (string | null)[]).map((fee) =>
+        fee === null ? null : BigInt(fee),
+      );
+      expect(entries.map((entry) => entry.fees ?? null)).toEqual(asWritten);
```
agronmurtezi: This fee assertion can't tell an explicit `null` from a missing field, since `entry.fees ?? null` turns both into `null`.

.changeset/facade-finalized-block-optional.md
```diff
@@ -0,0 +1,16 @@
+---
+'@midnightntwrk/wallet-sdk-facade': major
```
agronmurtezi: The PR title is `fix:` with no `!`, but this and `read-pre-lifecycle-tx-history` are major changesets.

packages/unshielded-wallet/src/v2/Serialization.ts
```diff
@@ -27,6 +30,22 @@ export type DefaultSerializationConfiguration = {
...
+ * refuses the payload and names both the surface it was reading and the version it found, so the failure is actionable
+ * without the reader having to parse a schema tree.
+ *
+ * This variant writes `v2`: its verifying key carries the signature scheme, where the V1 variant's is a bare string. A
+ * retyped field is a new format version, so a `v1` payload is upgraded in one step before this schema sees it
+ * (`upgradeSnapshotV1ToV2` in `../SnapshotFormat.ts`, which also holds both constants so the twins cannot drift). The
+ * V1 variant never meets a `v2` snapshot: `../Restore.ts` routes each snapshot to the variant that owns its
+ * `protocolVersion`.
+ */
+const SnapshotVersionSchema = Schema.Literal(SNAPSHOT_FORMAT_VERSION).annotations({
```
agronmurtezi: This `SnapshotVersionSchema` block is copied six times, once in each wallet's V1 and V2 `Serialization.ts`.

packages/facade/src/transaction.ts
```diff
@@ -139,7 +139,7 @@ const v8FinalizedTransactionTrait: PendingTransactions.TransactionTrait<ledgerV8
...
```
agronmurtezi: This `hasTTLExpired` body is a copy of the one at line 75. That duplication is why the TTL fix had to be made twice.

packages/abstractions/src/InMemoryTransactionHistoryStorage.ts
```diff
@@ -91,21 +124,56 @@ export class InMemoryTransactionHistoryStorage<
...
     merge?: (existing: T, incoming: T) => T,
-  ): InMemoryTransactionHistoryStorage<T, Encoded> {
-    const decode = Schema.decodeUnknownSync(Schema.Array(schema));
-    const decoded = decode(JSON.parse(serialized));
-    const storage = new InMemoryTransactionHistoryStorage<T, Encoded>(schema, merge);
-    for (const entry of decoded) {
-      storage.#storage.set(entry.hash, entry);
-    }
-    return storage;
+  ): Either.Either<InMemoryTransactionHistoryStorage<T, Encoded>, TransactionHistoryRestoreError> {
```
agronmurtezi: `InMemoryTransactionHistoryStorage.restore` now returns an `Either`. That's a major change, and it doesn't match the wallets, which pair a throwing `restore` with an `Either`-returning `tryRestore`.

PR #573, opened 2026-07-14
Fast-sync review fixes: nullifier matching, foreign spends, schema nullability

agronmurtezi:
Fixes for the 12 findings from the review of the fast-sync branch (7 correctness/API issues, 5 immutability-rule violations). Targets the feature branch (`js/fast-sync`, checked out locally as `js/fast-sync-v1`), not `main`.

## Correctness fixes

1. **Fixed-width nullifier matching** (`Sync.ts`, `Utils.ts`) — `leBigintToHex` produces SCALE-compact *minimal-length* bytes, while the indexer's `nullifierLeBytes` is documented as fixed *"32-byte little-endian form"*. Any nullifier with a zero top byte (~1/256 of them) never matched its own spend record, silently leaving the UTXO unspent locally and losing the successor (fee-change) UTXO. New `bigintToLeHex(n, byteLength)` produces the fixed-width form; the exact-match filter now uses it.
2. **Foreign dust spends no longer kill sync** (`Sync.ts`) — a matched transaction carries *every* dust spend it contains, including other parties' (multi-intent transactions, e.g. swaps). Unresolvable nullifiers previously raised `SyncWalletError`, permanently blocking sync on an on-chain transaction; they are now skipped with a debug log. A missing `genInfo` for an *owned* qdo still fails hard.
3. **Nullable merkle roots** (`SyncSchema.ts`) — `dustCommitmentMerkleTreeRoot`/`dustGenerationMerkleTreeRoot` are nullable in the indexer schema (generated type `string | null`); decoding now maps null to `''`, which matches the wallet-side encoding of an empty tree in the root check (and encodes back to null).
4. **BlockData timestamp encode** (`SyncSchema.ts`) — encode emitted `getTime() * 1000`; decode/encode now round-trips (unit-tested).
5. **Progress formula** (`Sync.ts`) — a misplaced parenthesis divided the settled-nullifier term by the phase ceiling, making it inert. Extracted as `nullifierPhaseProgress` with an explicit clamp to the phase ceiling.
6. **`blockData(height)` forwarding** — `makeEventLessSyncService` now passes the delegate function through instead of wrapping it and dropping the argument.
7. **`WalletFacade.doSync`** — dropped the unused `shieldedSecretKeys` parameter and the commented-out shielded call; added JSDoc for `start`/`doSync`; updated the e2e call sites.

## Immutability-rule fixes (CLAUDE.md hard requirements)

8–9. `CoreWallet.applyDustGenerations` / `applyDustCommitments` — `let` + `for` loops converted to folds; the no-collapsed-updates edge case collapses into the trailing insert.
10. Collapsed-commitment range grouping — `groups.push(...)` across if/else + `forEach` replaced by pure `gapRanges(lastAppliedIndex, maxIndex, skipIndexes)` (table-tested against the old implementation's outputs, including consecutive-skip and boundary cases).
11. `leBigintToHex` — `let` reassignment replaced with a const ternary; removed the dead odd-length guard (`Buffer.toString('hex')` is always even-length, and prepending `'0'` to LE hex would corrupt the value if it could ever fire).
12. `calculatePrefixLength` — `let` pipeline replaced with consts (existing test unchanged and green).

## Tests

- TDD per repo policy: the two key behavior tests were verified **red against the original code** (only those two fail, and the owned-spend positive control passes both before and after) before the fixes were finalized.
- New: fixed-width LE hex vectors incl. the zero-top-byte case; `gapRanges` table tests; `BlockDataSchema` round-trip + null-root decoding; a stubbed-subscription test proving exact-match filtering keeps zero-top-byte nullifiers and drops anonymity-set decoys; `createDustUtxoUpdates` foreign-spend skip + owned-spend resolution (spent + successor); `nullifierPhaseProgress` bounds; CoreWallet characterization tests (written against the loop implementation, green after the fold refactor).
- `yarn test:unit` for dust-wallet (58) and facade (32), `typecheck` (incl. e2e-tests), `lint`, and Effect LS diagnostics all pass.

## Decisions taken (flagged in the review as needing confirmation)

- **Null root ⇔ empty local tree compare equal** (both encode to `''`). If you'd rather fail sync when the indexer has no root but the local tree is non-empty, that needs its own error path.
- **Foreign spends: skip + `Effect.logDebug`** rather than surfacing a warning-level state.
- **Progress formula intent**: settled nullifiers count as a full commitment-space scan, live chains contribute their freshest successor's mt-index — my reconstruction; the unit tests encode it, adjust if the intent differed.

### DIFF
--- a/packages/dust-wallet/src/v1/test/sync.test.ts
+++ b/packages/dust-wallet/src/v1/test/sync.test.ts
@@ -22,7 +22,6 @@ import { DustLedgerEvents, DustNullifierTransactions } from '@midnightntwrk/wall
 import type {
   DustLedgerEventsSubscription,
   DustLedgerEventsSubscriptionVariables,
-  DustNullifierTransactionsSubscription as WireDustNullifierTransactionsSubscription,
   DustNullifierTransactionsSubscriptionVariables,
 } from '@midnightntwrk/wallet-sdk-indexer-client';
 import { type SubscriptionClient } from '@midnightntwrk/wallet-sdk-indexer-client/effect';
@@ -198,7 +197,7 @@ describe('V1 projections nullifier subscription', () => {
       recorded.value = variables;
       return Stream.fromIterable(
         [topByteSetHex, decoyHex, topByteZeroHex].map((hex) => ({ dustNullifierTransactions: wireRecord(hex) })),
-      ) as Stream.Stream<WireDustNullifierTransactionsSubscription, ClientError | ServerError, SubscriptionClient>;
+      );
     };
 
     const service = makeIndexerSyncService({
--- a/packages/dust-wallet/src/v1/test/sync.test.ts
+++ b/packages/dust-wallet/src/v1/test/sync.test.ts
@@ -10,20 +10,39 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-import { DustSecretKey, LedgerParameters } from '@midnight-ntwrk/ledger-v8';
+import {
+  DustSecretKey,
+  dustFirstNonce,
+  dustNullifier,
+  type Event as LedgerEvent,
+  LedgerParameters,
+} from '@midnight-ntwrk/ledger-v8';
 import { NetworkId } from '@midnightntwrk/wallet-sdk-abstractions';
-import { DustLedgerEvents } from '@midnightntwrk/wallet-sdk-indexer-client';
+import { DustLedgerEvents, DustNullifierTransactions } from '@midnightntwrk/wallet-sdk-indexer-client';
 import type {
   DustLedgerEventsSubscription,
   DustLedgerEventsSubscriptionVariables,
+  DustNullifierTransactionsSubscription as WireDustNullifierTransactionsSubscription,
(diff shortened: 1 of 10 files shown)


packages/dust-wallet/src/v1/test/sync.test.ts
```diff
@@ -151,3 +170,178 @@ describe('V1 projections sync capability', () => {
...
+    const topByteSetHex = '12' + '00'.repeat(30) + '44';
+    const topByteZeroHex = '0c' + '00'.repeat(29) + '01' + '00';
+    const decoyHex = '12' + 'ff'.repeat(30) + '44'; // anonymity-set member sharing topByteSet's prefix
+
+    const recorded: { value?: DustNullifierTransactionsSubscriptionVariables } = {};
+    const stub = (variables: DustNullifierTransactionsSubscriptionVariables) => {
+      recorded.value = variables;
+      return Stream.fromIterable(
+        [topByteSetHex, decoyHex, topByteZeroHex].map((hex) => ({ dustNullifierTransactions: wireRecord(hex) })),
+      ) as Stream.Stream<WireDustNullifierTransactionsSubscription, ClientError | ServerError, SubscriptionClient>;
```
kapke: why it type casting needed?

kapke (approved):

Issue #577, opened 2026-07-16
chore: align effect-language-service configuration and re-enable deterministicKeys


iangregsondev:
Feature Request: The `@effect/language-service` configuration and the codebase currently disagree: `main` has ~22 pre-existing `deterministicKeys` violations, so repo tooling temporarily runs with that diagnostic downgraded to "off". Decide the configuration the team actually wants, make code and config agree, and re-enable full enforcement.

## Describe the proposed feature:

1. **Review the ELS configuration** (`tsconfig.base.json` plugin block) and agree on what "good" looks like for this repo — in particular the `keyPatterns` that drive `deterministicKeys`.
2. **Resolve the ~22 pre-existing violations** (survey of 4 packages: utilities 7, shielded-wallet 8, dust-wallet 5, capabilities 2 — run a full repo sweep first). Two options, team decision required:
   - **Option A — fix the code:** rename tags to the deterministic keys (e.g. `'LeftError'` → `'@midnightntwrk/wallet-sdk-utilities/EitherOps/LeftError'`). ⚠️ **Breaking:** `_tag` is a runtime value; consumers using `catchTag`/`_tag` matching silently stop matching. Needs a compat assessment and (likely major) changeset.
   - **Option B — fix the config:** adjust `keyPatterns` so the existing short tags are compliant. Zero runtime change, no breakage — but locks in the current naming convention.
3. **Remove the temporary override** in `scripts/els-changed.mjs`, so all ELS tooling enforces the full configuration.

## User Need:

Developers and coding agents need trustworthy, fully-enabled diagnostics. A rule that is "error" in config but suppressed in tooling is worse than either enforcing or dropping it — it drifts, and nobody knows which convention is real.

## Expected Benefit:

- `deterministicKeys` enforced everywhere with zero suppressed rules
- A deliberate, documented decision on error-tag naming instead of an accidental one
- No risk of a coding agent "fixing" tag names unprompted (the reason the override exists today)

## Details:

Discovered while integrating ELS diagnostics into the repo tooling (#576). The override in `scripts/els-changed.mjs` is built dynamically from `tsconfig.base.json` with only `deterministicKeys` downgraded — deleting that one block is this ticket's definition of done.

Side-finding worth folding into the config review: `importAliases` violations do **not** surface via CLI diagnostics (editor/quickfix-tier only), so hook/CI cannot currently enforce that rule.

[...]

PR #578, opened 2026-07-16
docs: restructure CLAUDE.md and Claude Code config

### LINKED ISSUE #576: docs: restructure CLAUDE.md — slim always-loaded core, extract conventions, add path-scoped rules
Documentation Improvement: CLAUDE.md has grown to a size where it dilutes its own instructions; restructure it into a lean always-loaded core plus conditionally-loaded conventions.

### Documentation URL:

- [`CLAUDE.md`](../blob/main/CLAUDE.md) (repo root)
- Related prior art: #252

### Description of Improvement:

`CLAUDE.md` is currently **1,117 lines / ~10,600 tokens**, loaded into every Claude Code session. Over half of it (~5,600 tokens) is the *Functional Programming Principles* section — tutorial-style content with worked examples and idiom-translation tables. Long always-loaded context measurably degrades instruction adherence: the hard rules (immutability, Either vs Effect, test suffixes, TDD gate) get lost in reference material, and agents "forget" them mid-session.

PR #252 (March 2026) identified the same problem and cut the file to 281 lines, but it predates two things:

1. **Claude Code `.claude/rules/` with `paths:` frontmatter** (v2.1.198+) — rules files that load *only when files matching a glob are touched*. This is strictly better than #252's soft "see FUNCTIONAL_PROGRAMMING.md" pointer, which relies on the agent choosing to read it.
2. **Four months of CLAUDE.md changes** — the unit/integration test split (#500), the `@midnightntwrk` scope rename, shelf, Common Gotchas. #252 would clobber all of these and reintroduces dead references (GitNexus, `input-output-hk` URLs).

There is also an internal inconsistency to fix while here: contributors' personal global configs carry conventions (commit format, branch naming) that belong in the repo — and can even *conflict* with it (e.g. `DEV_GUIDE.md` mandates `feat/*` branch names while a common personal convention is hyphen-only). The repo should be the single source of truth for team conventions.

### Tasks

Implementation: PR #578. The notes under each task record decisions and findings made along the way.

1. **Create `docs/CodingConventions.md`** — move the FP Principles content out of `CLAUDE.md` verbatim (mechanical extraction, no transcription); add an intro linking ADR 0004 / ADR 0006 / `Design.md` / effect.website; re-verify every cited canonical file against `main`.
   - Fixed in the move: the Either/Effect "when to use which" table had been mangled into unrenderable run-on prose (restored as a real table); dead `packages/dapp-connector-reference/src/parsing.ts` references (package no longer exists) repointed to `packages/address-format/src/index.ts`. Every other cited file verified to exist (`UnshieldedState.ts`, `pendingTransactions*.ts`, `ProtocolVersion.ts`, `EitherOps.ts`, `StateChange.ts`, …).
   - Verdict on #252's draft: current CLAUDE.md is a strict superset — nothing to mine beyond the table formatting; it shared the same dead reference.

2. **Create `.claude/rules/functional-style.md`** — distil the hard FP rules into a path-scoped rule (`paths: ['packages/**/*.ts']`, ~55 lines) loading only when SDK code is touched; point to `docs/CodingConventions.md` for rationale (pull-on-demand link, not auto-loaded).
   - `.claude/rules/` mechanics verified against official docs: real feature (v2.1.198+); rules **with** `paths:` globs lazy-load only when matching files are read/edited, rules without load always.
   - Covers: immutability (no `let`/loops/mutation), Either (pure) vs Effect (I/O) with boundary conversion, parse-don't-validate + branded types + tagged unions + total functions, no unjustified casts, Effect import aliases, gen-vs-pipe, tagged-error discipline, ref-update discipline, canonical-files pointer list.

3. **Create `.claude/rules/testing.md`** — test rules loading only when tests are touched: `paths: ['**/*.test.ts', '**/vitest.config.ts', 'packages/**/test/**', 'packages/e2e-tests/**']` (the `test/**` glob added beyond the original plan so helper/fake files trigger too).
   - Covers (~49 lines): unit/integration/e2e suffix routing (binary decision: needs infra or not), never-mix, the unit-project `exclude` gotcha, run-from-root; the TDD contract compressed to its core (observe red for the expected reason, user gate, **never weaken a failing test**, escalation path) with the full loop delegated to the `tdd` skill; fakes-over-`vi.fn` + `Layer.succeed` for Effect services; Effect test unwrapping (`runPromise`, `runPromiseExit` + `Exit`).
   - CI-matrix/`.env` details deliberately excluded — reference material, not rules; moved to `DEV_GUIDE.md` (task 5).

4. **Create `.claude/rules/transactions.md`** — transaction domain knowledge, loaded only in transaction-handling packages.
   - Paths verified by usage (grep for `imbalances`/`DustActions`/`Intent`): the three wallet packages, `capabilities`, `facade`, `docs-snippets`, `wallet-integration-tests`, `e2e-tests`; `prover-client`/`node-client` excluded — they treat transactions as opaque payloads.
   - Content: ledger type + spec pointers, key facts (type parameters, segments, imbalances, DustSpend, Dust-is-not-a-token), working-example pointers.
   - Stale claims fixed in the move: `ledger-v7` → `ledger-v8` (v7 no longer installed); dead `dapp-connector-reference` test refs replaced with verified successors (`shielded-wallet` transacting tests, `capabilities` Balancer tests). CLAUDE.md's Transaction Inspection section fully relocated here.

5. **Slim `CLAUDE.md` to ~250 lines / ~2,500 tokens** — every cut must have a named successor.
   - Landed at 262 lines (from 1,089). Successors: FP Principles → `docs/CodingConventions.md` + `functional-style.md` rule (6-line anchor kept); TDD → `testing.md` rule + `tdd` skill (4-line anchor kept, since rules only fire on test-file touch); Effect test patterns + unit/integration detail → `testing.md`; CI-matrix prose → `DEV_GUIDE.md` (new "Testing Tiers & CI" section); Key Specifications → 3-row table + note that shelf caches the spec repos locally; ELS section compressed (per-edit diagnostics instruction dropped — the Stop hook does it mechanically); Versioning → the `changeset` skill; JSDoc worked example → `docs/CodingConventions.md` (new Documentation Standards section, requirements kept in core).
   - Added: Documentation Map table (incl. the three rules files) and the Downstream Impact table adopted from #252 (person name replaced with "coordinate before merging"). `docs/` naming reviewed and deliberately kept PascalCase (matches pre-existing `Design.md`); root stays UPPER_SNAKE per ecosystem convention.

6. **Move team conventions into the repo** — the repo is the single source of truth; personal global configs shouldn't carry (or contradict) them.
   - Git & GitHub conventions section added to repo `CLAUDE.md` (never commit to `main`, Conventional Commits, attended commits/pushes/PRs, use `.github/` templates).
   - `docs/ClaudeCode.md` added: settings-file scopes + recommended personal `deny`/`ask` permission blocks — linked from `DEV_GUIDE.md`, deliberately *not* from `CLAUDE.md` (human setup info doesn't belong in agent context).
   - Contributors' personal global configs slimmed to genuinely-personal preferences.
   - `DEV_GUIDE.md` branch naming aligned with actual practice: strongly recommended (not enforced) pattern `<type>/<ticket>-<short-description>`.

7. **Implement TDD as a skill** (suggested by @kapke) — the rules-file TDD mandate (task 3) states the contract; the skill walks the loop.
   - `tdd` skill in the `wallet-sdk` plugin: choose test type (unit vs integration suffix) → design → write → **observe red** (for the expected reason) → **user gate** (review/commit tests before implementation) → implement → **observe green** (same `yarn test --filter` command) → review (`verify:changed`) → consider refactor → next item; plus the never-weaken escalation path.

8. **Reconcile `SKILLS.md` with reality** — it documented ~10 slash commands, none of which existed (`.claude/skills/` contained only `shelf`, and nothing linked to `SKILLS.md`).
   - **Deleted.** Triage: `/test` `/build` `/format` `/verify` were pure yarn aliases; `/pre-push` referenced a non-existent `yarn check` (superseded by `verify:changed` + the Stop hook); `/check-fp` superseded by ESLint + ELS hook + the task-2 rule; `/new-error` and `/inspect-state` too trivial to be skills.
   - The one real find: **`changeset`** — `yarn changeset add` is an interactive CLI an agent cannot drive, so the skill writes the changeset file directly (package selection, SemVer bump judgment, empty-changeset case, `changeset:check`).
   - Skills packaged as the **`wallet-sdk` plugin** (`.claude/skills/wallet-sdk/` — manifest in `.claude-plugin/plugin.json`, skills under `skills/`, one shared namespace). Auto-enable confirmed against official docs: skills-dir plugins load automatically after workspace trust, no marketplace and no settings entry needed. Live-tested: skills appear in the picker by short name (`/tdd`) with a `(wallet-sdk)` label; `/wallet-sdk` fuzzy-lists the plugin's skills; there is **no** `wallet-sdk:name` colon input syntax — docs must never advertise one (the `audit` skill polices this).
   - Gotcha found live: plugin loading is gated on `hasTrustDialogAccepted`, which known bugs (anthropics/claude-code#36403, #9113) can leave `false` — and for git worktrees it's read from the main checkout's entry. Documented with fix in `docs/ClaudeCode.md`.

9. **Turn mechanically-checkable prose rules into hooks** — every rule that can be enforced mechanically should be a hook (or CI/permission gate), not always-loaded CLAUDE.md text.
   - PostToolUse: replaced the check-only `format:check:root` hook (exit 0 — the agent never saw failures) with `yarn format:changed`, which auto-**fixes** all changed files after each `Edit`/`Write` (~0.5 s).
   - Stop: `scripts/claude/hook-stop-verify.mjs` runs `yarn verify:changed` when the agent tries to end its turn; on failure blocks the stop (exit 2) and feeds the errors back so the agent fixes its own work, with a `stop_hook_active` guard so an unfixable error surfaces to the user instead of looping forever.
   - Root scripts added: `changed-files`, `format:file`, `format:changed`, `lint:changed` (= `turbo run lint --affected`: builds `^dist` + `typecheck` first, cached), `verify:changed` (format → lint; format first because `eslint-plugin-prettier` reports unformatted code as lint errors).
   - ELS adopted: `effect-language-service diagnostics` **works without built `dist/`** (its rules resolve types from the `effect` package itself — no phantom errors in fresh worktrees; `floatingEffect` verified firing), proper exit codes, ~1.5–2 s per file. `yarn els:changed` (`scripts/els-changed.mjs`) chained into `verify:changed`, so the Stop hook enforces it. Caveats: `deterministicKeys` downgraded to "off" inside the script (pre-existing violations on `main` — cleanup tracked in #577); `importAliases` violations don't surface via CLI diagnostics (editor/quickfix-tier only), so the hook doesn't cover that rule.
   - Why lint/typecheck run on `Stop`, not per edit: turbo caching cannot help the package being edited (the edit invalidates its cache), so a per-edit hook would pay a full `tsc` + type-aware ESLint run on every change — and mid-turn code is often legitimately half-finished (red TDD test, multi-file refactor in progress). Naive single-file eslint in an unbuilt worktree also yields ~100 phantom `no-unsafe-*` errors.

10. **Create the claude-config `audit` skill** (`wallet-sdk` plugin) — agent-run, human-invoked sweep of the repo's Claude Code config layer for rot.
    - Mechanical checks: dead file refs in CLAUDE.md/rules/skills/docs, rule `paths:` globs matching no tracked files, hook commands/yarn scripts that don't resolve, version claims vs `package.json`, slash-name references vs real skills, CLAUDE.md size budget (~250 lines), formatter-mangled markdown. Judgment checks: rules-vs-docs drift, stale gotchas/tables. Report-only — fixes on approval. No script, pure checklist skill.
    - Rationale: every fix in this ticket was exactly this kind of rot, and nothing else checks the config layer (hooks verify code, not config).

11. **Final pass** — run all four skills through Matt Pocock's `writing-great-skills` skill, then verify everything end-to-end in fresh sessions.
    - Skill review applied: `tdd` — trigger-synonym collapse in the description, completion criterion added to the "write" step; `changeset` — removed the inline copy of the ignored-packages list (it had **already drifted** from `.changeset/config.json`, now deferred to as the single source of truth); `audit` — description halved; `shelf` (vendored) — prettier had reflowed its escaped code fences into run-on lines that **taught wrong one-line shelffile syntax**, rewritten with real fences (formatter-stable) and trimmed 103 → ~55 lines. The cleaned `shelf` copy doubles as the seed for upstreaming to Rika-Labs/shelf (out of scope here).
    - Manual end-to-end verification: **complete** — rules trigger on matching files (each rule confirmed individually, including a behavioral refusal test), hooks behave (format-on-edit, Stop-hook block + feedback), skills fire on natural phrasing. Full method and results: [verification comment](https://github.com/midnightntwrk/midnight-wallet/issues/576#issuecomment-4991577367).

### Decisions

- **Team-wide `ask` gates: no.** A developer should not be dictated how they work; the commit/push/PR/issue `ask` gates stay a **recommendation**, documented in `docs/ClaudeCode.md` (personal `settings.json`). The tracked `.claude/settings.json` keeps its "no `permissions`" rule.
- **Plugin over plain project skills.** Condition was that plugin skills behave like standard skills for the agent — confirmed: same available-skills listing, same `description`-driven triggering. The standalone-plugin layout (one plugin, `skills/` folder inside) was chosen over per-skill manifests for a single shared namespace. Accepted cost: one manifest to maintain.
- **Explicitly not carried over from #252:** the "Key People" reviewer table (goes stale, was already flagged incomplete in review), the GitNexus instruction (tooling no longer present), and `scripts/review-patterns.ts` (speculative).

**Net effect:** always-loaded context drops ~10.6k → ~2.5k tokens, and conventions load deterministically exactly when code/tests are being edited.

iangregsondev:
# Description

Restructures the repository's Claude Code setup: the always-loaded `CLAUDE.md` shrinks from 1,089 lines (~10.6k tokens) to 262 (~2.5k), and every convention it used to carry moves to a home where it loads deterministically — path-scoped rules when matching files are touched, skills when a workflow is invoked, hooks that enforce mechanically instead of asking the agent to remember. Task-by-task spec, findings, and decisions: #576 (to be closed manually after QA).

# Proposed Solution

| Layer | What changed |
| --- | --- |
| `CLAUDE.md` | Slim always-loaded core: overview, documentation map, build commands, architecture, spec pointers, downstream-impact table, gotchas, team git/GitHub conventions |
| `.claude/rules/` | Path-scoped rules loading only when matching files are touched: `functional-style.md` (SDK code), `testing.md` (tests), `transactions.md` (transaction-handling packages) |
| `docs/` | `CodingConventions.md` — full FP conventions moved verbatim, plus documentation standards; `ClaudeCode.md` — human setup: recommended personal permissions, hooks/scripts reference, plugin troubleshooting |
| `wallet-sdk` plugin | `.claude/skills/wallet-sdk/` with `tdd`, `changeset`, and `audit` skills; phantom `SKILLS.md` deleted (documented ten commands that didn't exist) |
| Hooks | Tracked `.claude/settings.json`: auto-format changed files after every edit; on Stop, `yarn verify:changed` (format → lint → Effect diagnostics) blocks the agent until its own work passes |
| Root scripts | `changed-files`, `format:file`, `format:changed`, `lint:changed`, `els:changed`, `verify:changed` — used by hooks and humans alike |
| `DEV_GUIDE.md` | Branch naming softened to a strong recommendation; new "Testing Tiers & CI" section (moved from CLAUDE.md) |

Stale content was fixed during relocation rather than copied: `ledger-v7` → `v8`, dead `dapp-connector-reference` paths, a mangled Either/Effect table, formatter-broken examples in the vendored `shelf` skill.

- `deterministicKeys` is temporarily downgraded to off inside `scripts/els-changed.mjs` — pre-existing violations on `main`; cleanup tracked in #577
- The tracked `.claude/settings.json` carries hooks only; permissions stay personal (recommendations documented in `docs/ClaudeCode.md`)
- `SKILLS.md` is deleted — see #576 task 8 for the per-command triage

# Testing

- Hooks and scripts exercised live with planted failures: the Stop hook blocks on exit 2 and feeds errors back until fixed; ELS catches `floatingEffect`; `yarn changeset:check` passes
- Manual end-to-end verification in fresh sessions is **complete** — rules trigger on exactly the matching files (including a behavioral refusal test), hooks behave, skills fire on natural phrasing: [full results on #576](https://github.com/midnightntwrk/midnight-wallet/issues/576#issuecomment-4991577367)
- To see it yourself: open a fresh Claude Code session on this branch and read any `packages/**/*.ts` file — `functional-style.md` injects; touch a `*.test.ts` path — `testing.md` injects
- If plugin skills don't appear in the `/` picker: trust-flag gotcha — see `docs/ClaudeCode.md` → troubleshooting


.claude/skills/shelf/SKILL.md
```diff
@@ -1,102 +1,56 @@
 ---
 name: shelf
```
kapke, line 2: This to make sense in the repo should be accompanied by installing shelf as a dev dependency in the workspace root, and the commands adjusted to be run with `yarn` prefix. Otherwise we would make a step in the direction of "one needs to spend a week to be able to do anything useful with the repo".
It's a useful tool and skill though, so let's keep it, just make sure shelf is installed with the project.

.claude/skills/wallet-sdk/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,8 @@
+{
```
kapke, line 1: Why creating a plugin for some of these and not the rest?

iangregsondev: This is because the others happened before the plugin. 

I will look at moving others into that. 

I think there was shelf skill but that is not ours. 

I introduced plugins to kind of namespace skills. 

I will look into it.

.claude/skills/wallet-sdk/skills/tdd/SKILL.md
```diff
@@ -0,0 +1,74 @@
...
+- Design so the test can be implemented **without later modification**.
+
+## 2. Write the test
+
+One behavior, precise assertions — exactly what was designed in step 1.
+
+## 3. Observe RED
+
+```bash
+yarn test --filter=@midnightntwrk/wallet-sdk-<package> -- test/<File>.test.ts
```
kapke: File name is enough:
```suggestion
yarn test --filter=@midnightntwrk/wallet-sdk-<package> --<File>.test.ts
```

.claude/skills/wallet-sdk/skills/tdd/SKILL.md
```diff
@@ -0,0 +1,74 @@
...
+description: >
+  Test-driven development loop for the Midnight Wallet SDK. Use whenever implementing new behavior or fixing a bug in
+  packages/**, or on any mention of TDD — the test is written, observed failing, and approved by the user BEFORE any
+  implementation. Not needed for docs, config, or pure refactors with existing coverage.
+---
+
+# TDD Loop — Wallet SDK
+
+**The test is the specification.** Once a test fails for the right reason, it must not be weakened to accommodate the
+implementation. Work one test at a time.
```
kapke:
```suggestion
implementation. Consult the user for the feedback loop size. It might be one test at a time, but also a full suite prepared ahead of time. It all depends on context and the task at hand. If unsure - prefer less tests in one cycle. 
```

docs/CodingConventions.md
```diff
@@ -0,0 +1,665 @@
...
+});
+```
+
+**Parallel** - use when operations are independent:
+
+```typescript
+// Both run concurrently
+Effect.all([fetchA(), fetchB()], { concurrency: 'unbounded' });
+
+// Or with Do notation for named results
```
kapke: These are not concurrent - this is equivalent to what the `.gen` example does above.

scripts/els-changed.mjs
```diff
@@ -0,0 +1,40 @@
...
+import { execFileSync, spawnSync } from 'node:child_process';
+import { readFileSync } from 'node:fs';
+
+const changed = execFileSync('node', ['scripts/changed-files.mjs'], { encoding: 'utf8' })
+  .split('\n')
+  .filter((file) => /^packages\/.*\.tsx?$/.test(file));
+
+if (changed.length === 0) process.exit(0);
+
+const baseConfig = JSON.parse(readFileSync('tsconfig.base.json', 'utf8').replace(/^\s*\/\/.*$/gm, ''));
```
kapke: Why not using json5 or similar parser, which understands comments?

.claude/rules/transactions.md
```diff
@@ -0,0 +1,42 @@
...
+  - 'packages/e2e-tests/**'
+---
+
+# Transactions — domain knowledge
+
+When building or inspecting transactions, consult the spec — don't guess protocol semantics.
+
+## Type and spec sources
+
+- **Ledger types**: `node_modules/@midnight-ntwrk/ledger-v8/ledger-v8.d.ts` — `Transaction`, `Intent`, `ZswapOffer`,
```
agronmurtezi: we need to make this version agnostic if possible

agronmurtezi (dismissed):
solid work!

scripts/els-changed.mjs
iangregsondev: Thank you, this has now been replaced with ts.readConfigFile which is JSONC aware.

Thanks!

docs/CodingConventions.md
iangregsondev: This has been fixed.

.claude/rules/transactions.md
iangregsondev: I have done that, it needed to support both orgs too (one with the dash and one without)


Package | Latest | Published
-- | -- | --
@midnight-ntwrk/ledger-v8 | 8.1.0 (+ 6 versions) | 2026-05-13
@midnightntwrk/ledger-v8 | 8.2.0-rc.1 only | 2026-07-10

@agronmurtezi - this should cover it!

.claude/skills/wallet-sdk/.claude-plugin/plugin.json
iangregsondev: So, things have changed a little, we have some company wide shareable skills but I think we are early!

So, what I propose is that we get this merged and I will try and work with the other repo to get some of our skills added over there.

I actually have some improved skills that have been generalized, so there is no mention of wallet etc - these can land in the company skills repo and then we can pull them in.

@kapke  ok ? 

I will create a ticket for it.

.claude/skills/shelf/SKILL.md
iangregsondev: So, shelf is actually a binary, I think the reason it appeared to be modified  is a mistake. I am going to revert this. I will create another PR to look at what we can do here. Shelf has been in the repo for a while.

It can't be added as a devDependency because it is actually a executable.

What I suggest we do is leave it right now, and I will open a new dedicated PR to fix and improve some of this.

@kapke

.claude/skills/wallet-sdk/.claude-plugin/plugin.json
iangregsondev: Currently, any skills you notice that are not via the plugin is because they are 3rd party skills that were not available as plugins.

agronmurtezi (approved):

Issue #608, opened 2026-07-27
chore: consolidate Renovate PRs into grouped weekly updates


iangregsondev:
## Describe the proposed feature:

Restructure `renovate.json` `packageRules` so Renovate opens a small, predictable
set of grouped PRs each week instead of one PR per dependency.

Group **by update type** rather than by package manager:

- **All non-major** (patch + minor) → a single PR, spanning npm, GitHub Actions, and Docker
- **All majors** → a single PR
- **All Midnight packages/images** (`@midnight-ntwrk/**`, `ghcr.io/midnight-ntwrk/**`) → a single PR

Also remove the redundant `{ "matchManagers": ["npm"], "enabled": true }` rule — npm is
already enabled by default via `config:best-practices`, so it is a no-op.

## User Need:

Renovate is currently opening a large volume of individual PRs (21 open at time of
writing), each bumping a single dependency. This creates review fatigue and noise, and
makes it hard to see the wood for the trees. Maintainers want the *same* dependency
coverage delivered as a handful of reviewable PRs.

## Expected Benefit:

**Roughly 21 → 3–4 open Renovate PRs per week.**

Breakdown of the current 21 open PRs under the new grouping:

| Group | Example PRs today | Result |
| --- | --- | --- |
| All dependencies (non-major) | `@effect/platform`, `undici`, `shell-quote`, `protobufjs`, `graphql-ws`, `effect`, `tar`, `postcss`, `bn.js` 5.2.5, `setup-node` v6.5.0, `docker/login-action`, `checkout`, existing `devdependencies (non-major)` (~13) | **1 PR** |
| All dependencies (major) | `commander` v15, `@noble/curves` v2, `picomatch` v4, `js-yaml` v5, `bn.js` v5, `setup-node` v7 (~6) | **1 PR** |
| Midnight libraries and docker images | (none currently open) | **1 PR when present** |
| Ungrouped | `lock file maintenance`, `pin yaml` | as-is |

Grouping **by update type** (not manager) means the non-major PR automatically sweeps up
npm, GitHub Actions, and Docker bumps — no per-manager rules needed. Every update is
either non-major or major, so it always lands in one of the two type buckets (or the
Midnight bucket).

### Final proposed config

```json
{
  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
  "nix": {
    "enabled": true
  },
  "extends": ["config:best-practices", "schedule:weekly", ":disableRateLimiting"],
[...]

PR #609, opened 2026-07-27
chore(deps): group renovate PRs by update type

iangregsondev:
# Description

Restructures `renovate.json` `packageRules` so Renovate opens a small, predictable set of
grouped PRs each week instead of one PR per dependency (~21 open today → ~3–4 expected).

Refs #608

# Proposed Solution

Group **by update type** rather than by package manager:

- **All non-major** (patch + minor) → one PR, spanning npm, GitHub Actions, and Docker
- **All majors** → one PR
- **All Midnight packages/images** (`@midnight-ntwrk/**`, `ghcr.io/midnight-ntwrk/**`) → one PR

The Midnight rule is placed **last** because Renovate's last matching `groupName` wins, so
Midnight bumps always group together and never leak into the generic non-major/major buckets.

Also removed the redundant `{ "matchManagers": ["npm"], "enabled": true }` rule — npm is
already enabled by default via `config:best-practices`, so it was a no-op.

**Trade-off (accepted):** collapsing all majors into one PR means that PR will sometimes be
red — a breaking major blocks the others behind it until fixed or split out. For a
low-traffic SDK repo this is an acceptable cost for the reduced noise.

Config-only change; no code or runtime behaviour affected. Empty changeset added (non-release).

# Testing

After merge, verify against the Dependency Dashboard that open PRs collapse to the expected
groups.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -16,17 +16,16 @@
   ],
   "packageRules": [
     {
-      "matchManagers": ["npm"],
-      "enabled": true
+      "groupName": "All dependencies (non-major)",
+      "matchUpdateTypes": ["patch", "minor"]
+    },
+    {
+      "groupName": "All dependencies (major)",
+      "matchUpdateTypes": ["major"]
     },
     {
       "groupName": "Midnight libraries and docker images",
       "matchPackageNames": ["@midnight-ntwrk/**", "ghcr.io/midnight-ntwrk/**"]
-    },
-    {
-      "groupName": "devDependencies (non-major)",
-      "matchDepTypes": ["devDependencies"],
-      "matchUpdateTypes": ["patch", "minor"]
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -16,17 +16,16 @@
...
-      "matchPackageNames": ["@midnight-ntwrk/**", "ghcr.io/midnight-ntwrk/**"]
+      "groupName": "All dependencies (major)",
+      "matchUpdateTypes": ["major"]
     },
     {
-      "groupName": "devDependencies (non-major)",
-      "matchDepTypes": ["devDependencies"],
-      "matchUpdateTypes": ["patch", "minor"]
+      "groupName": "Midnight libraries and docker images",
+      "matchPackageNames": ["@midnight-ntwrk/**", "ghcr.io/midnight-ntwrk/**"]
```
agronmurtezi: we should add @midnightntwrk here too now

iangregsondev: good catch!!

agronmurtezi (approved):

PR #617, opened 2026-07-29
feat(dust-wallet): projections-based fast sync (v2 port)

agronmurtezi:
# Dust fast-sync (projections-based) — v2 port

This is the **v2 port of #318** ("Fast Dust Wallet Synchronization"). #318 stays open against `main` and is untouched; this PR merges its head (`js/fast-sync` @ c040534d) into `v2` and adapts it to the ledger-v9 toolchain.

## Adaptations made for v2

- **Ledger imports**: all fast-sync code moved from `@midnight-ntwrk/ledger-v8` to `@midnightntwrk/ledger-v9`. Verified by published-typings diff before the port: the dust API surface (dustNullifier, successorDustUtxo, dustFirstNonce, DustStateChanges, DustLocalState, DustStateMerkleTreeCollapsedUpdate, DustGenerationTreeInsertionPath, QualifiedDustOutput, LedgerParameters, Event) is identical between ledger-v8@8.2.0-rc.1 and ledger-v9@1.0.0-rc.3, so this is import renames only — no logic changes.
- **Async signers (one code adaptation)**: `projectionsBasedSync.undeployed.test.ts` passed sync `(payload) => keystore.signData(payload)` callbacks; v2's API expects async signers. Switched to `keystore.signDataAsync`, mirroring v2's own e2e tests (`dust.undeployed.test.ts`, `tokenTransfer.undeployed.test.ts`, …).
- **`initWalletWithSeed` signature merge**: v2 added `unshieldedScheme` (3rd param) and the branch added `customWallets` (dust wallet factory + `manualSync`). Kept v2's signature intact and appended `customWallets` as the 4th param; projections-test call sites updated.
- **Infra**: kept all of v2's infra (node `2.0.0-rc.3`, proof-server `9.0.0-rc.3`, toolkit `2.0.0-rc.3`) except the indexer image, bumped `ca3e554` → `…-blockhash-dust-f06af4b4`, because the snapshot-based `dustGenerations(dustAddress, blockHash, dtimeCutoffHeight)` subscription the sync depends on landed in indexer commit f06af4b4 (a strict descendant of v2's current ca3e554 — 11 commits ahead, 0 behind, same feature line).
- **Changesets**: dropped `.changeset/ledger-8-2-rc-1.md` (an "upgrade to ledger-v8 8.2.0-rc.1" changeset is wrong on v2, already on v9). All other fast-sync changesets kept.
[...]

Issue #288, opened 2026-04-07
[Sync] Event-less subscription for Dust wallet


kapke:
## What is needed

1. Sync process of Dust wallet, which does not require fetching and processing all events, while still enabling privacy if used properly
2. A configuration option indicating how much privacy should be retained
3. Flexible implementation, so that builders of Lace of 1AM are not bound to the choices made in the SDK when it comes to:
   - when sync is triggered
   - how secret keys are stored and provided to the process

## Process outline

It is possible, but more involved, to sync the Dust wallet without pulling all the events, by following a process outlined at https://github.com/midnightntwrk/midnight-indexer/issues/974 and https://github.com/midnightntwrk/midnight-architecture/blob/main/components/WalletEngine/Specification.md#indexing-service-for-dust:
1. Get latest block to have references to which rest of queries should align
2. Sync up Dust generations using a similar mechanism to the one used by shielded wallet
3. Repeatedly, until Indexer does not return empty results:
   1. check all of known coins and compute their nullifiers
   2. make request(s) to indexer for transactions featuring those nullifiers (one can investigate events to learn them locally)
4. Finally, repeatadely make queries for collapsed Merkle tree updates to rebuild Dust commitments Merkle tree:
   1. analyze gap between all coins received in previous step, the end of the block fetched in step 1) and the `first_free` of the local commitments Merkle tree copy
   2. fetch and apply collapsed Merkle tree update to fill the gap
   3. apply next coin's commitment to the commitment tree
5. verify Merkle tree roots against the ones lerned at step 1)

From efficiency standpoint relationship to transaction history is important - wallet only needs to discover transactions from past generations for transaction history purposes. 

## Privacy configuration

It affects how long nullififer prefixes will be in relation to number of spends known in the network (end index of the commitments tree is a proxy for that). In reference to https://github.com/midnightntwrk/midnight-ledger/blob/main/spec/dust.md#wallet-recovery it's probably best if the config value is exponent $n$ of 2, so that $2^n$ becomes the desired size of anonymity set. Nice side-effect is that such value creates a nice gauge of overhead:
- _0_ means no privacy, no overhead (anonymity set equals $2^0 = 1$)
- _7_ means 128x overhead, as well as 128 size of the anonymity set (127 additional values expected per query)

### Flexibility 

[...]

PR #624, opened 2026-07-31
test(dust-wallet): cover cNight generation discovery, the nullifier resolution loop, and pin a progress defect

jtsang586:
## What

Adds unit coverage for the projections-based (event-less) Dust sync. Test-only — **no production code behaviour is changed**. The single non-test edit is one added `export`.

Closes the two remaining test cases named in the Testing section of #288, and pins a defect found on the way, filed as #623.

## Why

The projections sync shipped in #617 with three of its named test cases uncovered and the nullifier-resolution loop never exercised past a single round. On PRs this feature currently gets unit coverage only, so gaps here are not caught anywhere else.

## Coverage added

**cNight generation discovery** (#288 case 2 — "this is the case on Mainnet"). This lands entirely on `DustGenerationsSyncUpdate.create`, the projections-model counterpart of the spec's `apply_system_transaction`, which had no coverage at all. Six tests pin registration fan-out ordered by `generationMtIndex`, a distinct nullifier per registration, the foreign-owner filter, the strict `>` restore boundary, dtime extraction with `__typename` stripped, and progress items being ignored while `rawUpdates` is preserved.

The open question of whether this needed Docker is answered: it does not.

**Nullifier resolution loop, 50 rounds deep** (#288 case 3). Three tests drive `resolveNullifierSpends` over a 50-round stubbed chain, asserting it terminates, re-queries only still-live nullifiers, leaves exactly one live UTXO at the chain tip, and reports progress once per productive round. Previously the deepest chain any test reached was 2.

**Progress monotonicity — known defect, see #623.** Drives the real `doEventlessSync` and captures every reported `appliedIndex`. On a restored wallet the sync reports `[13, 3, 10, 17]` — progress jumps backwards. Marked `it.fails` so runs stay green while the defect is outstanding; fixing it makes the test pass, which makes `it.fails` report a failure and prompts removing the marker.

## Not vacuous

The cNight tests characterise pre-existing code, so passing on the first run proves nothing on its own. Six mutations were seeded into `DustGenerationsSyncUpdate.create` — drop the owner filter, `>` to `>=`, drop each of the two `toSorted`s, change `seq`, use a constant first-nonce — and **all six were caught**. The schema was restored and re-verified clean afterwards.

The `it.fails` marker was likewise verified: applying a candidate fix makes the run go red, confirming the alert fires.

## Also

Tags the projections/event-sync parity e2e `@smoke`, so the feature has a PR-lane regression gate instead of nightly-only coverage. Selection confirmed with `vitest list --project undeployed -t @smoke`.

## Test plan

- `yarn test:unit --filter=@midnightntwrk/wallet-sdk-dust-wallet` → 10 files, **79 passed | 1 expected fail**
- typecheck, lint, prettier, and `@effect/language-service` diagnostics all clean
- `changeset:check` exits 0 — test-only, and dust-wallet is already bumped by pending changesets

## Reviewer notes

- `resolveNullifierSpends` is now exported so the loop can be driven directly. This does widen the package's public surface via `src/index.ts`, but matches existing precedent in the same file (`nullifierPhaseProgress`, `createDustUtxoUpdates`, `doEventlessSync` are already exported the same way).
- One justified type cast: a real `DustGenerationTreeInsertionPath` cannot be constructed in a unit test (the ledger throws `invalid index into sparse merkle tree` on an empty `DustGenerationState`). `create` treats the field as an opaque passthrough, so the test uses a sentinel and asserts identity.
- The dead `@healthcheck` tag on this file's fourth test is left alone — out of scope here.

### DIFF
--- a/packages/dust-wallet/src/v1/Sync.ts
+++ b/packages/dust-wallet/src/v1/Sync.ts
@@ -265,7 +265,9 @@ export const nullifierPhaseProgress = (
     totalNullifiers * maxCommitmentEndIndex,
   );
 
-const resolveNullifierSpends = (
+// Exported for test: this is the ticket's step 3 ("repeatedly query until the indexer returns empty"), and the
+// only way to exercise the loop past one round is to drive it directly with a stubbed nullifier subscription.
+export const resolveNullifierSpends = (
   initialNullifiers: DustNullifier[],
   initialNewUtxos: DustUtxoMap,
   pendingDust: Map<DustNullifier, Dust>,
--- a/packages/dust-wallet/src/v1/test/sync.test.ts
+++ b/packages/dust-wallet/src/v1/test/sync.test.ts
@@ -12,6 +12,7 @@
 // limitations under the License.
 import {
   type DustGenerationTreeInsertionPath,
+  DustLocalState,
   DustSecretKey,
   dustFirstNonce,
   dustNullifier,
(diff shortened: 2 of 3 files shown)


jtsang586:
Closing and reverting this work — superseded by a change of direction. No production code was affected.

PR #628, opened 2026-08-03
feat(runtime): variant activation hook and registration-derived activation range

agronmurtezi:
## What

First work package of the hard-fork foundation milestone: the `packages/runtime` machinery a variant handoff needs at a
protocol fork. No behavior changes for a single-variant wallet — this is additive runtime API plus one new field on the
variant start context.

Mainnet will hard-fork from ledger-v8 to ledger-v9, and one SDK release has to run v8 behavior pre-fork and v9 post-fork,
switching automatically mid-sync. The variant architecture was built for that, but two runtime gaps meant the handoff
could never complete:

- **Gap 1 (restart)**: after a migration, `initHeadVariant` only calls `variant.start(...)`. Nothing re-invokes
  `startSyncInBackground`, so the newly activated variant sits idle.
- **Gap 2 (keys)**: the secret keys sync needs arrive as the `startAux` of `startSyncInBackground` — they are not in
  state and not serialized, so they are unavailable at `migrateState`/`start(context)`.

## Changes

**`Runtime.onVariantActivation<E>(impl)`** — registers a poly-function watcher notified with the running variant that
became current after a migration, and never for the variant the runtime started with. This is the seam the per-wallet
work uses to re-dispatch `startSyncInBackground` with keys the wallet class retains in memory, so keys never have to
enter serialized state (Gaps 1+2).

Implementation notes:
- Unbounded activation `PubSub`, acquired with `Effect.acquireRelease` under the runtime scope captured via
  `Effect.scope`. Unbounded because a bounded pubsub would let a slow watcher block the publish inside the migration's
  `followEffect` and stall the migration itself; dropping when nobody is subscribed is the desired semantics.
- Published after `SynchronizedRef.set(currentVariantRef, …)` and before control passes to the new variant's stream.
- Registration does `Scope.extend(PubSub.subscribe(…), runtimeScope)` and only then `Effect.forkIn(consumer, …)`, so an
  activation triggered immediately after the effect resolves cannot be missed.
- Each notification is `Poly.dispatch(…).pipe(Effect.ignoreLogged)`: a failing watcher neither fails the runtime nor
  unsubscribes itself.

**`VariantContext.activationRange: ProtocolVersion.Range`** — the half-open `[sinceVersion, nextSinceVersion)` window the
runtime already derived from `withVariant` registration (`validVersionRange`) is now handed to the variant at `start`.
The upcoming per-wallet work needs it inside the pure sync capability to split a batch that straddles the fork (apply the
prefix, stop at the boundary, annotate the out-of-range version so the runtime migrates) and for the restore-healing
emission. The value existed but wasn't reachable from inside a starting variant: the `RunningVariant` record that holds
it is built *from* the result of `start`, and a variant has no back-reference to the runtime. Threading it through the
context makes `withVariant(sinceVersion)` the single source of truth, so the migration boundary and the sync split point
cannot drift apart — and keeps the not-yet-final fork version out of the code entirely.

**`Variant.migrateState` error channel widened** to `WalletRuntimeError`. Such a failure already surfaced on the state
stream at runtime; now the contract says so. Existing implementations remain valid, since `Effect<TState>` is assignable
to `Effect<TState, WalletRuntimeError>`.

## Compatibility

`runtime` **minor** (changeset included). `onVariantActivation` is additive. `VariantContext.activationRange` is breaking
only for code that constructs a `VariantContext` and calls `Variant.start` (or a `RunningVariant` constructor) directly,
which is not part of the normal wallet lifecycle. All three such call sites in this repo are test stubs and are updated
here: `packages/dust-wallet/test/DustWallet.test.ts`, plus `runningV1Variant.test.ts` in `shielded-wallet` and in
`dust-wallet` — the latter two drive sync fan-out on a single variant and never cross a version boundary, so they get the
widest range (`MinSupportedVersion` to `MaxSupportedVersion`).

## Testing

TDD per CLAUDE.md — red tests are the first commit (`dbae745b`), reviewed before the implementation commit
(`9604ffe4`). Six new tests in `packages/runtime/src/test/runtime.test.ts`, driven by the synthetic
`InterceptingVariant` harness:

1. `onVariantActivation` fires exactly once after a migration, with the newly activated variant
2. it does not fire when no migration occurs
3. it observes every migration of a 3-variant chain in order, surviving a handler that fails
4. the subscription is live as soon as `onVariantActivation` resolves (register → immediately trigger → observed)
5. each variant receives its registration-derived `activationRange`: `[0n,50n)`, `[50n,100n)`, `[100n, Max)`
6. a `migrateState` failure surfaces as `WalletRuntimeError` on the state stream

Each of the six is a single effect — one `Effect.gen` ending in `.pipe(Effect.scoped, Effect.runPromise)` — with wallet
startup as an `Effect.acquireRelease`, so `stop()` also runs when a test fails mid-way.

Verified on the branch tip: repo-wide `typecheck` (35 tasks), `lint` (55) and `format:check` (21) clean, and `test:unit`
green across all 50 tasks (runtime 41/41, dust-wallet and shielded-wallet included). Effect diagnostics on the touched
files report only two pre-existing hits on the untouched `'No variant to init'` lines — nothing on the PubSub/watcher
forks or the reworked tests.

## Review follow-ups (in `d9eb2d7c` and `d8cdad9c`)

- The activation/context tests were rewritten from sequences of awaited `Effect.runPromise` calls into one effect per
  test, per review.
- The two design points raised in review are now documented where they apply rather than only here: `migrateState`
  carries the invariant that an implementation is expected to be written so that it succeeds, and states that the error
  channel exists for the part of the migration that is not the variant's own code — the v8-to-v9 translation is a
  ledger-side tool across a WASM boundary that signals failure by throwing, so typing it as infallible would only turn a
  diagnosable stream failure into a defect. `activationRange` documents why it is handed through the context even though
  the runtime keeps the same range on its `RunningVariant` record.
- Two variant-context stubs missed by the original commit (`runningV1Variant.test.ts` in shielded-wallet and
  dust-wallet) broke `typecheck`, which skipped format/lint/barrel-export and left every test stage skipped. Fixed; CI's
  `General Checks` is green again.

## Follow-ups (same milestone, not in this PR)

Per-wallet work for shielded / unshielded / dust: write the indexer-reported `protocolVersion` into state (the version
signal itself — today `ProtocolState.version` pins at `0n`, so fork detection cannot fire), boundary splitting in
`applyUpdate`, the `migrateState` strategy seam, the restore-healing emission, and `start()` retaining its aux plus
registering the activation watcher. The handoff tests that prove the full transition land with those.

### DIFF
--- a/packages/runtime/src/abstractions/Variant.ts
+++ b/packages/runtime/src/abstractions/Variant.ts
@@ -21,6 +21,16 @@ import type * as StateChange from './StateChange.js';
 
 export interface VariantContext<TState> {
   stateRef: SubscriptionRef.SubscriptionRef<TState>;
+  /**
+   * The half-open protocol version range `[sinceVersion, nextVariantSinceVersion)` the variant is active for, derived
+   * from the registration order of the variants.
+   *
+   * @remarks
+   *   A variant uses it to recognize data that belongs to another variant: observing a protocol version outside of this
+   *   range is what makes it emit a {@link StateChange.VersionChange}, so the runtime can migrate to the variant that
+   *   owns that version.
+   */
+  activationRange: ProtocolVersion.ProtocolVersion.Range;
 }
 
 /**
@@ -38,7 +48,7 @@ export type Variant<
 > = Poly.WithTag<TTag> & {
   start(context: VariantContext<TState>): Effect<TRunning, WalletRuntimeError, Scope.Scope>;
 
-  migrateState(previousState: TPreviousState): Effect<TState>;
+  migrateState(previousState: TPreviousState): Effect<TState, WalletRuntimeError>;
 };
 
 export type RunningVariant<TTag extends symbol | string, TState> = Poly.WithTag<TTag> & {
--- a/packages/runtime/src/abstractions/Variant.ts
+++ b/packages/runtime/src/abstractions/Variant.ts
@@ -29,6 +29,12 @@ export interface VariantContext<TState> {
    *   A variant uses it to recognize data that belongs to another variant: observing a protocol version outside of this
    *   range is what makes it emit a {@link StateChange.VersionChange}, so the runtime can migrate to the variant that
    *   owns that version.
+   *
+   *   The runtime keeps the same range on its own `RunningVariant` record, but that record is built _from_ the result of
+   *   {@link Variant.start} and a variant holds no back-reference to the runtime, so it is unreachable from inside a
+   *   starting variant. Handing it over through the context keeps `withVariant(sinceVersion)` the single source of
+   *   truth, so the migration boundary the runtime enforces and the boundary a variant splits its sync batches on
+   *   cannot drift apart.
    */
   activationRange: ProtocolVersion.ProtocolVersion.Range;
 }
@@ -48,6 +54,20 @@ export type Variant<
 > = Poly.WithTag<TTag> & {
   start(context: VariantContext<TState>): Effect<TRunning, WalletRuntimeError, Scope.Scope>;
 
+  /**
+   * Produces this variant's initial state from the previous variant's state, at the protocol version boundary.
+   *
+   * @remarks
+   *   An implementation is expected to be written so that it succeeds: a migration that fails leaves the wallet unable to
+   *   follow the chain past the fork, so re-shaping state must not be able to reject state a previous variant
+   *   considered valid.
+   *
+   *   The error channel is not an invitation to weaken that; it exists because part of the migration is not the variant's
+   *   own code. The v8-to-v9 state translation is a ledger-side tool reached across a WASM boundary
+   *   (`LedgerStateTranslator` in `capabilities`): it is loaded at that moment, run to completion, and signals failure
+   *   by throwing. A typed failure surfaces that on the state stream — which is what already happened at runtime before
+   *   this signature said so — instead of turning it into a defect that tears the runtime down with no diagnosis.
+   */
   migrateState(previousState: TPreviousState): Effect<TState, WalletRuntimeError>;
 };
 
--- a/packages/runtime/src/test/runtime.test.ts
+++ b/packages/runtime/src/test/runtime.test.ts
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 import { type ProtocolState, ProtocolVersion } from '@midnightntwrk/wallet-sdk-abstractions';
-import { Deferred, Effect, Ref } from 'effect';
+import { Deferred, Effect, Either, Option, Ref } from 'effect';
 import * as rx from 'rxjs';
 import { describe, expect, it } from 'vitest';
 import { StateChange, VersionChangeType, WalletRuntimeError } from '../abstractions/index.js';
@@ -363,6 +363,18 @@ describe('Variant activation and context', () => {
   const First = 'first' as const;
   const Second = 'second' as const;
 
+  /** Awaits the first state emission matching `predicate` (any emission by default) as an effect. */
+  const firstStateWhere = <TState>(
+    states: rx.Observable<ProtocolState.ProtocolState<TState>>,
+    predicate: (state: ProtocolState.ProtocolState<TState>) => boolean = () => true,
+  ): Effect.Effect<ProtocolState.ProtocolState<TState>> =>
+    Effect.promise(() => rx.firstValueFrom(states.pipe(rx.filter(predicate))));
+
+  const atVersion =
+    (version: ProtocolVersion.ProtocolVersion) =>
+    ({ version: emitted }: ProtocolState.ProtocolState<unknown>): boolean =>
+      emitted === version;
+
   const makeTwoVariantWallet = (secondBuilder?: InterceptingVariantBuilder<typeof Second, number>) => {
     const firstBuilder = new InterceptingVariantBuilder<typeof First, number>(First);
     const Wallet = WalletBuilder.init()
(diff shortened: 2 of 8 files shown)


packages/runtime/src/abstractions/Variant.ts
```diff
@@ -38,7 +48,7 @@ export type Variant<
 > = Poly.WithTag<TTag> & {
   start(context: VariantContext<TState>): Effect<TRunning, WalletRuntimeError, Scope.Scope>;
 
-  migrateState(previousState: TPreviousState): Effect<TState>;
+  migrateState(previousState: TPreviousState): Effect<TState, WalletRuntimeError>;
```
kapke, line 51: I am hesitant about this change. No error type allowed means the migration needs to be implemented in a way, which always succeeds. I find it to be an important invariant to maintain.

packages/runtime/src/test/runtime.test.ts
```diff
@@ -336,3 +358,264 @@ describe('Wallet runtime', () => {
...
+      .withVariant(
+        ProtocolVersion.ProtocolVersion(100n),
+        secondBuilder ?? new InterceptingVariantBuilder<typeof Second, number>(Second),
+      )
+      .build();
+    return { Wallet, firstBuilder };
+  };
+
+  it('invokes onVariantActivation exactly once, with the newly activated variant, after a migration', async () => {
+    const { Wallet } = makeTwoVariantWallet();
```
kapke, line 379: It's better to implement a single effect running the test, instead multiple awaited "Effect.runPromise" calls. It's more idiomatic and in line with the way effect systems are meant to work. It should also be easier to read.

packages/runtime/src/abstractions/Variant.ts
```diff
@@ -21,6 +21,16 @@ import type * as StateChange from './StateChange.js';
...
+  /**
+   * The half-open protocol version range `[sinceVersion, nextVariantSinceVersion)` the variant is active for, derived
+   * from the registration order of the variants.
+   *
+   * @remarks
+   *   A variant uses it to recognize data that belongs to another variant: observing a protocol version outside of this
+   *   range is what makes it emit a {@link StateChange.VersionChange}, so the runtime can migrate to the variant that
+   *   owns that version.
+   */
+  activationRange: ProtocolVersion.ProtocolVersion.Range;
```
kapke, line 33: Is this necessary here? The `RunningVariant` in Runtime already has these.

packages/runtime/src/abstractions/Variant.ts
agronmurtezi: Agreed on the invariant, and I've written it into the contract rather than left it implied — `migrateState` now documents that an implementation is expected to be written so that it succeeds, and that re-shaping state must not be able to reject state a previous variant considered valid.

The error channel isn't there for the variant's own code, though; it's there for the part of the migration that isn't the variant's code at all. Since this PR was opened, the real translation landed downstream: `LedgerStateTranslator` (`packages/capabilities/src/simulation/LedgerTranslation.ts`, backed by the WASM crate in `packages/state-translation`) is a ledger-side tool reached across a WASM boundary — loaded at that moment, run incrementally to completion under a per-step cost budget, and signalling failure by throwing. That can't be typed as infallible.

With `Effect<TState>` the failure doesn't disappear, it just changes shape: it becomes a defect that tears the runtime down with no diagnosis, instead of a typed failure on the state stream. And surfacing on the state stream is what already happened at runtime before the signature admitted it — the new test `surfaces a migrateState failure as a WalletRuntimeError on the state stream` pins that behaviour. Existing infallible implementations stay valid, since `Effect<TState>` is assignable to `Effect<TState, WalletRuntimeError>`.

Rationale is now a `@remarks` block on `migrateState` in d9eb2d7c.

packages/runtime/src/abstractions/Variant.ts
agronmurtezi: It does, as `RunningVariant.validVersionRange` — but that record is built *from* the result of `start` (`Runtime.ts`: the context goes into `variant.start(...)`, and the `RunningVariant` is constructed from what it returns), and a variant holds no back-reference to the runtime. So the value is unreachable from precisely the place that needs it: inside a starting variant, where the pure sync capability has to split a batch that straddles the fork — apply the prefix, stop at the boundary, annotate the out-of-range version so the runtime migrates — and where the restore-healing emission needs the same bound.

The alternative is each variant knowing its own boundary, which duplicates what `withVariant(sinceVersion)` already declares and lets the runtime's migration boundary drift from the variant's split point. Threading it through the context keeps registration the single source of truth and keeps the not-yet-final fork version out of variant code entirely.

That reasoning is now in the `activationRange` JSDoc rather than only in the PR description (d9eb2d7c).

packages/runtime/src/test/runtime.test.ts
agronmurtezi: Done in d9eb2d7c — every test in this block is now a single effect: one `Effect.gen` ending in `.pipe(Effect.scoped, Effect.runPromise)`, with no intermediate `Effect.runPromise`.

A few things fell out of it:
- Wallet startup is an `Effect.acquireRelease` under `Effect.scoped`, so `stop()` also runs when a test fails mid-way. Before, a failed assertion leaked a running wallet.
- The ad-hoc rx awaits are two helpers, `firstStateWhere(states, predicate)` and `atVersion(version)`.
- The `migrateState` failure test asserts over `Either` (`Either.isLeft` / `Either.getLeft`) instead of the hand-rolled `{ kind, ... }` wrapper. It still subscribes before the migration is triggered so the terminal failure can't be missed — that one subscription is a plain statement rather than an `Effect.sync`, since wrapping a promise-returning thunk in `Effect.sync` trips `lazyPromiseInEffectSync`, and forking it would race the failure.

41/41 runtime unit tests green under `--force`; typecheck, lint and Effect diagnostics on the file clean (0 errors, 0 warnings).

agronmurtezi:
Consolidated for review: folded into #689, which carries the hard-fork foundations slice (#628+#629+#631+#644) as one net diff against `v2`. Branch and commits untouched — this description remains the per-increment story.

PR #629, opened 2026-08-03
test(capabilities): prove ledger-v8 / ledger-v9 WASM coexistence (fork-simulation spike)

agronmurtezi:
> **Stacked on #628** (`hard-fork-support`). Base is that branch, so this PR's diff is only the spike commit. Merge #628
> first, or retarget this to `v2` after it lands.

## What

De-risking spike for byte-faithful hard-fork simulation: proves that the ledger-v8 and ledger-v9 WASM modules can be
loaded and used side by side in a single process.

This gates the whole approach. The planned fork harness runs a v8-typed simulator to produce pre-fork blocks and hands
over to the v9 simulator at the fork block, so both WASM modules are live in one vitest process. If that didn't work, the
approach would need rethinking before anything gets built on it. It works.

## Tests

Five tests in `packages/capabilities/src/simulation/test/dualLedger.test.ts` (unit tier, no Docker):

1. blank `LedgerState` constructed from each module in the same process, asserting genuinely distinct class identity —
   a v8 state is not an instance of `v9.LedgerState` and vice versa
2. v8 `LedgerState` serialize → deserialize → serialize round-trip with the other ledger loaded
3. the same for v9
4. interleaved `ZswapSecretKeys.fromSeed` / `createShieldedCoinInfo` / `coinCommitment` alternating between modules on
   every step, so any shared global WASM state would surface; each module still produces its original commitment after
   the other was exercised
5. both `LedgerState`s still serialize to their original bytes after heavy interleaved use of the other module

All 5 pass.

**Deliberately not asserted:** any cross-version compatibility — v9 deserializing v8 bytes, or commitments/token types
matching across versions. Those are separate open questions for the ledger team, and the spike would be misleading if it
implied an answer.

## Dependency scoping

`@midnight-ntwrk/ledger-v8@^8.1.0` is added as a **devDependency** of `capabilities`, not a dependency. `capabilities` is
published, and the only importer of v8 today is this test — consumers should not be made to install a second WASM blob
for code that doesn't exist yet. It gets promoted to a real dependency when the vendored v8 simulator lands in `src/`, at
which point the published-bundle-size question becomes real and worth a browser check.

Note the two ledger packages live under different npm scopes (`@midnight-ntwrk` for v8, `@midnightntwrk` for v9), which
is what makes coexistence possible at the package-manager level in the first place.

## Verification

`yarn test:unit --force --filter=@midnightntwrk/wallet-sdk-capabilities` (5/5), plus `typecheck`, `lint`, `format:check`
and `changeset:check` clean. No changeset needed: test + devDependency only, no published behavior change.

### DIFF
--- a/packages/capabilities/package.json
+++ b/packages/capabilities/package.json
@@ -67,7 +67,8 @@
     "@effect/platform-node": "0.107.0",
     "@effect/rpc": "0.75.1",
     "@effect/sql": "0.51.1",
-    "@effect/workflow": "0.18.2"
+    "@effect/workflow": "0.18.2",
+    "@midnight-ntwrk/ledger-v8": "^8.1.0"
   },
   "scripts": {
     "typecheck": "tsc -b ./tsconfig.json --noEmit --force",
--- a/packages/capabilities/src/simulation/test/dualLedger.test.ts
+++ b/packages/capabilities/src/simulation/test/dualLedger.test.ts
@@ -1,0 +1,100 @@
+// This file is part of MIDNIGHT-WALLET-SDK.
+// Copyright (C) Midnight Foundation
+// SPDX-License-Identifier: Apache-2.0
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


agronmurtezi:
Consolidated for review: this increment is folded into #644, which will carry the whole hard-fork foundations slice (#628+#629+#631+#644) as one net diff. Branch and commits are untouched; this description stays readable here.

agronmurtezi:
Correction: the foundations slice lives in #689, not #644 — GitHub's stack feature base-locked #644, so it was replaced rather than retargeted.

PR #631, opened 2026-08-03
feat(capabilities): dual simulator + fork harness (Phase 2)

agronmurtezi:
Phase 2 of the [hard-fork support plan](https://github.com/midnightntwrk/midnight-architecture): a byte-faithful fork-simulation harness in `capabilities`, so wallet behaviour across the ledger-v8 → ledger-v9 hard fork can be tested against real ledger bytes on both sides of the boundary.

**Stacked**: base is `test/dual-ledger-coexistence-spike` (#629), which stacks on `hard-fork-support` (#628). The diff here is Phase 2 only. Note both parent branches were force-updated since the plan was written, so the SHAs it records (`780da9bb`, `a28e2853`) are stale — this branch is cut from the current spike head `a4b2d39a`.

## What lands

**The pre-fork twin.** `main`'s simulator, vendored verbatim as `src/simulation/v8/` — it is a pure import-swap twin of the current ledger-v9 one, so it was copied unmodified (`git show origin/main:...`), including `main`'s own simulator test file so the twin has real end-to-end coverage. It is exported as the `V8` namespace of the `simulation` entry point, because both twins export identical names over structurally distinct ledger types. The two `Simulator.ts`/`SimulatorState.ts` pairs remain byte-identical apart from their ledger import line, verified by `diff`; every change below was made once and mirrored mechanically.

**`@midnight-ntwrk/ledger-v8` promoted devDependency → dependency**, now that vendored source in `src/` requires it.

**Version timeline on both twins.** `Block` and `SimulatorState` carry `protocolVersion`, stamped at production time from `protocolVersionAt` so a block and the chain's current version cannot disagree, and produced blocks are never revised. `Simulator.setProtocolVersion` switches version going forward; `Simulator.scheduleFork(atBlock, version)` activates one at a height. `SimulatorConfig` gains `protocolVersion`, `genesisBlockNumber` and `genesisTime`, the last two so a chain can continue an existing one's height and clock instead of restarting at zero and the epoch. `Simulator.produceEmptyBlock()` reaches a height without traffic, since a fork happens at a height whether or not transactions were submitted. `Simulator.fromState` becomes public — the harness seeds a chain from a state it constructed rather than from a config.

No protocol version is built in anywhere: `forkVersion` is a required field with no default, and the tests run the whole timeline over arbitrary pairs (`1n→2n`, `7n→9999n`, `0n→1234567n`). The real fork version is not final, so this is enforced by test rather than by convention.

**`ForkSimulator`.** Runs the pre-fork chain with the fork pre-scheduled, so the boundary block carries the fork version however the chain is driven there. That block is the signal a wallet's pre-fork variant migrates on. On reaching it the handover runs and the post-fork chain is constructed continuing from the same height.

**`ForkHandover`** is the seam for how value crosses, a tagged union with both mechanisms the plan records: `ReMint` recreates carried value in the post-fork genesis block (what a resync-style wallet migration expects, since commitments are deterministic in coin and key, so re-minted coins reproduce identical commitments), and `MigrateLedger` supplies a converted post-fork ledger directly — the slot a ledger-provided v8→v9 migration drops into. Both are implemented and tested; `MigrateLedger` is exercised today with a stand-in converter so the seam is demonstrably real rather than aspirational.

## Module structure

The `simulation` entry point holds one simulator per ledger version, plus their shared internals:

```
simulation/
  core/          ← written once: never names a ledger type
  v8/            ← pre-fork twin
  v9/            ← post-fork twin (also exported unqualified)
  ForkSimulator  ← the two composed into a chain that crosses the fork
```

Naming both versions replaced an asymmetry worth calling out: the v9 simulator used to be the *unnamed* one at the top of `simulation/` while only its v8 twin was labelled, so "which is which" was a question you answered by reading imports. Both are now reachable as `V8`/`V9`, with the v9 names still exported unqualified — so existing code and the published `/simulation` subpath are unaffected.

`core/` holds the parts that never name a ledger type, so they are written once instead of twice: the protocol-version timeline (stated over the smallest structure it needs, so it applies to any simulator state), block hashing and `BlockContext` (a plain record **defined identically by both ledgers**, verified against both `.d.ts` files, so one definition is accepted by either), and the strictness configuration and presets (plain data — turning one into the ledger's `WellFormedStrictness` stays per-version, since that constructs a ledger object). Each version's `SimulatorState` re-exports the shared names, so both barrels still present one complete surface and no caller learns core exists.

What stays duplicated is what genuinely differs: constructing and applying ledger objects — genesis minting, night rewards, transaction application, ledger state. Sharing those would need the simulator generic over ~7 ledger types; that cost in signature noise isn't worth paying yet. The twins therefore remain identical apart from their import line, and the v8 side is still *regenerated* from the v9 side by swapping that line and `diff`ing, never hand-edited.

## Design points worth a reviewer's attention

**The boundary height exists on both chains.** The pre-fork chain's block `forkBlock` is a ledger-v8 block bearing the post-fork version tag — observed but not applied by a wallet — and the post-fork chain's genesis is the same number, re-delivering that height with v9 content. Two blocks sharing a height is deliberate: it is what makes the planned suffix-re-fetch semantics (D2) and the recorded cursor-semantics risk testable, mirroring a wallet re-fetching the boundary with its new codec.

**The fork fires on chain height, not on a harness call.** A watcher fiber performs the handover when the pre-fork chain reaches `forkBlock`, so Phase 5 can fork a chain mid-sync without poking the harness. Two orderings matter and are commented in place: the watcher reads current state *before* subscribing and the pre-fork chain's shared stream replays its latest value, so the boundary cannot slip between the two; and the post-fork reference is set *before* the deferred completes, so nothing woken by `awaitPostFork` sees an absent chain.

**Not enforced:** the pre-fork chain can still produce blocks past the boundary, which a real chain could not. Enforcing it costs complexity for no Phase-2 or Phase-5 benefit, so the constraint is documented on the class instead.

## Bundle impact

`capabilities` is a published package and from here on ships two WASM blobs: ledger-v8 `midnight_ledger_wasm_bg.wasm` at **9.7 MB** alongside ledger-v9's **9.8 MB** (~10 MB installed per package). Browser consumers roughly double the ledger WASM they download unless the bundler can drop the unused one — and it cannot here, since `simulation/index.ts` statically re-exports the v8 twin. Worth deciding before release whether the simulation entry point should be a separate optional export or the v8 twin moved to a test-only package; flagged as a release-gating question rather than resolved here, since it is a packaging decision beyond this phase.

## Also in this diff (out of the stated scope, called out deliberately)

Two `RunningV1Variant` test call sites — `packages/shielded-wallet/src/v1/test/runningV1Variant.test.ts` and `packages/dust-wallet/src/v1/test/runningV1Variant.test.ts` — construct a `VariantContext` without the `activationRange` that Phase 1 made required. This had the **repo-wide typecheck red on the base branch** before any change of mine (verified by stashing), so it was blocking the verification gate. Fixed the same way Phase 1 fixed `DustWallet.test.ts`, passing a full-span range. It arguably belongs on #628 instead — happy to move it there if you would rather keep this diff pure.

## Verification

- 35 new tests, written and confirmed red on the missing API first (commit `d9145c59`), then passing **unmodified** after implementation (`f03881c2`).
- The module restructure (`5aac37e2`) is behaviour-preserving: all **128** unit tests pass unmodified, which is why it was done after the tests existed rather than before.
- Repo-wide: `yarn typecheck --force` 35/35 clean, `yarn test:unit --force` 50/50 tasks green, `yarn lint --force` 55/55 clean.
- The vendored v8 twin's 34 tests pass unchanged.
- Effect diagnostics on all 12 touched files: 0 errors, 0 warnings. Two informational `unnecessaryFailYieldableError` messages remain in each `Simulator.ts` on pre-existing lines inherited from `main` and not touched here (the same two Phase 1 recorded).
- `yarn dist --force` clean; `dist/simulation/v8/` emits as expected.
- Changeset: `capabilities` **minor**, spelling out the two-WASM-blob consequence.

## Out of scope

Nothing under `packages/shielded-wallet` beyond the typecheck fix above (Phases 3–5), no wallet-side `applyUpdate` boundary logic (Phase 4), no runtime changes.


agronmurtezi:
Consolidated for review: folded into #689, which carries the hard-fork foundations slice (#628+#629+#631+#644) as one net diff against `v2`. Branch and commits untouched — this description remains the per-increment story.

PR #637, opened 2026-08-04
test(dust-wallet): cover the event-less Dust sync, and fix the deadlock it exposed

jtsang586:
Test coverage for the event-less ("projections") Dust sync from #288 / #617 — and one real defect found and fixed along the way.

The feature shipped with unit coverage on `Utils`/`BlockDataSchema` and two `undeployed`-only e2e footprints. That turned out to be why a hard deadlock survived to a live network.

## The defect: projections sync deadlocked on any wallet with a spend history

`doEventlessSync` runs its whole pass inside `Stream.asyncEffect`'s register effect and passed no buffer argument. That default is a **suspending, bounded queue of 16** (`effect/dist/esm/internal/stream.js:170-175`), not an unbounded one. Effect does not begin draining until the register effect returns, so every update of a pass had to fit in the buffer at once. The terminal `emit.end()` takes one slot, leaving 15 usable emissions — and a pass emits one per resolved Dust spend on top of five fixed ones.

The failure was **silent**: no error, no timeout, 0.1% CPU, frozen network counters. `isSynced` never became true and `waitForSyncedState()` / `doSync()` never returned, so it read as a slow sync.

| | emissions | outcome |
| --- | --- | --- |
| small wallet (7 Dust UTXOs) | 6 | synced in 2.4 s |
| funded faucet (90 Dust UTXOs) | ~20 | **deadlocked indefinitely** |

Any wallet whose deepest Dust chain exceeds ~10 spends was affected. Every `undeployed` wallet sits far below that, which is why local runs were always green.

Fixed by creating the stream with an unbounded buffer. On stagenet the wallet that previously hung indefinitely now syncs in **10.6 s**.

> Reviewers: use `git diff -w` on `Sync.ts`. Adding the second argument makes prettier reflow the whole generator body, so the real change is 8 added lines and the rest is re-indentation.

## Coverage added

**Unit (`dust-wallet`)** — closes all three cases named in #288's Testing section, plus four gaps:

- the nullifier-resolution loop over a **50-deep** spend chain: terminates, never re-queries settled nullifiers, one query per round, final state at the chain tip, monotonic progress within the loop
- cNight generation discovery on `DustGenerationsSyncUpdate.create`: registration fan-out ordered by generation index, distinct nullifier per registration, foreign-owner filter, restored-tree boundary, dtime extraction, progress items ignored while `rawUpdates` are preserved
- indexer request shape: an empty wallet short-circuits to one block query, one generations subscription, one commitment update and **zero** nullifier queries; and the wallet's own commitment indexes are left out of the collapsed-update ranges
- the two `SyncSchema` merkle decoders: a bad payload surfaces as a schema `ParseError`, not a thrown ledger exception
- the emission-count regression test for the deadlock above

Every assertion in the first three groups was mutation-verified — seeded mutations in the owner filter, the boundary comparison, both sorts, the sequence number and the first-nonce constant are all caught.

**E2E (`undeployed`)** — restore-then-projections (gap R1). Nothing previously exercised a non-zero baseline, so `resumeFrom`, the `generationMtIndex > lastSyncedGenerationIndex` filter and the non-zero-baseline gap ranges never ran. This snapshots a projections-synced wallet, restores it through the `provideWallet` seam, drives a resumed pass over a real delta and asserts parity with a wallet that replayed the whole event history. It guards against `provideWallet`'s *silent* from-scratch fallback by asserting the pre-`doSync` UTXO count and both tree indices, which a fresh build cannot satisfy. This is also the hard-fork migration shape, since `restore()` is how a wallet crosses a protocol version.

## Known defects pinned, not fixed

- **Progress goes backwards.** `appliedIndex` drops mid-pass because the four emissions use inconsistent bases — report 1 credits the commitment tree and the nullifier term, reports 2 and 3 drop both. Pinned with `it.fails` plus a companion test that fixes the scenario, so the suite stays green and the marker deletion is the whole fix. Fixing it needs all four emissions on a common basis and would close the open `Sync.ts:613` review thread from #318.
- **`Root hashes don't match` on snapshot-restored wallets against a live network.** Unexplained. The obvious hypothesis — the non-zero-baseline resume path — is **not supported**: R1 exercises exactly that path and produces zero root-hash errors. Needs an instrumented stagenet run against a restored wallet. This is the feature's biggest remaining unknown.

## Also in here

- **An e2e flake fixed properly.** The projections parity test asserted `max(seq)` over the funded wallet's Dust UTXOs, requiring consecutive transfers to deepen one chain. That is unsatisfiable: `seq` is per chain, the funded wallet holds five equally sized chains saturated at their generation cap, coin selection takes the smallest-valued UTXO, and a spent chain regenerates a fee in under a millisecond — so by the next transfer all five are equal again and the tie breaks on merkle-index order, placing the successor last. Now sums `seq`, which counts the spends the nullifier loop must follow however selection distributes them. Three consecutive green runs, no retries.
- **A selectable dust sync model in the testkit** (additive): optional `dustWallet` factory and `manualSync` flag on `provideWallet`/`initWalletWithSeed`, forwarded by the dust and token-transfer scenarios via `walletOptions`. Defaults are unchanged, so consumers are unaffected. `projectionsDustSyncOptions` bundles the factory *with* `manualSync` because the projections sync is one-shot: its `updates` ends after one pass, so background sync converges once and never observes anything again, and the variant's background retry only re-runs on failure. A caller must still drive `doSync` at every point that would otherwise wait for background convergence.
- **Dead `@healthcheck` tags removed** from `*.undeployed.test.ts`, where they can never select — the healthcheck lane runs `--project remote`.

## Not included, deliberately

Pointing the remote lane at the projections sync. A commit doing that is in the history and was reverted: it is wrong usage until the one-shot contract is addressed, since every remote test submits a transaction and then waits for the wallet to observe it. The follow-up is a single `Stream.repeat` in `startSyncInBackground`, which would make the event-less sync a genuine background sync and reduce that change to a one-line default flip. Worth its own PR — it changes behaviour in a published package.

Consequence to be aware of: the projections sync currently has **no live-network coverage**. Only the `undeployed` lane and the `dust-fast-sync` docs snippet exercise it, and both are nightly.

## Verification

- `dust-wallet` unit suite: 88 passed, 1 expected fail (the monotonicity pin)
- `projectionsBasedSync.undeployed.test.ts`: 5/5, including the projections-versus-events root-equality parity assertion
- R1 verified against local Docker in 222 s, with the log confirming the restore path rather than the fallback
- deadlock fix verified on **stagenet**: previously an indefinite hang, now 10.6 s with `strictlyComplete` true
- typecheck, lint (`--max-warnings 0`), Effect language-service diagnostics and `changeset:check` all clean

Closes the testing portion of #288. Supersedes #318, which should be closed as superseded by #617.


packages/dust-wallet/src/v2/RunningV2Variant.ts
```diff
@@ -186,6 +186,18 @@ export class RunningV2Variant<TSerialized, TSyncUpdate, TTransaction, TStartAux>
...
       ),
+    );
+
+    // A service that synchronizes in finite passes would otherwise leave the wallet frozen at whatever its first pass
+    // saw. Repeating re-runs `startSync`, which re-reads the wallet state, so each pass resumes from what the previous
+    // one applied. The retry above cannot serve this purpose: it re-runs a pass on failure, not on completion.
+    // Repeating outside it keeps a transient failure to the pass it happened in, rather than restarting the cycle.
+    const passes = BackgroundRepeat.$match(this.#v2Context.syncService.backgroundRepeat, {
+      Once: () => pass,
+      WithDelay: ({ delay }) => Stream.repeat(pass, Schedule.spaced(delay)),
```
agronmurtezi: [P2] Close a separate resource scope after each background pass

With projections background sync enabled, every pass constructs HttpQueryClient.layer, whose Layer.effect acquires its client in the surrounding scope. Stream.repeat does not close that scope: it remains the variant-wide scope until the wallet stops. A controlled-clock reproduction using the real HTTP client layer and a mocked client factory acquired three clients over three passes, disposed of none between passes, and disposed of all three only when the variant scope closed. This makes clients and cleanup closures accumulate throughout normal background polling. Give each attempt its own scope that closes on success, failure, or interruption before repeating/retrying; keep the transaction-history fan-out on the variant scope. Apply the equivalent lifecycle change to V1 as well.

packages/dust-wallet/src/v2/RunningV2Variant.ts
```diff
@@ -186,6 +186,18 @@ export class RunningV2Variant<TSerialized, TSyncUpdate, TTransaction, TStartAux>
...
+    });
+
+    return passes.pipe(
       Stream.runScoped(Sink.drain),
       Effect.forkScoped,
       Effect.provideService(Scope.Scope, this.#scope),
```
agronmurtezi: [P2] Prevent duplicate background polling workers

Calling wallet.start() again creates another permanent worker for a WithDelay service. The existing sync lock protects individual passes, but it is released between passes and therefore cannot protect the repeating worker. With a five-second delay, two starts one second apart reproduced six passes over twelve seconds instead of three; the extra worker continues polling until shutdown. Add an atomic worker-lifetime guard that remains held during polling delays and retries and is released when the worker exits. The mirrored V1 implementation reproduces the same behavior and needs the same protection.

agronmurtezi (commented):
Reviewed the current diff against v2 (base 92f8b565). The focused unit suites report 46 passed and 1 expected failure. Separate controlled-clock reproductions confirm both lifecycle issues below in V1 and V2. Docker/live-network e2e tests were not rerun.

packages/dust-wallet/src/v2/RunningV2Variant.ts
jtsang586: Fixed on this branch. Confirmed the mechanism you describe: the source's client layers
(`HttpQueryClient.layer`, `connectionLayer`) are `Layer.effect` over an `acquireRelease`, so they
declare `Scope` as an *input* requirement rather than discharging one of their own —
`Stream.provideSomeLayer` leaves `Scope` in the stream's requirements, and `Stream.runScoped` then
bound it to the variant scope. A unit test that counts acquisitions and releases through a layer of
that exact shape recorded 3 acquired / 0 released over three passes, with the counters read before
the variant scope closed so a shutdown-time release cannot mask it.

`startSyncInBackground` is now `runScoped → Effect.scoped → Effect.retry → Effect.repeat`. Moving the
retry from `Stream.retry` to `Effect.retry` is load-bearing rather than cosmetic: a stream-internal
retry cannot close a scope between attempts, so this is what lets a failed attempt hand its clients
back before the backoff instead of holding them for the length of it.

The transaction-history fan-out is untouched, as you asked. It is pinned to the variant scope by its
own `Effect.provideService(Scope.Scope, this.#scope)`, and `getTransactionDetails`/`put` have
`R = never`, so a closing pass scope cannot strand an in-flight lookup.

Mirrored into V1 — `startSyncInBackground` is identical between the twins.

packages/dust-wallet/src/v2/RunningV2Variant.ts
jtsang586: Fixed on this branch, and your reproduction matched exactly: two starts a second apart produce six
passes over twelve seconds (`[0,1,2,3,4,5]`) where one worker produces three (`[0,1,2]`), under
TestClock at the five-second delay.

There is now a `#backgroundWorkerLock`, taken with `Ref.modify` before the worker is forked and
released by `Effect.ensuring` *inside* the forked effect, so it comes back on normal exit, on failure,
and on interruption when the variant scope closes. That covers the polling delays and retries the sync
lock cannot, since that one is released at the end of every pass.

One consequence worth flagging: the guard asks only whether a worker is running, not what it was
started with, so a second `start()` carrying *different* key material is now ignored rather than
starting a second worker for it. I've documented `start` accordingly — changing keys means stopping
first. Two workers were not better here; they would apply two keys' updates to one state ref.

Mirrored into V1. Its shipped sync services are all `Once`, where a second worker loses the lock race
and exits immediately, so there the guard is consistency between the twins rather than a live fix.

agronmurtezi (approved):
Verified both reported lifecycle issues are fixed in V1 and V2. Per-attempt resources close on completion, failure, and interruption; repeated and concurrent starts retain a single polling worker. Transaction-history work still completes after the pass scope closes. Validation: 62 focused tests and 14 independent lifecycle checks passed. Docker/live-network tests were not rerun.

PR #644, opened 2026-08-05
feat(capabilities): real v8-to-v9 state translation at the fork boundary

agronmurtezi:
Stacked on **#631** (Phase 2: dual simulator + fork harness), which stacks on #629 → #628.

**Green.** Five commits. The seam's tests were committed red and never touched afterwards; all 7 pass as authored.

The PR now covers the whole thing end to end, in three parts:

1. **The seam** — `ForkHandover.TranslateLedger`, stated in serialized bytes and as an `Effect`.
2. **The real translation behind it** — a `wasm-bindgen` crate wrapping the ledger's translation crate, plus the loader that reaches it. Real ledger-v8 bytes now become a real ledger-v9 state.
3. **CI that builds and exercises it**, so the fork tests run rather than skip.

`midnight-ledger` is **untouched**.

## Why

Phase 2 shipped the handover as a seam with two branches, and neither actually translated state:

- `ReMint` **throws the pre-fork ledger away** and re-creates value by minting into a fresh post-fork genesis block.
- `MigrateLedger` delegated to a caller-supplied `convert: (preForkState) => LedgerState` — synchronous, infallible, object-to-object. Nothing can really implement that shape, and the only thing exercising it was a stub returning `LedgerState.blank()`.

So the post-fork chain had never started from the pre-fork chain's own state. That is the thing a fork actually does.

## 1. The seam

```ts
type LedgerStateTranslator = (preForkLedger: Uint8Array) => Effect.Effect<Uint8Array, LedgerTranslationError>;

ForkHandover.TranslateLedger({ translator })
```

`ForkSimulator` does `preForkState.ledger.serialize()` → translator → `v9.LedgerState.deserialize(...)`. Three properties, each forced by how the ledger-side tool actually works:

- **Bytes**, because that is the only thing that crosses a WASM boundary. The translation links *both* ledgers at once — two majors of `onchain-state` included — which no JavaScript module can do.
- **`Effect`**, because translation is incremental under a cost budget (`start` then `run(budget)` until `result()`), not one call. Driving that loop lives inside the translator, so the seam holds either shape.
- **Fallible**, because translation can fail and the old signature could not say so.

`ReMint` stays, relabelled as an explicit approximation.

## 2. The real translation

New private package `@midnightntwrk/wallet-sdk-state-translation`, holding both halves of the wallet side:

```
src/          the loader — resolve the module, check it, hand it bytes
wasm/         the wasm-bindgen crate (standalone Cargo project)
wasm/pkg/     build output — gitignored, and the loader's default location
scripts/      build-wasm.sh, verify-wasm.mjs
```

**The bindings live here, not in `midnight-ledger`.** Needing the translation reachable from JavaScript is a wallet concern; the ledger has no use for a wasm target. So the ledger keeps only the pure Rust crate it already had, and changing the binding does not need a PR against someone else's release branch. `wasm/Cargo.toml` depends on it as an ordinary crate dependency, pinned to a git branch only until it is released — at which point it becomes `version = "0.1.0"` and nothing else changes.

Two things make consuming it from outside the ledger workspace non-obvious, both commented in place:

- **`[patch.crates-io]` has to be replicated.** The v9 line is in pre-release and not on crates.io; the ledger workspace redirects those crates to git tags, and Cargo only honours `[patch]` from the top-level manifest. The block is copied verbatim and **must be kept in step when those tags move**. The standalone `v8-to-v9-state-translation-replay` crate does the same, so there is precedent.
- **Cargo reads `.cargo/config.toml` from the working directory, not `--manifest-path`.** The build script `cd`s into the crate for exactly this reason.

### Verified end to end, against the real artifact

Five integration tests, all passing:

- A blank v8 state: **816 bytes in → 883 out**, accepted by `v9.LedgerState.deserialize`.
- A minted-and-transferred state: **every coin commitment the pre-fork chain accumulated is present post-fork**.
- **A post-fork transaction applies on top of the translated state** — commitment count is exactly `carried + 1`, which can only hold if the post-fork chain grew the carried tree rather than a fresh one. Deserializing proves the bytes parse; this proves the state is usable.
- The same bytes translate identically.
- A state the pre-fork chain never touched still translates.

Checked for vacuity: against a stub returning `LedgerState.blank()`, the two carry-tests fail. They detect a translation that does not carry state.

**Artifact size:** 5.8 MB raw wasm → **1.74 MB** after `wasm-opt -Os`. Small because this crate needs the state structures but not proving.

## 3. CI

New `Build WASM Translation` job — **the first Rust in this repository**. It installs the wasm32 target plus pinned prebuilt wasm-bindgen 0.2.104 and binaryen 131, builds, verifies, and uploads `wasm/pkg/` for the integration matrix to download. Gated only on `build`, so it runs alongside `general-checks` rather than lengthening the chain; `integration` waits on it, because without that dependency the fork file would quietly skip instead of failing. Cargo's registry and target directory are cached on `Cargo.lock`.

`verify:wasm` translates a blank v8 state and deserializes the result before the artifact is uploaded. Compiling proves little here — see the blocker below — and this names that failure where it happens rather than leaving it to surface as a bare `RuntimeError: unreachable` inside a fork test.

## ⚠️ The one external dependency: `midnight-storage` needs a wasm-safe clock

`midnight-storage`'s `state_translation.rs` calls `std::time::Instant::now()` at eight sites in its metered loop, including inside `run()`. That intrinsic is unimplemented on `wasm32-unknown-unknown` and traps — so the crate links fine and then dies on the first real state. Confirmed in isolation: a wasm export whose entire body is `Instant::now()` traps identically.

**This PR carries a vendored workaround**, and reviewers should know it is there: a committed 24-line patch switching to [`web-time`](https://crates.io/crates/web-time) (a re-export of `std::time` off-wasm, `performance.now()` on it), applied by the build script to the exact published version named in `Cargo.lock`, fetched from crates.io into a gitignored `.vendor/`. `[patch.crates-io]` points there, so a missing directory is a hard cargo error rather than a silently trapping build.

**The real fix has to be published, not just committed to the ledger repo** — this crate takes `midnight-storage` from crates.io, and redirecting it to a local checkout does not work: the git-tagged v9 crates fail against the ledger workspace's copy with 151 trait-mismatch errors in `midnight-onchain-state`, that copy having drifted past the released one. Note the resolved version is **2.0.2**, newer than the 2.0.1 the ledger workspace pins, and 2.0.2 does not contain the fix either.

`wasm/README.md` says exactly which four things to delete once a release carries it.

**Open question for the ledger team:** was running this translation in WASM ever intended? `v8-to-v9-state-translation-replay` is a native mainnet-replay tool and nothing in the crate targets wasm, so in-process WASM translation may be a new requirement rather than a bug fix.

## Resolved: the encoding question from the previous revision

This PR previously asked whether JS `LedgerState.serialize()` emits the same `tagged_serialize` encoding the crate's `tagged_deserialize` expects. **It does**, confirmed from both directions:

- JS side: `midnight:ledger-state[v13]` for `@midnight-ntwrk/ledger-v8` 8.1.x, `[v18]` for `@midnightntwrk/ledger-v9`, each followed by the same structural prefix.
- Rust side: the built artifact rejects bad input with `expected header tag 'midnight:ledger-state[v13]:'` — byte-for-byte what JS v8 emits — and accepts the real thing.

No framing step is needed anywhere.

## The bug fix (independent of the above)

The handover runs in a fiber forked at `ForkSimulator.init`, and its deferred was **only ever completed on the success path**. Anything that went wrong in that fiber killed it silently and left the deferred unresolved, so `awaitPostFork` and `advanceToFork` blocked forever — a broken handover was indistinguishable from a slow one.

Fixed with `Effect.intoDeferred`. The success path is unchanged and still sets the post-fork reference *before* completing the deferred. No type change needed — defects travel in the `Cause`, not the error channel.

This bug is in code #631 introduces. It is fixed here because its pinning test was written in the same pass as the translation tests. Say the word and I will move both down into #631 so that PR does not land the hang.

## Verification

- `turbo run typecheck lint test:unit` → **74/74 tasks**. Individually: typecheck 37/37, lint 58/58, test:unit 53/53.
- **135 unit tests** in `capabilities`; 22/22 in `forkSimulator.test.ts`, including the 7 committed red, unmodified.
- **18 unit tests** in `state-translation`, run against real `.mjs` stub modules — the actual dynamic-import path, no `vi.mock`.
- **5 integration tests** against the real WASM artifact.
- `changeset status` clean; both workflow files parse and the job graph resolves (`tests` → `integration` → `build-wasm`, so a wasm failure fails the required check).

## Notes for reviewers

- **Nothing about the published surface changes.** No wallet runtime path translates ledger state — a resync with the post-fork ledger is the wallet's migration mechanism — so the loader is a `devDependency` only, `unavailableTranslator` stays the shipped default, and consumer bundles are unaffected.
- **The fork tests are integration tier** because building the WASM is setup, not because they need infra. They skip with a loud message when the artifact is absent.
- Effect diagnostics: `LedgerTranslation.ts` reports one `deterministicKeys` error, because `LedgerTranslationError`'s tag is the short form rather than the fully-qualified path. Deliberate and pinned by tests; it matches `LedgerOps.LedgerError` and `submissionService.SubmissionError`, which carry the identical pre-existing diagnostic.


packages/capabilities/src/simulation/ForkSimulator.ts
```diff
@@ -35,28 +36,31 @@ import * as V8 from './v8/index.js';
...
+   * Translate the pre-fork chain's own ledger state into post-fork form, and start the post-fork chain from it.
    *
-   * The slot a ledger-provided v8-to-v9 state migration drops into.
+   * The faithful path, and the slot the ledger-side v8-to-v9 state translation tool drops into: see
+   * {@link LedgerStateTranslator} for why the seam is stated in bytes. Until that tool is wired up, callers supply the
+   * translation.
    */
-  MigrateLedger: {
-    readonly convert: (preForkState: V8.SimulatorState) => LedgerState;
+  TranslateLedger: {
```
kapke, line 62: IMO instead of selecting, just always accept the translator function. Is the "ReMint" option actually needed?

packages/state-translation/src/test/stateTranslation.test.ts
```diff
@@ -0,0 +1,195 @@
...
+  });
+
+  it('awaits a translation that completes asynchronously', async () => {
+    // The real translation runs an incremental, cost-metered loop; driving it to completion may have to yield.
+    const translate = createLedgerStateTranslation({ moduleSpecifier: fixture('asyncTranslation.mjs') });
+
+    expect(await translate(preForkLedger)).toEqual(reversed);
+  });
+
+  it('reports an unloadable module as unavailable, naming what to do about it', async () => {
```
kapke, line 71: Why playing with module loading at all?

packages/state-translation/wasm/README.md
```diff
@@ -0,0 +1,98 @@
...
+1. **Serialize inside the translation's lifetime.** Returning the v9 state and serializing after the
+   `TypedTranslationState` drops panics with
+   `storage-core/src/arena.rs: "Arena should contain current serialization target"` — the translated nodes live in the
+   arena the translation owns. With `panic = "abort"` the first panic also leaves an arena `RefCell` borrowed, so every
+   later call fails with a misleading `RefCell already borrowed`; chase the first panic, not the repeat.
+2. **The panic hook is load-bearing.** The `wasm` profile aborts on panic, so without `console_error_panic_hook` every
+   failure reaches JavaScript as a bare `RuntimeError: unreachable`. Both bugs above were undiagnosable until it was
+   added.
+
+## Toolchain
```
kapke, line 86: IIRC cargo supports toolchain files to automatically pick toolchain according to a spec.

kapke (changes requested):
General comment - Rust toolchain integration needs to be better:
1. turbo-based workflows do not seem to be able to pick Rust dependency tasks
2. No mention in the top-level files about Rust and toolchain requirements
3. I believe a top-level cargo.toml with definition of a workspace containing the translation would be beneficial for Rust discovery

packages/capabilities/src/simulation/ForkSimulator.ts
agronmurtezi: Removed. `ForkSimulatorConfig` takes `translator` directly and `ForkHandover` is gone — the union only existed because nothing could implement translation when it was written.

Two tests went with it: one asserted `SimulatorState` fields a bytes-only translator never sees (covered now by the input-identity and genesis-numbering assertions), and one relied on genesis-mint *events* being replayable by a wallet, which the translated path does not produce.

`6ef03806`

packages/state-translation/src/test/stateTranslation.test.ts
agronmurtezi: Gone — `src/` imports `wasm/pkg/…` statically now. The indirection existed because the artifact used to be an unpublished nix output in another repo; it is built by this package, so it was solving a problem that no longer exists. The adapter is ~20 lines.

Four of the tests covered branches that became unreachable once both sides of the boundary were ours — a second accepted export name, a `Uint8Array | Promise` return, a non-bytes guard — so those branches went with them. Three survive, in the integration tier against the real WASM.

`6ef03806`

packages/state-translation/wasm/README.md
agronmurtezi: Done — `rust-toolchain.toml` at the workspace root pins the channel and lists `wasm32-unknown-unknown`, so rustup installs both on the first `cargo` call. Dropped `rustup target add` from CI and from the README.

Pinned rather than tracking `stable`, since this produces a binary we ship and a silent compiler bump could change it or break the `blst` C build.

`6ef03806`

agronmurtezi:
All six points addressed in `6ef03806` — net **+464 / −709**.

**Rust toolchain integration**

1. **turbo cannot pick up Rust tasks** — `build:wasm` is now a real turbo task with `inputs`/`outputs`, so it caches, and `test:integration` in `capabilities` and `state-translation` declares a dependency on it. Turbo builds the artifact before the tests that need it instead of leaving the ordering to CI. Consequence worth naming: running integration tests in those two packages now needs a Rust toolchain. `dist`, `typecheck`, `lint` and `test:unit` still do not — the `.d.ts` the import types against is committed.
2. **No top-level mention** — `README.md` and `CLAUDE.md` both have Rust sections now, leading with the fact that most work needs none of it.
3. **Top-level `Cargo.toml`** — added, with the crate as a workspace member; `cargo metadata` resolves from the repo root. `[patch.crates-io]` and the `wasm` profile moved up with it, since Cargo only honours those in a workspace root, and `Cargo.lock` moved to the root too.

**Two places I went further than asked**

- The dual-simulator changeset described the handover as `ForkHandover`, which this branch removes. Restating it was not enough: the sibling changeset announced *removing* `ForkHandover`, and `ForkSimulator.ts` does not exist on `v2` — so it never reached a release. Both are now stated in terms of `translator`, and neither mentions it.
- `state-translation` declares the two ledger modules its tests import rather than relying on hoisting; the vendor stamp covers a digest of the storage patch, so editing the patch without moving the version no longer builds from a stale tree; root `target/` is excluded from the root format gate.

**Verified:** `typecheck` + `lint` + `test:unit` 74/74 · `forkSimulator.test.ts` 20/20 · state-translation integration 3/3 · fork integration 5/5 against the real WASM · `cargo metadata` from root · artifact still 1.75 MB, so the relocated `wasm` profile still applies · `turbo run build:wasm` cache hit on rerun · from a simulated fresh checkout, `turbo run test:integration` built the WASM and passed.

Ready for another look.

agronmurtezi:
Consolidated for review: replaced by #689 — same head branch, identical commits, based on `v2`, carrying the whole foundations slice (#628+#629+#631+#644) as one net diff. GitHub's PR-stack feature base-locks this PR, hence a replacement rather than a retarget. The review thread here (the translation crate and its six follow-ups) applies unchanged to the commits in #689.

PR #645, opened 2026-08-06
feat(dust-wallet): run the projections dust sync in the background

jtsang586:
Makes the projections-based ("event-less") Dust sync usable under background synchronization. Stacked on #637 — review that first; this branch's own contribution is 5 files.

## The problem

The projections sync synchronizes in **finite passes**: `doEventlessSync` ends its stream once a pass completes. The event-based service instead holds a long-lived indexer subscription that never completes.

Background synchronization was written for the latter. `RunningV1Variant.startSync` reads the wallet state **once** and invokes `updates` **once**, and `startSyncInBackground` wraps that in `Stream.retry` — which re-runs a pass on *failure*, not on *completion*.

So a wallet built with the projections sync and started in the background converged once and then never observed anything again, **silently**. The only usable shape was `manualSync: true` plus an explicit `facade.doSync()` at every point that would otherwise wait for the background sync to catch up — which is not how an application uses the SDK. Any consumer calling `wallet.start()` with a projections Dust wallet hit this.

## The change

A sync service may now declare `backgroundRepeatDelay`; background synchronization re-runs its pass on that delay.

```ts
repeatDelay === undefined ? identity : Stream.repeat(Schedule.spaced(repeatDelay)),
```

Because the repeat re-runs `startSync`, each pass re-reads the wallet state and resumes from what the previous pass applied — no change to the `SyncService` contract's shape, and no state threading.

The projections service declares it, defaulting to 5s and configurable as `backgroundSyncInterval` (ms). The event-based and simulator services omit it and are **untouched** — their `updates` never completes, so `Stream.repeat` never fires.

Three details worth checking during review, because each is easy to get wrong:

| Decision | Why |
| --- | --- |
| Repeat lives in `startSyncInBackground`, not the shared `startSync` | `sync()` — which backs `facade.doSync()` — also calls `startSync`. A repeat there would mean `doSync()` never returns. |
| Repeat sits **outside** `Stream.retry` | A transient failure is then retried within the pass it happened in, rather than restarting the whole cycle. |
| Existing `#syncLock` left alone | Each pass acquires and releases it, so passes cannot overlap. A pass that cannot take the lock is skipped and waits for the next interval. |

## Tests

Three, in `runningV1Variant.test.ts`, using `TestClock` so they are deterministic rather than timing-dependent:

- **`re-runs the pass on the declared delay, each pass starting from the state the last one produced`** — asserts the observed applied indexes are `[0, 1, 2]`. This is the load-bearing one: it fails unless the repeat re-reads state, which is the entire reason it wraps `startSync` rather than `updates`.
- **`leaves a service that declares no delay running exactly one pass`** — guards the event-based service against acquiring a repeat behind its back.
- **`keeps an explicitly driven sync to a single pass even when a repeat delay is declared`** — guards `doSync()`'s contract.

The latter two pass before the change as well as after; they are the regression net, not the demonstration. The first fails before and passes after.

`dust-wallet` unit suite: 91 passed, 1 expected fail (a pre-existing `it.fails` pinning the progress-monotonicity defect).

## Verified against a live network

On stagenet, a projections wallet in background mode reaches a synced state with **no `doSync` call** — `connected=true`, `gap=0`, `complete=true` — which is only possible if passes are repeating. Before this change that configuration converged once and stopped.

That run was done from a **clean** state cache, deliberately: e2e state snapshots on that network had been corrupted by an unrelated harness bug (`afterEach` saves wallet state even when a test fails, so one failure poisons every later run). Earlier stagenet readings taken on poisoned snapshots are not trustworthy, and are not offered as evidence here.

## Reviewer notes

- **The 5s default is a judgement call and I would like a second opinion.** It was not chosen by measurement. At 5s each wallet issues roughly 12 block queries a minute indefinitely. An idle pass is cheap — the generation subscription and the commitment load both short-circuit on an unchanged tip — but nothing here justifies 5s over 15s or 30s.
- **A persistent apply failure now recurs every interval** rather than once. That is correct behaviour given bad input, but it turns one error into a stream of them; surfacing a repeatedly-failing pass rather than retrying it silently would be a reasonable follow-up.
- Also included: a JSDoc correction in the testkit's `dust-sync.ts`, which told callers the projections sync was one-shot and had to be hand-driven. This change makes that untrue, so the text is fixed here rather than left to contradict the code.

## Not in this PR

Pointing the testkit healthcheck scenarios at the projections sync by default. That is split out to `feat/testkit-healthcheck-projections-default`, because it is a breaking change for downstream monitoring and is additionally blocked on the state-cache bug above. Keeping them separate means this change can land on its own merits.

### DIFF
--- a/packages/dust-wallet/src/v1/RunningV1Variant.ts
+++ b/packages/dust-wallet/src/v1/RunningV1Variant.ts
@@ -21,6 +21,7 @@ import {
   Duration,
   Schedule,
   Array as Arr,
+  identity,
   Ref,
 } from 'effect';
 import { type TransactionHistoryService } from './TransactionHistory.js';
@@ -147,6 +148,8 @@ export class RunningV1Variant<TSerialized, TSyncUpdate, TTransaction, TStartAux>
   }
 
   startSyncInBackground(startAux: TStartAux): Effect.Effect<void> {
+    const repeatDelay = this.#v1Context.syncService.backgroundRepeatDelay;
+
     return this.startSync(startAux).pipe(
       Stream.retry(
         pipe(
@@ -160,6 +163,13 @@ export class RunningV1Variant<TSerialized, TSyncUpdate, TTransaction, TStartAux>
           }),
         ),
       ),
+      // A sync service whose `updates` is finite — the projections service ends its stream after one pass — would
+      // otherwise leave the wallet frozen at whatever that first pass saw. Repeating re-runs `startSync`, which
+      // re-reads the wallet state, so each pass resumes from what the previous one applied. The retry above cannot
+      // serve this purpose: it re-runs a pass on failure, not on completion. Repeating outside the retry keeps a
+      // transient failure to the pass it happened in, rather than restarting the whole cycle. A service with a
+      // long-lived `updates` declares no delay, and is left exactly as it was.
+      repeatDelay === undefined ? identity : Stream.repeat(Schedule.spaced(repeatDelay)),
       Stream.runScoped(Sink.drain),
       Effect.forkScoped,
       Effect.provideService(Scope.Scope, this.#scope),
--- a/packages/dust-wallet/src/v1/RunningV1Variant.ts
+++ b/packages/dust-wallet/src/v1/RunningV1Variant.ts
@@ -21,7 +21,6 @@ import {
   Duration,
   Schedule,
   Array as Arr,
-  identity,
   Ref,
 } from 'effect';
 import { type TransactionHistoryService } from './TransactionHistory.js';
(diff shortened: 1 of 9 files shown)


packages/dust-wallet/src/v1/RunningV1Variant.ts
```diff
@@ -185,6 +188,13 @@ export class RunningV1Variant<TSerialized, TSyncUpdate, TTransaction, TStartAux>
           }),
         ),
       ),
+      // A sync service whose `updates` is finite — the projections service ends its stream after one pass — would
+      // otherwise leave the wallet frozen at whatever that first pass saw. Repeating re-runs `startSync`, which
+      // re-reads the wallet state, so each pass resumes from what the previous one applied. The retry above cannot
+      // serve this purpose: it re-runs a pass on failure, not on completion. Repeating outside the retry keeps a
+      // transient failure to the pass it happened in, rather than restarting the whole cycle. A service with a
+      // long-lived `updates` declares no delay, and is left exactly as it was.
+      repeatDelay === undefined ? identity : Stream.repeat(Schedule.spaced(repeatDelay)),
```
kapke: Do not give specific meaning to the `undefined` value. Use a union type with discriminator instead to be explicit, more like:
```
type RepeatBehavior = "once" | {"type": "with-delay", delay: number};
```

jtsang586: Done,  
`backgroundRepeat` is now a required `Data.TaggedEnum`:

    export type BackgroundRepeat = Data.TaggedEnum<{
      Once: {};
      WithDelay: { readonly delay: Duration.DurationInput };
    }>;

kapke (approved):

Issue #646, opened 2026-08-07
[Fast sync] The fast dust sync re-emits UTXOs it has already applied


jtsang586:
## Summary

The projections ("event-less") Dust sync re-emits UTXOs it has already applied. Those stale indices then corrupt the
commitment-range arithmetic in `gapRanges`, which fabricates a collapsed-update range starting **inside** the already
applied, already collapsed part of the commitment tree. The ledger rejects both the stale insert and the bogus range, so
every sync pass fails and the Dust wallet stops advancing.

Two error strings that have been seen separately turn out to be the same defect:

```
Error: values inserted non-linearly into commitment tree; expected to insert index 3145, but received 3109.
Error: attempted update on collapsed sub-tree at 3111/0
```

Both are wrapped as `Wallet.Other: Error while applying sync update` at `RunningV1Variant.ts:203`.

## Mechanism

Instrumenting `CoreWallet.applyDustCommitments` and the pass setup in `Sync.ts` gives the whole picture in two lines:

```
DIAG applyDustCommitments: applyTimeFirstFree=3145 ownUtxoIndexes=[3109,3110] ranges=[]
  -> Error: values inserted non-linearly; expected 3145, but received 3109

DIAG   before range 3111: firstFree=3145 (overlap=true)
  -> Error: attempted update on collapsed sub-tree at 3111/0
```

The pass starts from a **current** state — `passStartFirstFree=3145`, `lastSyncedCommitmentIndex=3144` — so this is not a
stale-frontier problem. What is stale is the UTXO set: the pass reports the wallet's own UTXOs at mt-indexes 3109 and
3110, which are already in the tree.

`applyDustCommitments` filters `priorUtxos` by the frontier, but the final `insertCommitments` call after the last
collapsed range does not, so the stale UTXOs are inserted behind the frontier and the ledger rejects it.

The same stale indices are passed to `gapRanges` as `skipIndexes`, and that is where the second error comes from.
`gapRanges(3144, 3144, [3109, 3110])`:

| i | boundary | `start = boundary + 1` | `end` | kept? |
| --- | --- | --- | --- | --- |
| 0 | 3144 (`lastAppliedIndex`) | 3145 | 3108 | dropped (`start > end`) |
| 1 | 3109 | 3110 | 3109 | dropped (`start > end`) |
| 2 | 3110 | **3111** | 3144 | **kept** |

[...]

PR #647, opened 2026-08-07
feat(testkit)!: monitor the projections dust sync by default, selectable by DUST_SYNC

jtsang586:
Points the testkit healthcheck scenarios at the projections ("event-less") Dust sync, and makes the choice of sync model something a whole run can switch. #645 has since squash-merged into this branch's base, and the base has absorbed the dual-ledger hard-fork stack from `v2`; this branch is merged up in `a66b9af6`, so the diff below is its own contribution.

## Why this is more than a default flip

Flipping the scenarios was tried once before and reverted, because the projections sync could not work under background synchronization. #645 fixes that. What remained was that switching models was a five-file edit with a cache-clearing step to remember in one direction — so this branch makes the model a configuration choice instead, and closes the traps that made switching unsafe.

## The projections dust wallet is now two-variant, so it can transact

Merging the base surfaced a defect that would have defeated the whole point of this PR.

`eventLessDustWallet` was a **single-variant** composition. One variant answers for the whole protocol timeline, so it reports `MinSupportedVersion`; a facade acts at the lowest version its three sub-wallets report, so this sub-wallet held the entire facade below the ledger-v9 boundary. Every transfer then failed inside `mergeUnprovenTransactions`:

```
ProtocolVersionMismatchError: This transaction was built for protocol version 2000000 ...
The caller acts at protocol versions 0 up to (but not including) 2000000.
```

Both healthcheck scenarios pay fees and transfer, so the flip could not have worked.

**The constraint was the composition, not the projections sync.** `eventLessDustWallet` is now the shipped two-variant `DustWallet` with exactly one substitution — the V2 variant's sync service. The V1 variant keeps the event stream, because projections rests on `DustLocalState` members no ledger-v8 has. On a chain running ledger-v9 from its first block the V1 variant never applies, so projections is still reached immediately, and the `must not transact` warning is gone from the file.

Measured rather than reasoned: a two-variant projections wallet transferred end-to-end with `dust.protocolVersion === 2000000`, and `smoke.undeployed` is 4/4 under both `DUST_SYNC=projections` and `DUST_SYNC=events`.

A dust snapshot written by the previous single-variant composition declares protocol version 0, is routed to the ledger-v8 variant on restore and refused by its deserializer, so the wallet rebuilds from scratch. No cache migration is needed.

## `DUST_SYNC` selects the model for a whole run

`dustWalletFromEnv()` is the default behind the `dustWallet` option of `provideWallet` and `initWalletWithSeed`. Unset means the event-stream sync; the dust and token-transfer healthcheck scenarios fall back to projections, so the monitored network is exercised against the model the wallet ships to users.

**The fallback is not a pin.** The scenarios resolve through `dustWalletFromEnv(process.env, 'projections')`, so `DUST_SYNC` overrides them too. An earlier revision used a constant `projectionsDustSyncOptions`, which is an explicit `dustWallet` and therefore beat the environment — `DUST_SYNC=events` would have switched the three helper-built remote files and silently left `dust.remote` and `tokenTransfer.remote` on projections. A lane half on each model reports coverage the run did not have, which is worse than not switching at all.

An unrecognized value throws rather than falling back, for the same reason: a typo in `DUST_SYNC` must not mislabel a run's coverage.

## Dust snapshots are namespaced per sync model

A snapshot carries one progress value, `appliedIndex`, and the two models disagree about what it means — a ledger-event cursor to the event-stream sync, a composite of tree indices and nullifier count to projections. Restoring across models therefore resumed a subscription from a position that is not an event id, and **skipped events silently rather than failing**.

Snapshots now live at `dust-<model>-<filename>`, so a model switch finds none and rebuilds from scratch. There is no longer a cache to clear by hand in either direction. Shielded and unshielded snapshots are model-independent and stay shared.

Verified on stagenet in the hazardous direction: switching a projections-synced wallet to `DUST_SYNC=events` logs `Unable to restore dust wallet` and rebuilds, where previously it would have resumed wrongly.

## Two bugs this surfaced, both fixed here

**Settling state waiters could never resolve while a wallet was syncing.** `waitForFacadePendingClear`, `waitForFinalizedShieldedBalance` and `waitForUnshieldedCoinUpdate` placed `rx.debounceTime(10_000)` *upstream* of their predicate, so the condition was only evaluated during a 10-second gap in state updates — and a projections-synced Dust wallet emits roughly every 5 seconds indefinitely. Measured on stagenet, the largest gap over a 90-second window was 6 seconds, so the debounce could never fire. Two tests burned their full bound in a single run (600000ms and 900000ms) and, with the suite's `retry: 1`, each paid twice: about 50 minutes of lane time spent on conditions that already held. The settle window now applies to the predicate, and an unmet condition fails in 180s with a legible message instead of running to the enclosing test timeout.

**Serialize/restore tests hard-coded the event-stream sync as their restore target.** `smoke.undeployed` and `emptyWallet.universal` build their own `DustWallet(...)` and restore into it state serialized from a wallet that follows `DUST_SYNC`, so under projections the restore crossed models and the wallet never reported synced — `smoke.undeployed`'s dust test burned its full 300000ms bound, twice. Both now resolve the factory the same way the rest of the run does. Snapshot namespacing cannot help here, because the state never touches disk.

## Reviewer notes

| Point | Detail |
| --- | --- |
| Breaking: `eventLessDustWallet` registers both variants | It was single-variant, which pinned the facade below the ledger-v9 boundary and made every transfer fail. Callers get the same projections sync; what changes is that the wallet can now build transactions. |
| Breaking: `saveState` takes the `WalletInit`, not the facade | It needs the sync model to pick the snapshot namespace; taking both separately would let them disagree. One-line migration, described in the changeset. |
| Breaking: `projectionsDustSyncOptions` no longer implies `manualSync` | It is now the plain background-mode drop-in the name suggests. The old pairing is `manualProjectionsDustSyncOptions`. Spreading the old constant into a test that waits on the state stream would block — a trap the old name did nothing to signal. |
| The healthcheck scenarios cost ~20s more | Two `waitForFacadePendingClear` calls, one per wallet, each with a 10s settle window. |
| Pinning a model opts a wallet out of `DUST_SYNC` | Reserve an explicit `dustWallet` for tests that genuinely require one model; a pinned wallet is left behind when a lane is switched by environment. |

## Verified

Run on the merged tree (`a66b9af6`):

- `yarn verify:check` 82/82, `yarn lint` 58/58, `yarn typecheck` 37/37.
- `yarn test:unit` 55/55 tasks (~1517 tests); `yarn test:integration` 56/56 tasks, including the Rust/WASM state-translation suite.
- Testkit unit tests 16/16. The two load-bearing assertions stay mutation-verified: making the fallback un-overridable fails 5 tests, and defaulting instead of throwing on an unknown `DUST_SYNC` fails 4.
- `smoke.undeployed` 4/4 under `DUST_SYNC=projections` **and** under `DUST_SYNC=events`.
- `projectionsBasedSync.undeployed` 5/5 under projections, so the manual `doSync` path is undisturbed by the two-variant switch.

## Not yet proven: the healthcheck lane itself

The dust and token-transfer healthchecks **cannot be validated on a local undeployed chain**. They expect a wallet already holding several unregistered Night UTXOs — a long-lived stagenet shape — and a fresh chain has none, so they fail on funding (`Not enough unregistered UTXOs found: expected 0 to be greater than 1`) before exercising anything this PR changes. What a local run does establish is that the failure has moved off the protocol boundary and into `dust-wallet/src/v2/Transacting.ts`, i.e. the projections-synced ledger-v9 variant is building transactions.

**A stagenet run is still required before this flip should be trusted in monitoring.** The stagenet results previously reported here were produced before the base carried boundary enforcement and no longer describe this branch.

## The undeployed lane is red on `v2`, independently of this PR

`yarn test-undeployed` fails 5 files on this branch. Four of them fail **identically on `v2` itself**:

| File | `v2` `92f8b565` | base `4285c23c` | this branch |
| --- | --- | --- | --- |
| `facadeTransfer.undeployed` | 5/5 failed | 5/5 failed | 5/5 failed |
| `optionalBalancing.undeployed` | 15/21 failed | 15/21 failed | 15/21 failed |
| `swap.undeployed` | 3/4 failed | 3/4 failed | 3/4 failed |
| `tokenTransfer.undeployed` | 2/18 failed | 2/18 failed | 2/18 failed |
| **Totals** | **25 failed / 21 passed / 2 skipped** | same | same |

76 `ProtocolVersionMismatchError` occurrences in one `v2` run, on a branch where `DUST_SYNC` and `dustWalletFromEnv` do not exist and the failing files hand-build the shipped two-variant `DustWallet`. The symptom is the *inverse* of the one fixed above — a transaction stamped at version 0 reaching a facade acting at 2000000 — which points at a startup race rather than a composition problem. Pre-existing; it needs its own issue.

One methodology note for anyone re-running this: **compare runs at the same file count.** A 16-file run showed 2 extra `tokenTransfer` failures that disappear in a 4-file run, on every branch.

## Known gaps

- **Nine e2e tests hand-build `DustWallet(...)` via `startWithSeed` and ignore `DUST_SYNC`** — `dustRegistration`, `dustDeregistration`, `facadeTransfer`, `swap`, `optionalBalancing`, `multipleWallets`. They are internally consistent, so nothing breaks; they simply always exercise the event-based sync. Coverage gap, separable work.
- **`smoke.undeployed` carries no `@smoke` tags**, so all four of its tests are nightly-only. That is why a 300000 ms hang in it went unnoticed; the same tag-lane gap already applies to `@healthcheck` in undeployed files.
- `dust.remote` has not been run under `DUST_SYNC=events` since the fallback change, so the switch back is verified for the helper-built wallets but not for these scenarios.

## Related

Running the healthchecks against projections made #646 reachable — the stale-UTXO/`gapRanges` defect. It is pre-existing, but #645's background repeat converts one bad pass into a permanent wedge. #646 has been updated with the wedge evidence.

### DIFF
--- a/packages/e2e-tests/src/tests/projectionsBasedSync.undeployed.test.ts
+++ b/packages/e2e-tests/src/tests/projectionsBasedSync.undeployed.test.ts
@@ -41,9 +41,9 @@ describe('Projections-based synchronisation model', () => {
   beforeEach(async () => {
     fixture = getFixture();
     fundedEventsSynced = await utils.initWalletWithSeed(seedFunded, fixture);
-    funded = await utils.initWalletWithSeed(seedFunded, fixture, 'schnorr', utils.projectionsDustSyncOptions);
+    funded = await utils.initWalletWithSeed(seedFunded, fixture, 'schnorr', utils.manualProjectionsDustSyncOptions);
     receiverEventsSynced = await utils.initWalletWithSeed(seed, fixture);
-    receiver = await utils.initWalletWithSeed(seed, fixture, 'schnorr', utils.projectionsDustSyncOptions);
+    receiver = await utils.initWalletWithSeed(seed, fixture, 'schnorr', utils.manualProjectionsDustSyncOptions);
     logger.info('Two wallets started');
   });
 
@@ -388,8 +388,8 @@ describe('Projections-based synchronisation model', () => {
       // the test keeps it hermetic: an undeployed chain is new each run, so a snapshot left behind by a previous
       // run would describe a different chain.
       const filename = `projections-restore-${seedFunded.substring(0, 7)}.state`;
-      await utils.saveState(funded.wallet, filename);
-      const restored = await utils.provideWallet(filename, seedFunded, fixture, utils.projectionsDustSyncOptions);
+      await utils.saveState(funded, filename);
+      const restored = await utils.provideWallet(filename, seedFunded, fixture, utils.manualProjectionsDustSyncOptions);
 
       try {
         // `provideWallet` silently falls back to building from scratch when a restore fails, which would leave this
--- a/packages/wallet-sdk-testkit/src/state-waiters.ts
+++ b/packages/wallet-sdk-testkit/src/state-waiters.ts
@@ -22,6 +22,60 @@ import { logger } from './logger.js';
 /** Default dust-balance threshold to wait for: 7 * 10^14. */
 export const DEFAULT_DUST_BALANCE_THRESHOLD = 7n * 10n ** 14n;
 
+/** How long a condition must hold before a settling waiter accepts it. */
+const DEFAULT_SETTLE_MS = 10_000;
+
+/** How long a settling waiter waits for its condition before failing rather than running to the test timeout. */
+const DEFAULT_SETTLE_TIMEOUT_MS = 180_000;
+
+/**
+ * Resolves with the wallet's current state once `predicate` has held continuously for `settleMs`.
+ *
+ * A settle window is needed because conditions like `pendingCoins.length === 0` are also the _resting_ state before a
+ * transaction is submitted: a waiter that resolved on the first matching emission would hand back the stale
+ * pre-transaction state.
+ *
+ * The window has to be applied to the predicate rather than to the source. Debouncing the source waits for the wallet
+ * to stop emitting altogether, which never happens while a sub-wallet syncs in the background — a Dust wallet syncing
+ * from projections emits about every 5 seconds indefinitely — so the predicate is never evaluated and the wait runs to
+ * the test timeout while logging that its condition is already satisfied.
+ *
+ * `distinctUntilChanged` is load-bearing: without it, `switchMap` would cancel and restart the timer on every emission
+ * and the timer would never elapse, reproducing the same starvation.
+ *
+ * The matching value is carried through the pipeline rather than re-read from `source` once the window elapses, because
+ * `source` must only be subscribed once. The facade's `state()` replays its current value on subscribe, but a
+ * sub-wallet's `state` does not: a second subscription there receives nothing until the next change, so on an
+ * already-settled wallet it would never produce a value and the wait would hang. The state returned is therefore the
+ * one from the moment the condition began to hold — callers that need the very latest should follow with their own
+ * read.
+ */
+export const waitForStableState = <T>(
+  source: rx.Observable<T>,
+  predicate: (value: T) => boolean,
+  description: string,
+  settleMs: number = DEFAULT_SETTLE_MS,
+  timeoutMs: number = DEFAULT_SETTLE_TIMEOUT_MS,
+): Promise<T> =>
+  rx.firstValueFrom(
+    source.pipe(
+      rx.map((value) => ({ value, holds: predicate(value) })),
+      rx.distinctUntilChanged((a, b) => a.holds === b.holds),
+      rx.switchMap(({ value, holds }) => (holds ? rx.timer(settleMs).pipe(rx.map(() => value)) : rx.NEVER)),
+      rx.take(1),
+      rx.timeout({
+        first: timeoutMs,
+        with: () =>
+          rx.throwError(
+            () =>
+              new Error(
+                `${description}: condition did not hold for ${settleMs}ms within ${timeoutMs}ms. The condition is genuinely unmet — this is not a missed observation window.`,
+              ),
+          ),
+      }),
+    ),
+  );
+
 export const waitForSyncUnshielded = (wallet: UnshieldedWallet) =>
   rx.firstValueFrom(
     wallet.state.pipe(
@@ -60,24 +114,19 @@ export const waitForFacadePending = (wallet: WalletFacade) =>
   );
 
 export const waitForFacadePendingClear = (wallet: WalletFacade) =>
-  rx.firstValueFrom(
+  waitForStableState(
     wallet.state().pipe(
       rx.tap((state) => {
-        const shieldedPending = state.shielded.pendingCoins.length;
-        logger.info(`Shielded wallet pending coins: ${shieldedPending}, waiting for pending coins to clear...`);
-        const unshieldedPending = state.unshielded.pendingCoins.length;
-        logger.info(`Unshielded wallet pending coins: ${unshieldedPending}, waiting for pending coins to clear...`);
-        const dustPending = state.dust.pendingCoins.length;
-        logger.info(`Dust wallet pending coins: ${dustPending}, waiting for pending coins to clear...`);
+        logger.info(
+          `Pending coins — shielded: ${state.shielded.pendingCoins.length}, unshielded: ${state.unshielded.pendingCoins.length}, dust: ${state.dust.pendingCoins.length}; waiting for all to clear...`,
+        );
       }),
-      rx.debounceTime(10_000),
-      rx.filter(
-        (state) =>
-          state.shielded.pendingCoins.length == 0 &&
-          state.unshielded.pendingCoins.length == 0 &&
-          state.dust.pendingCoins.length == 0,
-      ),
     ),
+    (state) =>
+      state.shielded.pendingCoins.length === 0 &&
+      state.unshielded.pendingCoins.length === 0 &&
+      state.dust.pendingCoins.length === 0,
+    'waitForFacadePendingClear',
   );
 
 export const waitForDustBalance = (wallet: WalletFacade, threshold: bigint = DEFAULT_DUST_BALANCE_THRESHOLD) =>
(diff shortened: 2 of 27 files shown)


packages/wallet-sdk-testkit/src/state-waiters.ts
```diff
@@ -22,6 +22,60 @@ import { logger } from './logger.js';
...
+  predicate: (value: T) => boolean,
+  description: string,
+  settleMs: number = DEFAULT_SETTLE_MS,
+  timeoutMs: number = DEFAULT_SETTLE_TIMEOUT_MS,
+): Promise<T> =>
+  rx.firstValueFrom(
+    source.pipe(
+      rx.map((value) => ({ value, holds: predicate(value) })),
+      rx.distinctUntilChanged((a, b) => a.holds === b.holds),
+      rx.switchMap(({ value, holds }) => (holds ? rx.timer(settleMs).pipe(rx.map(() => value)) : rx.NEVER)),
```
agronmurtezi: [P2] Return the latest observed state after the settle window

The predicate-only distinctUntilChanged discards newer values while holds remains true, and the timer returns the value captured when the condition first became true. This changes the behavior of waitForFinalizedShieldedBalance, whose existing callers (for example the swap test in tokenTransfer.remote.test.ts) assert balances directly from its result. A controlled-timer reproduction starts with pendingCoins=[] and balance 0, emits pendingCoins=[] and balance 10 five seconds later, and gets balance 0 from the new helper at ten seconds; the previous implementation returns balance 10. Incoming balance updates need not change pendingCoins, so the settle window can now return exactly the stale pre-transaction state it is intended to avoid. Keep the timer tied to predicate transitions, but track and return the latest value from the same subscription.

packages/e2e-tests/src/tests/projectionsBasedSync.undeployed.test.ts
```diff
@@ -42,9 +42,9 @@ describe('Projections-based synchronisation model', () => {
   beforeEach(async () => {
     fixture = getFixture();
     fundedEventsSynced = await utils.initWalletWithSeed(seedFunded, fixture);
-    funded = await utils.initWalletWithSeed(seedFunded, fixture, 'schnorr', utils.projectionsDustSyncOptions);
+    funded = await utils.initWalletWithSeed(seedFunded, fixture, 'schnorr', utils.manualProjectionsDustSyncOptions);
     receiverEventsSynced = await utils.initWalletWithSeed(seed, fixture);
-    receiver = await utils.initWalletWithSeed(seed, fixture, 'schnorr', utils.projectionsDustSyncOptions);
+    receiver = await utils.initWalletWithSeed(seed, fixture, 'schnorr', utils.manualProjectionsDustSyncOptions);
```
agronmurtezi: [P2] Pin the parity suite reference wallets to event-based sync

With DUST_SYNC=projections, initWalletWithSeed now selects the projections factory by default, so fundedEventsSynced and receiverEventsSynced also use projections. The explicitly configured subject wallets use projections too. I confirmed that these two setup paths both select dustSyncModel=projections under that environment setting. Consequently the projections-versus-events and restored-snapshot parity assertions compare the same sync implementation against itself, allowing shared projection defects to pass; this also affects the projections-mode verification reported in the PR description. Pass an explicit eventBasedDustWallet factory for both reference wallets so the comparison remains independent of DUST_SYNC.

agronmurtezi (commented):
Reviewed this PR against its stacked base 4285c23c (#637), excluding findings already reported on the parent PR. The 16 new testkit unit tests pass; targeted reproductions confirm the two issues below. Docker/live-network tests were not rerun.

packages/wallet-sdk-testkit/src/state-waiters.ts
jtsang586: Fixed in 250a14b8. Reproduced your case first: with `pendingCoins: []` and balance 0 at t=0 and
balance 10 at t=5s, the waiter returned balance 0 at t=10s.

The mapped source is now multicast with `shareReplay({ bufferSize: 1, refCount: true })`, so the
timer stays tied to predicate transitions while the value is read with `withLatestFrom` once the
window elapses. That keeps the single upstream subscription the old comment was careful about — a
sub-wallet's `state` replays nothing to a second reader, so an already-settled wallet would never
produce a value for it. The value returned still satisfies the predicate: one that did not would
have been a `holds` transition, and `switchMap` would have cancelled the timer before it fired.

Three tests cover it, and two mutations confirm they bite: reverting to the opening value fails the
newest-value test, and removing the window fails that plus the restart test.

While in the file I also found `waitForFinalizedShieldedBalance` in the e2e copy of these waiters
still on a bare `firstValueFrom(filter(...))` with no settle window at all — the same stale-state
defect you describe, in the copy the swap tests actually call. Migrated in 16119bd8; both copies
now settle on the same three waiters.

packages/e2e-tests/src/tests/projectionsBasedSync.undeployed.test.ts
jtsang586: Fixed in f6da42d0, and your analysis is confirmed against a live stack rather than by reading.

Both reference wallets now take an explicit `{ dustWallet: eventBasedDustWallet }`, and the suite
asserts the four models in `beforeEach` rather than assuming them — the pin is invisible at the call
site once the options are hoisted, and this is what makes the comparison a comparison. Removing the
pin and re-running under `DUST_SYNC=projections` fails with `expected 'projections' to be 'events'`,
which is your finding reproduced.

The suite passes 5/5 under both `DUST_SYNC=projections` and `DUST_SYNC=events`, in ~252s each.

packages/wallet-sdk-testkit/src/wallet.ts
```diff
@@ -163,8 +175,30 @@ const restoreDustWallet = async (
...
+ *
+ * @param walletOptions What the caller passed, if anything.
+ * @returns The options to hand {@link provideWallet}.
+ */
+export const withProjectionsDefault = (
+  walletOptions: Pick<ProvideWalletOptions, 'dustWallet' | 'manualSync'> | undefined,
+): Pick<ProvideWalletOptions, 'dustWallet' | 'manualSync'> => ({
+  dustWallet: dustWalletFromEnv(process.env, 'projections'),
+  ...walletOptions,
+});
```
agronmurtezi, line 193: [P2] Preserve the projections default for an undefined factory

Spreading walletOptions after the default lets an explicitly undefined dustWallet overwrite it. This is a valid input: ProvideWalletOptions explicitly allows undefined, so callers forwarding an optional factory can encounter it without a type error. With DUST_SYNC unset, I reproduced withProjectionsDefault({ manualSync: false }) selecting projections, while withProjectionsDefault({ manualSync: false, dustWallet: undefined }) produces an undefined factory; provideWallet then falls back to dustWalletFromEnv() and silently selects events. The healthcheck therefore stops exercising projections even though the caller did not select a different model. Spread the caller options first, then set dustWallet to walletOptions?.dustWallet ?? dustWalletFromEnv(process.env, 'projections'), and add a regression test for the explicit-undefined case.

jtsang586: Fixed in a2925b21, and reproduced before fixing: with `DUST_SYNC` unset,
`withProjectionsDefault({ manualSync: false })` resolved to `projections` while
`withProjectionsDefault({ manualSync: false, dustWallet: undefined })` resolved to `events`, exactly
as you describe.

The default is now filled in per field — caller options spread first, then
`dustWallet: walletOptions?.dustWallet ?? dustWalletFromEnv(process.env, 'projections')` — so an
explicit `undefined` counts as not supplying a factory rather than as choosing the other model. That
is the right reading of the input: the field is declared `| undefined` precisely so an optional
factory can be forwarded under `exactOptionalPropertyTypes`, and forwarding an absent one is the
absence of a choice.

agronmurtezi (approved):
Verified all findings from this review are addressed: settling waiters return the latest observed state, parity reference wallets remain pinned to event-based sync under both DUST_SYNC settings, and an explicitly undefined dustWallet preserves the projections fallback. Validation: all 27 testkit tests passed at this revision; five independent waiter and model-selection checks passed on the preceding fix revision. Docker/live-network tests were not rerun.

PR #650, opened 2026-08-11
docs(prover-client): fix config field name in README (serverUrl -> url)

Jc-asastu:
The `prover-client` README's basic-usage example and API signature reference a `serverUrl` config field:

```ts
new HttpProverClient({ serverUrl: 'http://localhost:6300' });
```

But `HttpProverClient` takes a `ProverClient.ServerConfig` whose field is `url` (`src/effect/ProverClient.ts` declares `readonly url`, and `src/effect/HttpProverClient.ts` reads `config.url`). `serverUrl` does not exist in the sources, so copying the documented example fails at runtime.

This aligns the README (both the example and the API signature) with the actual field name.

### DIFF
--- a/packages/prover-client/README.md
+++ b/packages/prover-client/README.md
@@ -22,7 +22,7 @@ import { HttpProverClient } from '@midnightntwrk/wallet-sdk-prover-client';
 
 // Initialize the client with the Proof Server URL
 const proverClient = new HttpProverClient({
-  serverUrl: 'http://localhost:6300',
+  url: 'http://localhost:6300',
 });
 
 // Prove an unproven transaction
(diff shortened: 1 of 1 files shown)


Jc-asastu:
Note: this contribution was AI-assisted (research and drafting)  Happy to have the ai-assisted label applied if that's your convention.

Jc-asastu:
Note: this contribution was AI-assisted. Every change was verified against the source before submitting. Happy to have the `ai-assisted` label applied if that's your convention — I can't set labels as an outside contributor.

agronmurtezi:
@Jc-asastu thanks for the contribution - can you please add a changeset to this PR?

PR #652, opened 2026-08-11
feat(shielded): restructure variants — v1 = ledger-v8 (restored), v2 = ledger-v9

agronmurtezi:
Phase 3 of the hard-fork plan: make the shielded wallet's directory structure the real product structure. The `v2` line had migrated `src/v1/` to ledger-v9 in place, leaving "v1" a misnomer; this PR renames that variant to `v2` and restores the pre-fork ledger-v8 variant from `main` as `src/v1/`. Stacked on #644.

## What changed

**Commit 1 — rename the ledger-v9 variant `v1` → `v2`.** Pure `git mv` + mechanical identifier rename (`V1Builder`→`V2Builder`, `V1Tag`→`V2Tag`, …), with every external `./v1`-subpath consumer repointed to the new `./v2` (they all meant the production v9 code). Serialized snapshots never embed the variant tag or any `V1` string, so restore of existing states is unaffected — verified against `Serialization.ts` before the rename.

**Commit 2 — restore the ledger-v8 variant as `src/v1/`.** Byte-for-byte `main`'s shielded internals (the tx-history #532 work was already on this branch via the July main sync), with exactly three adaptations to current sibling APIs:
- the `VariantContext` test stub supplies the now-required `activationRange`;
- the simulator is reached through the `V8` namespace (the simulation entry point's unqualified names are the v9 twin's), and the variant's `Simulator` re-export narrows to the v8 twin — **a semantics change on the `./v1` subpath**, called out in the changeset;
- `transacting.test.ts` gets a local ledger-v8 proving stub, since capabilities' `makeSimulatorProvingServiceEffect` is typed on ledger-v9 and a published v8 twin isn't warranted for one test file.

**Commit 3 — changeset + READMEs.** The changeset says loudly that the `./v1` subpath's **contents** switch on the beta line, and that `@midnight-ntwrk/ledger-v8` becomes a runtime dependency (second WASM blob for `./v1` consumers). `wallet-sdk` now exposes both public subpaths: `shielded/v1` (v8) and `shielded/v2` (v9).

## What deliberately did NOT change

- Production `ShieldedWallet` still registers **only** the v9 variant (`V2Builder` at `MinSupportedVersion`); root entry-point API and serialized-state format are untouched.
- No test assertions changed — all existing tests pass with only mechanical renames/requalifications (14 files: 7 restored v1 + 7 renamed v2; 108 passed, 22 pre-existing skips).
- The fork-test factory that registers both variants is deferred to Phase 4, where the tests that need it will define its shape (`CustomShieldedWallet`'s single-element HList types are the design work).

## Verification

- `yarn dist --force`, `yarn typecheck --force` (37/37), `yarn lint` (58/58), `yarn test:unit --force` (53/53)
- `yarn test --filter=@midnightntwrk/wallet-sdk-shielded --force` — 14 files, 108 passed / 22 skipped
- Effect diagnostics clean on all changed files except five pre-existing `deterministicKeys` hits inherited verbatim from `main`'s `WalletError.ts` (`Wallet.*` tags are runtime behavior; the renamed v2 twin carries the identical hits — same category as the accepted `LedgerTranslationError` one from #644)
- `yarn format:check`, `yarn changeset status --since=origin/feat/fork-state-translation` — green
- Commit 1 shows clean rename detection (all 20 moved files R-tracked, similarity 78–100%)


agronmurtezi:

PR #655, opened 2026-08-12
feat(shielded): fork boundary wiring and cross-version state migration

agronmurtezi:
Phase 4 of the hard-fork plan: the foundation wiring (design decisions D1–D5) on both shielded variants, plus the cross-version v8→v9 `migrateState`. Stacked on #652. TDD structure preserved: red-test commits precede each implementation commit.

> **Design notes.** History was rewritten once (2026-08-12) when the wallet team confirmed the indexer **replays events after the hard fork**, replacing an earlier carried-coin/re-anchoring approach. The cursor semantics were then confirmed the same day — **ids continue from whatever id the indexer was at when the fork happened, never from 0** — and landed as the final three additive commits (red-first): the migration **parks progress at the fork** instead of resetting it.

## What this delivers

**Boundary wiring (D2/D3/D5), both variants.** `SyncCapability.applyUpdate` gains an `activeRange` parameter; the split/annotate/monotone logic lives in pure exported helpers (`splitAtVersionBoundary`, `annotateVersion`) shared by the indexer capability (per event) and the simulator capability (per block), so the two paths cannot drift. State now records the indexer-reported `protocolVersion` monotonically (`CoreWallet.withProtocolVersion`); a batch straddling the boundary applies only the prefix and records the first deferred item's version, which is what triggers `VersionChange` → migration. This closes **Gap 3**.

**Variant stream.** `RunningV1Variant`/`RunningV2Variant` store `context.activationRange`, thread it into both `applyUpdate` call sites, and emit an immediate healing `VersionChange` when a restored state is already outside the variant's range.

**Migration seam (D4).** New `Migration.ts` per variant — `StateMigration` dictionary with `makeEmptyWalletMigration`, `makeCarryOverMigration`, and `makeCrossLedgerMigration`; the builder gains a defaulted `TPreviousState` generic plus `withMigration`/`withMigrationDefaults`, and `build()` wires `migrateState` through it.

**Cross-version migration = fresh-state projection with a parked cursor (the replay design, confirmed semantics).** The indexer replays events post-fork continuing from its fork-time position, so nothing coin-shaped crosses the migration: `CoreWallet.fromPreviousVersion` carries publicKeys, networkId, and **the previous sync progress** (parked at the fork; `isConnected` deliberately does not cross) into a fresh empty v9 `ZswapLocalState`. The new variant resumes at the parked cursor and re-discovers its coins by ordinary sync of the replayed v9 events, decrypting with the keys that arrive via the restart path. Keys never enter `migrateState` or serialized state.

**Restart wiring (D1), closing Gaps 1+2.** `ShieldedWallet.start()` retains the start-aux in a private `Option` and registers the Phase-1 `onVariantActivation` watcher once; the watcher re-dispatches `startSyncInBackground(retainedAux)` on the new variant; `stop()` clears the aux.

## Tests

Baseline 130 tests → 28 confirmed red at the red-test commits → 150 passed / 22 pre-existing skips at the tip (19 files), none weakened. The park-at-fork flip followed the same discipline: the flipped expectation went red against the reset implementation (`expected 0n to be 4321n` on both variant twins) before the one-line implementation change.

## Notable decisions

- `ChangesResult.protocolVersion` reports the last **applied** item's version, not the batch tail's — tx-history tags changes with the version they were applied under.
- The v1 twin carries the identical `Migration.ts` (inert there) to keep the trees mechanically comparable.
- No two-variant factory here — the watcher tests stub the 5-member `Runtime` interface directly; the factory lives in #656 with its consumers.

## Verification (branch tip, standalone)

`yarn dist --force` 17/17 · `typecheck --force` 37/37 · `lint` 58/58 · `test:unit --force` 53/53 · shielded suite 19 files / 150 passed / 22 skipped · Effect diagnostics 0/0/0 on all 22 changed files · `format:check` · changeset status clean. All commits GPG-signed.

### DIFF
--- a/.changeset/shielded-fork-boundary-wiring.md
+++ b/.changeset/shielded-fork-boundary-wiring.md
@@ -1,0 +1,49 @@
+---
+'@midnightntwrk/wallet-sdk-shielded': minor
+---
+
+Make the shielded wallet able to follow the chain across a hard fork: sync now recognises the protocol version the
+indexer reports, hands over at the variant boundary, and starts the next ledger version's variant on a fresh state.
+
+**Behavioural change: wallet state now tracks the reported protocol version.** Until now nothing wrote the indexer's
+`protocolVersion` into the wallet state, so it stayed pinned at its initial value and a protocol version change could
+never be signalled. It is now recorded on every applied batch. The value only ever increases — a source that briefly
+reports an older version (a reconnect replaying from an earlier cursor, say) cannot drag the wallet back below a
+boundary it has already crossed.
+
+**Breaking for custom `withSync` authors: `SyncCapability.applyUpdate` takes a third argument.** The signature is now
+`applyUpdate(state, update, activeRange)`, where `activeRange` is the half-open protocol version range the running
+variant owns, derived from `withVariant(sinceVersion)` registration. A custom capability must accept it and split its
+batch at the boundary: everything from the first item at or beyond the range's end belongs to the next variant and must
+be left unapplied, with the applied index stopping at the last item actually applied so that suffix is re-fetched rather
+than skipped. `splitAtVersionBoundary` and `annotateVersion` are exported from the `Sync` module and implement exactly
+that rule — the built-in indexer and simulator capabilities both go through them, and a custom capability should too.
+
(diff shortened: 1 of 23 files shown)


agronmurtezi:

PR #656, opened 2026-08-12
feat(shielded): byte-faithful fork-simulation proof and carried-coin re-anchoring

agronmurtezi:
Phase 5 of the hard-fork plan — the proof. A shielded wallet is driven through a simulated v8→v9 hard fork: it syncs the pre-fork chain, observes the version signal, migrates to a **fresh** v9 state with its **cursor parked at the fork**, restarts with retained keys, **re-discovers its coins from the replayed post-fork events**, and spends them. Stacked on #655.

> **Design notes.** The real indexer **replays events after the hard fork, continuing from whatever id it was at when the fork happened — never from 0** (wallet-team confirmed). The simulator has no indexer, so the harness models the replay by **transferring the coins on the post-fork timeline**: each pre-fork coin is re-paid as a v9 transaction with the exact same type/nonce/value/recipient (`src/test/forkReplay.ts`), on a chain **numbered as a continuation of the pre-fork chain** (block numbers stand in for indexer event ids — the same convention `ForkSimulator`'s translated chain already follows). A migrated wallet parked at the fork consumes exactly the replayed suffix.

## What the proof shows (per test)

**The crossing (unit tier — no toolchain needed, 3 tests):** v8 mints → v8 variant syncs → fork at block F → `VersionChange` → `migrateState` captured with the exact pre-fork state → the migrated state's cursor **equals the pre-fork cursor** (the park assertion, checked on the captured migration before any sync runs) → v2 restarts with retained keys → coins re-discovered from the replayed events above the boundary (asserted: the first consumed replay block is `> forkBlock`) → post-fork spend → balances correct. Negatives: a within-range version bump does not migrate; a fresh post-fork wallet syncs through the boundary and consumes the replay (scenario 2).

**The fidelity link (integration tier, real WASM translation, 3 tests):** the replay model and the ledger-team translation vouch for each other — translated-chain root == pre-fork wallet root; replay-chain root == translated root; **post-fork wallet root == translated root** (a wallet that never saw the translated state rebuilt that exact tree from replayed events), and its spend — Merkle path derived entirely from replay — is **accepted by the chain holding the translated ledger**.

## Mutation verification (all caught)

Force reset-to-0 in the migration → the park assertion fails on both migrating tests; corrupt a replayed coin's value → integration fails on root equality, unit strands below balance; replay numbered from 0 while parked → the wallet consumes nothing (this was the natural red when the confirmed semantics landed); corrupt a coin with structural checks stripped → the translated chain reports the spend as `'failure'`; swap the real translation for a blank one → the oracle equality itself fails (the fidelity test cannot pass vacuously).

**Honest limit, documented in `forkReplay.ts`:** beyond the captured-migration cursor assertion, this harness cannot behaviorally separate park from reset — the simulator capability filters `blocks.number >= appliedIndex`, so a cursor *behind* the replay still reads all of it, and the real-world harm of a reset (being re-served the pre-fork v8 events) is not representable against a post-fork source that has no blocks below the boundary. The state assertion is the discriminator, and forcing a reset fails it.

## Supporting work

- **Two-variant test factory** (`src/test/forkWallet.ts`, NOT exported): V1Builder at `MinSupportedVersion`, V2Builder at a per-test fork version. Its three documented workarounds (colliding merged variant configs; start-aux that cannot cross ledger versions — the harness retains the *seed* and derives per-variant keys, verified to yield identical commitments/nullifiers across v8/v9; the post-fork source not existing at build time) are direct input to the deferred public-API redesign.
- **This branch adds no production code** — under the replay design Phase 5 is purely proof. Its changeset is a patch (integration-tier plumbing only: `state-translation` devDependency + `turbo.json` `build:wasm` dependency + the integration workflow's toolchain gate extended to `packages/shielded-wallet/*`, so running shielded integration tests requires the Rust toolchain, mirroring capabilities).
- Consequence recorded in the plan: the indexer `zswapMerkleTreeCollapsedUpdate` codegen idea is moot for the wallet path — it only existed to serve re-anchoring.

## Verification (branch tip)

`yarn dist --force` · `typecheck --force` 37/37 · `lint` 58/58 · `test:unit --force` 53/53 · shielded suite 21 files / 156 passed / 22 pre-existing skips · integration 3/3 with the built artifact (re-run independently before pushing) · Effect diagnostics 0/0/0 on all 27 changed files · `format:check` · changeset status clean. All commits GPG-signed.

### DIFF
--- a/.changeset/shielded-fork-simulation-proof.md
+++ b/.changeset/shielded-fork-simulation-proof.md
@@ -1,0 +1,32 @@
+---
+'@midnightntwrk/wallet-sdk-shielded': patch
+---
+
+Prove the hard-fork crossing end to end, against real ledger bytes.
+
+A ledger-v8 chain pays a wallet, the pre-fork variant syncs it, the chain reaches the boundary height, and the wallet
+hands over to the ledger-v9 variant with a fresh state — then finds the same coins again by syncing the indexer's
+replayed timeline, and spends them. That is the corrected design stated as a test: nothing about the coins crosses the
+boundary, and everything the wallet ends up holding it re-earned through the sync path it uses every day. The two
+negatives are covered too: a version bump inside the running variant's range does not migrate, and a wallet whose very
+first sync already contains the fork reaches the same end state as one that lived through it.
+
+The replay is modelled with the thing the wallet's sync path already consumes — a second chain that re-announces every
+pre-fork commitment as a post-fork transaction. A commitment is a function of the coin and its owner's public key, so
+re-paying the same coin reproduces it exactly, and replaying the whole sequence in order reproduces the tree. No state
+translation is needed to model that, so the crossing is proven **unit tier**, with no toolchain.
+
+**The fidelity link is integration tier.** What a unit test cannot say is whether the model is faithful. So the same
+crossing runs again with the pre-fork chain handing its own serialized ledger to the ledger team's real v8-to-v9 state
+translation, and the two reconstructions are checked against each other: the tree the wallet rebuilds from replayed
+events has the same Merkle root as the tree the translation produced, and the chain holding the translated ledger
+accepts a spend whose Merkle path the wallet built from a tree it learned entirely from the replay. Both hold.
+
+That proof is integration tier because the translation is a WASM artifact built from `packages/state-translation`. This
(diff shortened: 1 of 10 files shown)


agronmurtezi:

Issue #658, opened 2026-08-13
Verifiable sync anchors for shielded and Dust sync modes (node API + protocol updates)


iangregsondev:
Sub-issue of #584, covering **items 2, 3 and 4** of the work sequence proposed in
[@kapke's triage comment](https://github.com/midnightntwrk/midnight-wallet/issues/584#issuecomment-5048997180).
Item 1 is tracked separately in #657.

## Describe the proposed feature:

Anchor "synced" to something the wallet can verify against consensus in the shielded and Dust sync
modes, rather than to the indexer's self-reported maximum. Unlike item 1, this requires
specification work and cross-repo orchestration (wallet + indexer + node) before any implementation
starts.

**Item 2 — node API for Dust tree roots at a given block.** The node already exposes the Zswap
commitment-tree root at a block (`midnight_zswapStateRoot`), which is what makes the shielded checks
below possible. There is no equivalent for the Dust trees. Both Dust sync modes depend on this
primitive, so it comes first.

**Item 3 — event-based synchronization protocol updates.** For shielded event-based sync, and by
analogy Dust event-based sync, the wallet should verify the expected commitment-tree root at
designated checkpoints, in addition to block recency and finalization. The approach is undecided:
extend the progress payload with checkpoint roots, have the indexer expose checkpoint roots, or let
a subscription end at a caller-specified block so the wallet can recompute its root there and
compare against the node.

**Item 4 — viewing-key-based and projection-based synchronization updates.** In these modes the
indexer can withhold data and cause the wallet to miss coins. The client still fetches
commitment-tree updates periodically, and those points are where the block status and the tree root
can be checked against the node. One wrinkle specific to viewing-key sync: the indexer may still be
building the relevance index for a particular wallet, which looks like lag. The check must
distinguish "indexer still catching up on my relevance index" from "indexer behind the finalized
head" — this is why the indexer reports several progress indices.

## User Need:

For shielded and Dust, a stronger guarantee than liveness is achievable: the wallet can recompute a
commitment-tree root and compare it against consensus. Without it, a compromised or
man-in-the-middled indexer can withhold events indefinitely while the wallet reports itself fully
synced, and integrators gating on `isSynced` under-credit against that view. Dust matters
additionally because a wallet that under-counts its Dust cannot correctly reason about fee capacity.

## Expected Benefit:

- Detects withholding, not only staleness — the wallet verifies data completeness against the chain
  rather than trusting the indexer's word.
[...]

Issue #657, opened 2026-08-13
Unshielded wallet: cross-check the indexer's reported tip against the node's finalized head


iangregsondev:
Sub-issue of #584, covering **item 1** of the work sequence proposed in
[@kapke's triage comment](https://github.com/midnightntwrk/midnight-wallet/issues/584#issuecomment-5048997180).

## Describe the proposed feature:

The unshielded wallet decides it is synced by comparing its applied transaction id against
`highestTransactionId` — a value the indexer reports about itself (`SyncProgress`, fed from `Sync.ts`).
Nothing reconciles that number against the chain, so an indexer that is stalled, lagging, or pinning
its reported maximum leaves the wallet reporting `isSynced === true` over a stale view.

Add a liveness check: periodically ask the node for its highest finalized block, compare it against
the latest block the indexer reports having processed, and surface a distinct verdict when the indexer
trails the finalized head instead of a silent `isSynced`.

Scope note from triage: the unshielded view is inherently indexer-scoped — progress is tracked per
address — so a liveness check is the only independent check available in this mode. Closing
address-targeted withholding would require a verifiable, address-indexed ledger storage model and is
out of scope here.

## User Need:

Anything gating on `isSynced` — a merchant flow, a payment integration, a UI showing a balance —
needs "synced" to mean "caught up with the chain", not "caught up with whatever this indexer last
said about itself". Today a stalled or hostile indexer under-credits the user silently: no error,
just a stale balance, with recently received funds missing entirely.

## Expected Benefit:

- A stalled, lagging, or withholding indexer becomes visible to the caller instead of silent.
- This is the one part of #584 actionable today — no indexer or node API changes required.
- It establishes the shape (wallet-side node cross-check, distinct staleness state) that items 3 and
  4 reuse for the shielded and Dust sync modes.

## Details:

**A finalized-head read on the node client.** `NodeClient.Service` exposed only
`sendMidnightTransaction` and `getGenesis`. It gains `getFinalizedBlock(): Effect<FinalizedBlock,
NodeClientError>`, returning the `{ hash, height }` of the highest block GRANDPA has finalized, read
over `chain_getFinalizedHead` + `chain_getHeader` by `PolkadotNodeClient` — which already reaches
`api.rpc.chain` — and `getGenesisHash(): Effect<string, NodeClientError>`, answered from state the
client already holds.

**A chain-identity check before any height comparison.** Heights from two different chains mean
[...]

Issue #584, opened 2026-07-22
Sync completion is measured against the indexer's self-reported tip, with no independent finality check


CharlesHoskinson:
**Component:** wallet sync (`SyncProgress`), all three wallet types
**Version:** `midnight-wallet` @ `e744d994`, indexer API v4
**Type:** hardening / defense in depth; not a vulnerability report, which is why this is a public issue rather than a private disclosure

I ran into this while building on the SDK. A wallet decides it has caught up to the tip by comparing its applied cursor against a value the indexer reports about itself, and nothing else. If that value is behind the real chain, the wallet still reports itself fully synced, silently.

On scope: the spec already places the indexer inside the trusted set. `docs/spec/Specification.md:740-745` (this repo) says node-fed sync is the best option for security and privacy, and that using an indexer comes "at the cost of having to trust said service." That framing is mostly about privacy and resource use, though. The completeness and liveness of the tip a wallet uses to decide it is "fully synced" is a different axis, one that can be trust-minimized on its own, and today nothing reconciles it against anything but the indexer itself.

### Details

The completion predicate:

- Shielded and dust: `packages/abstractions/src/SyncProgress.ts:31-33`, `isConnected && |highestRelevantWalletIndex - appliedIndex| <= maxGap` (strict `== 0` for "strictly complete", lines 65-66).
- Unshielded: `packages/unshielded-wallet/src/v1/SyncProgress.ts:29-31`, the same shape over `highestTransactionId - appliedId`.

The tip term in each predicate (`highestRelevantWalletIndex` / `highestTransactionId`) comes from the indexer subscription: `packages/shielded-wallet/src/v1/Sync.ts:293`, `packages/dust-wallet/src/v1/Sync.ts:330`, `packages/unshielded-wallet/src/v1/Sync.ts:112`. Its ultimate origin is a `MAX(id)` over the indexer's own database. That part lives in the midnight-indexer repo, not this one: `indexer-api/src/infra/storage/ledger_events.rs:92`, with the unshielded equivalent at `indexer-api/src/infra/storage/transaction.rs:522-544`. No finalized head or finality proof travels with it.

[...]

PR #659, opened 2026-08-13
feat(unshielded-wallet): cross-check the indexer's reported tip against the node's finalized head

iangregsondev:
# Description

Implements #657 (item 1 of #584).

The unshielded wallet decided it was synchronized by comparing its applied transaction id against
`highestTransactionId` — a number the indexer reports about itself. Nothing reconciled that against
the chain, so a stalled, lagging or withholding indexer left the wallet reporting `isSynced === true`
over a stale view: no error, just a balance missing recently received funds.

The wallet now periodically compares the indexer's latest block against the node's highest finalized
block, and reports itself unsynchronized while the indexer trails it, until the first comparison has
landed, or — permanently — when the two endpoints prove to be on different chains.

**This is on by default.** It reads the node already named by `relayURL` for submission, which the
facade requires, so no new configuration is needed and every existing integrator is affected.

**Scope:** unshielded only — #658 covers shielded and Dust, which need cross-repo protocol work. And
this detects staleness, not withholding: an indexer reporting the true block height while withholding
transaction events still passes. That needs commitment-tree root verification, also #658.

# Proposed Solution

An `IndexerLiveness` verdict on `SyncProgress`. Three of its seven cases gate the completion result —
`isStrictlyComplete()`, `isCompleteWithin()`, and therefore `waitForSyncedState()`:

| Verdict | Gates completion? | What it means |
| --- | --- | --- |
| `Behind` | **yes** | indexer is missing finalized blocks — the view is stale |
| `Unknown` | **yes** | no verdict yet — staleness is not yet ruled out |
| `WrongNetwork` | **yes** | the endpoints report different genesis blocks — they are on different chains |
| `Ahead` | no | indexer claims more than the finalized head — read skew or a lagging node |
| `Unavailable` | no | a poll could not complete — `lastError` names the failed read |
| `Skipped` | no | no node named at all, or a simulation — no verdict is ever coming |
| `InSync` | no | indexer within tolerance of the finalized head |

`Behind` gates because staleness is proven. `Unknown` gates because it is not yet ruled out: the
first verdict needs a node connection and a metadata download — seconds — while the indexer's first
progress update lands in well under one, so an ungated `Unknown` would let start-up race past the
check over exactly the stale view it exists to catch. It cannot gate forever: every path out of it is
bounded — the poll succeeds (a verdict), fails (`Unavailable`), or is cut off by the poll timeout
(`Unavailable` again), none of which gate. `WrongNetwork` gates because it is proven and permanent:
the genesis hashes are compared once, before any height comparison — heights from two different
chains mean nothing — and a mismatch cannot heal until someone reconfigures an endpoint. A failed
hash read pins nothing; it surfaces as `Unavailable` and is retried.

A "no" means the verdict leaves completion alone: the result is decided entirely by the conditions
that existed before this feature — whether the wallet is connected, and how much of the indexer's
stream it has applied. The verdict itself is always
readable on `progress.indexerLiveness`, carrying its detail — how far ahead, how many failed attempts
and the last error, how far behind — so every case can be surfaced to a user even when the boolean
says the wallet is complete. Gating on the inconclusive ones would make a node outage report a
healthy wallet as unsynchronized.

Layered so no wallet package depends on `node-client`: the verdict type and comparison in
`abstractions`, `getFinalizedBlock()` and `getGenesisHash()` on `node-client`, the polling service and
its height and genesis-hash reads in `capabilities`, and the wiring in `unshielded-wallet` — and the node client answers the genesis hash
from state it already holds, so the once-per-wallet chain-identity check costs no extra connection. Tolerances are separate per direction —
staleness can be generous, read skew must be tight — defaulting to 10 blocks each and a 30-second
interval — about a minute of staleness allowance at Midnight's 6-second slots — overridable via `livenessConfiguration` and `livenessPollInterval` on the configuration.
Each poll is bounded (20s default), so a read that hangs becomes `Unavailable` instead of freezing the
check; verdicts are published only when they change, so an idle wallet is not re-notified every poll;
and the node client behind the check is built once per wallet and reused, not once per poll.

Three pre-existing `node-client` defects had to be fixed for any of this to work. They were invisible
because both integration suites touching that client were skipped in their entirety:

- `api.rpc` calls failed after the client was created, so `getGenesis()` had never worked against a
  real node
- connection failures arrived as defects, bypassing the error channel the signature declares
- an unreachable node hung indefinitely and unboundably, which made `Unavailable` unreachable

Review hardening on top of those: a `connect()` that rejects mid-reconnect is retried rather than
escaping as a defect; a node that accepts the socket but persistently fails RPC surfaces as a typed
`ConnectionError` instead of spinning the readiness loop forever; a finite timeout beyond
`setTimeout`'s 32-bit range is clamped rather than overflowing into an instant failure; and
`isConnected` on the sync progress is cleared when the sync stream fails — previously it latched true
on the first update, so a wallet whose subscription dropped kept reporting itself synchronized
through the whole retry backoff. The liveness check cannot catch that last case (the indexer itself
may be healthy while this wallet's subscription is dead), which is why the connection flag has to.
A fourth: `main` landed its own reference-counting of the shared connection while this PR waited
(#649). This branch keeps its `SynchronizedRef` variant, which serialises acquire/release transitions
so a call arriving while the last one disconnects cannot race the socket closed — #649's regression
tests are retained and pass against it.
A fifth: the liveness feed now runs at wallet scope, decoupled from the indexer subscription, so it
keeps publishing verdicts — and keeps its node connection — while the subscription is down or
retrying; previously it was rebuilt on every retry and went silent in exactly those windows.

**Breaking — `unshielded-wallet` (major).** `isStrictlyComplete()` and `isCompleteWithin()` return
`false` while the indexer is known to be behind the node's finalized head, until the first verdict
has arrived, and — permanently — when the endpoints report different genesis blocks
(`WrongNetwork`). They previously considered only whether the wallet had applied everything the
indexer had sent it, so a wallet reading from a stalled indexer reported itself complete. Nothing is
required of the caller — the check runs by itself, and every wallet naming a `relayURL` gets it,
which the facade requires. Wallets with no node (and simulations) report `Skipped` and complete as
before.

`waitForSyncedState()` is built on the same predicate, so it does not resolve while the indexer is
behind. It does not reject or time out either; it waits until the indexer catches up. An integrator
gating start-up on it should race it against a deadline of their own and, on expiry, read
`progress.indexerLiveness` and `progress.isConnected` to say why the wallet has not settled — the
migration note in the changeset shows the pattern.

`SyncProgressData` gains a required `indexerLiveness` field, so anything building that object as a
literal must supply one — `createSyncProgress()` fills it in.

Sync updates gain one more kind: liveness verdicts, alongside the indexer's own.
The type naming its contents changed from `WalletSyncUpdate` to `SyncUpdate`, which shows up on
`UnshieldedWallet`, `UnshieldedWalletClass`, `DefaultV1Variant` and `DefaultRunningV1`. Everything
the old type described is still in the new one, so constructing or matching the indexer's updates
compiles unchanged — but code that handles every update type exhaustively will need a branch for the
new case.

**Breaking — `node-client` (major).** `NodeClient.Service` gains required `getFinalizedBlock()` and
`getGenesisHash()` methods. Callers are unaffected; implementers must add them. Connection and RPC failures are now typed rather
than defects, so `catchTag` reaches them — a caller whose `catchAll` previously let these through as
crashes will now catch them. Not flagged by the compiler: any `ApiPromise` double must now answer
`rpc.system.chain()`.

**Transaction submission is untouched.** The new connection timeout applies only when a caller asks
for one, and submission doesn't — so it still waits for an unreachable node to come back rather than
failing the submission. A test asserts that specifically, because inheriting the timeout would turn a
brief node outage into a failed transaction.

# Testing

`yarn test:unit` — the full unit suite, no infrastructure required.

`yarn test:integration` for `node-client`, `capabilities` and `unshielded-wallet` — needs Docker and
`APP_INFRA_SECRET`. The one that matters most starts a real wallet against a real indexer and node
with only `relayURL` configured, and waits for the verdict to actually become `InSync`. Another
proves the two endpoints' real genesis-hash presentations compare as the same chain — the assumption
`WrongNetwork` rests on.

## Gaps a reviewer should know about

- **Transaction submission still has no working integration coverage.** The submission tests in
  `node-client` and `capabilities` need a generated fixture that nothing produces — no generator, no
  CI job, gitignored output. Pre-existing, and the reason the `node-client` defects survived. The
  tests that passed vacuously on the missing fixture are now skipped with the reason recorded.

Refs #657, #584, #658

### DIFF
--- a/.changeset/indexer-liveness-cross-check.md
+++ b/.changeset/indexer-liveness-cross-check.md
@@ -21,7 +21,8 @@ is reported as `Skipped` (`no-liveness-feed`), so it can still reach "synced". T
 
 ### Fixes
 
-- `isConnected` on the sync progress clears when the indexer subscription drops; it previously latched `true`.
+- `isConnected` on the sync progress clears when the indexer subscription drops or is completed by the indexer, and the
+  subscription is rebuilt in both cases; it previously latched `true`, and a completed subscription was never rebuilt.
 - `api.rpc` calls (`getGenesis()`) work after client creation; they previously failed as disconnected.
 - Node connection failures are typed errors reachable by `catchTag`/`catchAll`, no longer defects.
 - A finite `reconnectionTimeout` also bounds the initial connection; no timeout (submission) stays unbounded.
--- a/.changeset/indexer-liveness-cross-check.md
+++ b/.changeset/indexer-liveness-cross-check.md
@@ -25,6 +25,9 @@ is reported as `Skipped` (`no-liveness-feed`), so it can still reach "synced". T
 - `api.rpc` calls (`getGenesis()`) work after client creation; they previously failed as disconnected.
 - Node connection failures are typed errors reachable by `catchTag`/`catchAll`, no longer defects.
 - A finite `reconnectionTimeout` also bounds the initial connection; no timeout (submission) stays unbounded.
+- The sync streams' retry backoff is genuinely capped at two minutes. The cap was applied to the schedule's output
+  rather than its interval, so retries kept doubling — half an hour apart by the twelfth — until the timer overflowed
+  and the stream stopped retrying.
 
 BREAKING CHANGE (`wallet-sdk`, `wallet-sdk-facade`, `wallet-sdk-unshielded-wallet`): `isStrictlyComplete()`,
 `isCompleteWithin()`, `FacadeState.isSynced` and `waitForSyncedState()` now also require the indexer not to trail the
--- a/.changeset/indexer-liveness-cross-check.md
+++ b/.changeset/indexer-liveness-cross-check.md
@@ -25,7 +25,9 @@ is reported as `Skipped` (`no-liveness-feed`), so it can still reach "synced". T
   subscription is rebuilt in both cases; it previously latched `true`, and a completed subscription was never rebuilt.
 - `api.rpc` calls (`getGenesis()`) work after client creation; they previously failed as disconnected.
 - Node connection failures are typed errors reachable by `catchTag`/`catchAll`, no longer defects.
-- A finite `reconnectionTimeout` also bounds the initial connection; no timeout (submission) stays unbounded.
+- A finite `reconnectionTimeout` also bounds the initial connection — the handshake and the close that follows it — and
+  the liveness poll's own timeout is hard even when a read is stuck in that build; no timeout (submission) stays
+  unbounded.
 - The sync streams' retry backoff is genuinely capped at two minutes. The cap was applied to the schedule's output
   rather than its interval, so retries kept doubling — half an hour apart by the twelfth — until the timer overflowed
   and the stream stopped retrying.
--- a/.changeset/indexer-liveness-cross-check.md
+++ b/.changeset/indexer-liveness-cross-check.md
@@ -30,8 +30,9 @@ first reached (`IndexerLiveness.equivalent` is the predicate). Tune with `livene
 - Node connection failures are typed errors reachable by `catchTag`/`catchAll`, no longer defects.
 - A finite `reconnectionTimeout` also bounds the initial connection — the handshake and the close that follows it — and
   the whole window is used: a node that is restarting connects on a later attempt rather than failing the build on the
-  first socket error. The liveness poll's own timeout is hard even when a read is stuck in that build; no timeout
-  (submission) stays unbounded.
+  first socket error. The same wait for the close, under the same bound, runs when the last in-flight call releases the
+  socket, so the next call does not start on a stale connection flag. The liveness poll's own timeout is hard even when
+  a read is stuck in that build; no timeout (submission) stays unbounded.
 - The sync streams' retry backoff is genuinely capped at two minutes. The cap was applied to the schedule's output
   rather than its interval, so retries kept doubling — half an hour apart by the twelfth — until the timer overflowed
   and the stream stopped retrying.
--- a/.changeset/indexer-liveness-cross-check.md
+++ b/.changeset/indexer-liveness-cross-check.md
@@ -26,8 +26,9 @@ is reported as `Skipped` (`no-liveness-feed`), so it can still reach "synced". T
 - `api.rpc` calls (`getGenesis()`) work after client creation; they previously failed as disconnected.
 - Node connection failures are typed errors reachable by `catchTag`/`catchAll`, no longer defects.
 - A finite `reconnectionTimeout` also bounds the initial connection — the handshake and the close that follows it — and
-  the liveness poll's own timeout is hard even when a read is stuck in that build; no timeout (submission) stays
-  unbounded.
+  the whole window is used: a node that is restarting connects on a later attempt rather than failing the build on the
+  first socket error. The liveness poll's own timeout is hard even when a read is stuck in that build; no timeout
+  (submission) stays unbounded.
 - The sync streams' retry backoff is genuinely capped at two minutes. The cap was applied to the schedule's output
   rather than its interval, so retries kept doubling — half an hour apart by the twelfth — until the timer overflowed
   and the stream stopped retrying.
--- a/.changeset/indexer-liveness-cross-check.md
+++ b/.changeset/indexer-liveness-cross-check.md
@@ -15,8 +15,9 @@ head (every 30 seconds by default) and, once, checks that both endpoints name th
 result is an `IndexerLiveness` verdict on `SyncProgress`: `Behind`, `Unknown` and `WrongNetwork` block completion; the
 other four (`InSync`, `Ahead`, `Unavailable`, `Skipped`) do not. A poll that fails after a `Behind` or `WrongNetwork`
 verdict leaves that verdict in place — an unreachable endpoint disproves nothing — so a node outage cannot release a
-caller waiting on a stale indexer. Tune with `livenessConfiguration` and `livenessPollInterval`. This detects
-staleness, not withholding.
+caller waiting on a stale indexer. A custom sync service supplied through `withSync` that exposes no `livenessUpdates`
+is reported as `Skipped` (`no-liveness-feed`), so it can still reach "synced". Tune with `livenessConfiguration` and
+`livenessPollInterval`. This detects staleness, not withholding.
 
 ### Fixes
 
--- a/.changeset/indexer-liveness-cross-check.md
+++ b/.changeset/indexer-liveness-cross-check.md
@@ -1,0 +1,35 @@
+---
+'@midnightntwrk/wallet-sdk': major
+'@midnightntwrk/wallet-sdk-facade': major
+'@midnightntwrk/wallet-sdk-unshielded-wallet': major
+'@midnightntwrk/wallet-sdk-node-client': major
+'@midnightntwrk/wallet-sdk-abstractions': minor
+'@midnightntwrk/wallet-sdk-capabilities': minor
+---
+
+feat(unshielded-wallet): cross-check the indexer's reported tip against the node's finalized head
+
+The wallet now verifies "synced" against the chain instead of the indexer's self-report: it polls a node's finalized
+head (every 30 seconds by default) and, once, checks that both endpoints name the same genesis block. The node is
+`nodeClientConnection` on the configuration, falling back to `relayURL`; a wallet naming neither is not checked. The
+result is an `IndexerLiveness` verdict on `SyncProgress`: `Behind`, `Unknown` and `WrongNetwork` block completion; the
+other four (`InSync`, `Ahead`, `Unavailable`, `Skipped`) do not. Tune with `livenessConfiguration` and
+`livenessPollInterval`. This detects staleness, not withholding.
+
+### Fixes
+
+- `isConnected` on the sync progress clears when the indexer subscription drops; it previously latched `true`.
+- `api.rpc` calls (`getGenesis()`) work after client creation; they previously failed as disconnected.
+- Node connection failures are typed errors reachable by `catchTag`/`catchAll`, no longer defects.
+- A finite `reconnectionTimeout` also bounds the initial connection; no timeout (submission) stays unbounded.
+
+BREAKING CHANGE (`wallet-sdk`, `wallet-sdk-facade`, `wallet-sdk-unshielded-wallet`): `isStrictlyComplete()`,
+`isCompleteWithin()`, `FacadeState.isSynced` and `waitForSyncedState()` now also require the indexer not to trail the
+finalized head, a matching genesis, and a first verdict (`Skipped` — no node, or a simulation — does not block). On by
+default for every wallet with a `relayURL`. Against a stale indexer, `waitForSyncedState()` waits — it neither rejects
+nor times out; race it against a deadline of your own and read `progress.indexerLiveness` and `progress.isConnected`
+(the `indexer-liveness` docs snippet shows the pattern). `SyncProgressData` gains a required `indexerLiveness` field,
+defaulted by `createSyncProgress()`; sync types are parameterised on `SyncUpdate`, a superset of `WalletSyncUpdate`.
+
+BREAKING CHANGE (`wallet-sdk-node-client`): `NodeClient.Service` gains required `getFinalizedBlock()` and
+`getGenesisHash()` methods. Callers are unaffected; implementers must add them.
--- a/.changeset/indexer-liveness-cross-check.md
+++ b/.changeset/indexer-liveness-cross-check.md
@@ -16,7 +16,10 @@ result is an `IndexerLiveness` verdict on `SyncProgress`: `Behind`, `Unknown` an
 other four (`InSync`, `Ahead`, `Unavailable`, `Skipped`) do not. A poll that fails after a `Behind` or `WrongNetwork`
 verdict leaves that verdict in place — an unreachable endpoint disproves nothing — so a node outage cannot release a
 caller waiting on a stale indexer. A custom sync service supplied through `withSync` that exposes no `livenessUpdates`
-is reported as `Skipped` (`no-liveness-feed`), so it can still reach "synced". Tune with `livenessConfiguration` and
+is reported as `Skipped` (`no-liveness-feed`), so it can still reach "synced". A verdict is republished only when it
+says something new — its kind, a `Behind`'s lag, an `Unavailable`'s failure count — not because the chain advanced, so
+an idle wallet's state does not change once per poll; the heights on a published `InSync` are those at which it was
+first reached (`IndexerLiveness.equivalent` is the predicate). Tune with `livenessConfiguration` and
 `livenessPollInterval`. This detects staleness, not withholding.
 
 ### Fixes
--- a/.changeset/indexer-liveness-cross-check.md
+++ b/.changeset/indexer-liveness-cross-check.md
@@ -9,7 +9,7 @@
 
 feat(unshielded-wallet): cross-check the indexer's reported tip against the node's finalized head
 
-The wallet now verifies "synced" against the chain instead of the indexer's self-report: it polls a node's finalized
+The unshielded wallet now verifies "synced" against the chain instead of the indexer's self-report: it polls a node's finalized
 head (every 30 seconds by default) and, once, checks that both endpoints name the same genesis block. The node is
 `nodeClientConnection` on the configuration, falling back to `relayURL`; a wallet naming neither is not checked. The
 result is an `IndexerLiveness` verdict on `SyncProgress`: `Behind`, `Unknown` and `WrongNetwork` block completion; the
@@ -20,12 +20,15 @@ is reported as `Skipped` (`no-liveness-feed`), so it can still reach "synced". A
 says something new — its kind, a `Behind`'s lag, an `Unavailable`'s failure count — not because the chain advanced, so
 an idle wallet's state does not change once per poll; the heights on a published `InSync` are those at which it was
 first reached (`IndexerLiveness.equivalent` is the predicate). Tune with `livenessConfiguration` and
-`livenessPollInterval`. This detects staleness, not withholding.
+`livenessPollInterval`. This detects staleness, not withholding. The shielded and dust wallets are not gated: their
+shared `SyncProgress` carries no verdict, and `FacadeState.isSynced` gates only through the unshielded wallet's progress
+(#743).
 
 ### Fixes
 
-- `isConnected` on the sync progress clears when the indexer subscription drops or is completed by the indexer, and the
-  subscription is rebuilt in both cases; it previously latched `true`, and a completed subscription was never rebuilt.
+- `isConnected` on the unshielded wallet's sync progress clears when the indexer subscription drops or is completed by
+  the indexer, and the subscription is rebuilt in both cases; it previously latched `true`, and a completed
+  subscription was never rebuilt. The shielded and dust wallets still latch it (#743).
 - `api.rpc` calls (`getGenesis()`) work after client creation; they previously failed as disconnected.
 - Node connection failures are typed errors reachable by `catchTag`/`catchAll`, no longer defects.
 - A finite `reconnectionTimeout` also bounds the initial connection — the handshake and the close that follows it — and
@@ -33,13 +36,14 @@ first reached (`IndexerLiveness.equivalent` is the predicate). Tune with `livene
   first socket error. The same wait for the close, under the same bound, runs when the last in-flight call releases the
   socket, so the next call does not start on a stale connection flag. The liveness poll's own timeout is hard even when
   a read is stuck in that build; no timeout (submission) stays unbounded.
-- The sync streams' retry backoff is genuinely capped at two minutes. The cap was applied to the schedule's output
-  rather than its interval, so retries kept doubling — half an hour apart by the twelfth — until the timer overflowed
-  and the stream stopped retrying.
+- The unshielded wallet's sync streams' retry backoff is genuinely capped at two minutes. The cap was applied to the
+  schedule's output rather than its interval, so retries kept doubling — half an hour apart by the twelfth — until the
+  timer overflowed and the stream stopped retrying. The shielded and dust wallets still carry the old schedule (#742).
 
-BREAKING CHANGE (`wallet-sdk`, `wallet-sdk-facade`, `wallet-sdk-unshielded-wallet`): `isStrictlyComplete()`,
-`isCompleteWithin()`, `FacadeState.isSynced` and `waitForSyncedState()` now also require the indexer not to trail the
-finalized head, a matching genesis, and a first verdict (`Skipped` — no node, or a simulation — does not block). On by
+BREAKING CHANGE (`wallet-sdk`, `wallet-sdk-facade`, `wallet-sdk-unshielded-wallet`): the unshielded wallet's
+`isStrictlyComplete()` and `isCompleteWithin()`, and through its progress `FacadeState.isSynced` and the facade's
+`waitForSyncedState()`, now also require the indexer not to trail the finalized head, a matching genesis, and a first
+verdict (`Skipped` — no node, or a simulation — does not block). On by
 default for every wallet with a `relayURL`. Against a stale indexer, `waitForSyncedState()` waits — it neither rejects
 nor times out; race it against a deadline of your own and read `progress.indexerLiveness` and `progress.isConnected`
 (the `indexer-liveness` docs snippet shows the pattern). `SyncProgressData` gains a required `indexerLiveness` field,
--- a/.changeset/indexer-liveness-cross-check.md
+++ b/.changeset/indexer-liveness-cross-check.md
@@ -13,8 +13,10 @@ The wallet now verifies "synced" against the chain instead of the indexer's self
 head (every 30 seconds by default) and, once, checks that both endpoints name the same genesis block. The node is
 `nodeClientConnection` on the configuration, falling back to `relayURL`; a wallet naming neither is not checked. The
 result is an `IndexerLiveness` verdict on `SyncProgress`: `Behind`, `Unknown` and `WrongNetwork` block completion; the
-other four (`InSync`, `Ahead`, `Unavailable`, `Skipped`) do not. Tune with `livenessConfiguration` and
-`livenessPollInterval`. This detects staleness, not withholding.
+other four (`InSync`, `Ahead`, `Unavailable`, `Skipped`) do not. A poll that fails after a `Behind` or `WrongNetwork`
+verdict leaves that verdict in place — an unreachable endpoint disproves nothing — so a node outage cannot release a
+caller waiting on a stale indexer. Tune with `livenessConfiguration` and `livenessPollInterval`. This detects
+staleness, not withholding.
 
 ### Fixes
 
--- a/packages/abstractions/src/IndexerLiveness.ts
+++ b/packages/abstractions/src/IndexerLiveness.ts
@@ -1,0 +1,288 @@
+// This file is part of MIDNIGHT-WALLET-SDK.
+// Copyright (C) Midnight Foundation
+// SPDX-License-Identifier: Apache-2.0
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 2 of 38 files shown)


agronmurtezi (commented):
Review of 6e9701d5 against `main`.

Inline comments cover the correctness, test and convention findings that anchor to changed lines. One finding has no anchor in the diff:

**The commit message is not marked breaking.** The changeset bumps four packages `major` with two `BREAKING CHANGE` notes, but the commit is `feat(unshielded-wallet): …` with neither `!` after the scope nor a `BREAKING CHANGE:` footer (CLAUDE.md → Git & GitHub Conventions). Conventional-commit tooling and `git log` readers will not classify it as breaking. Suggested fix before merge: `feat(unshielded-wallet)!: …` plus a `BREAKING CHANGE:` footer.

Two candidates were checked and refuted, so they are not raised: the `catchAllCause` in `#poll` writing a bogus `Unavailable` on interruption (the Effect runtime skips failure continuations on an interrupted fiber), and the three-probe cap regressing submission (`main` already failed the first real RPC immediately on a connected-but-broken node, so the new code is strictly more tolerant).

This review was produced with AI assistance.

packages/abstractions/src/IndexerLiveness.ts
```diff
@@ -0,0 +1,288 @@
...
+ *   ```ts
+ *   const next = IndexerLiveness.afterFailedPoll(previous, 'websocket closed');
+ *   ```;
+ *
+ * @param previous - The verdict before this poll.
+ * @param lastError - The message from the failure that just occurred.
+ * @returns An {@link IndexerLiveness.Unavailable} verdict whose count continues the previous run of failures, or starts
+ *   a new one.
+ */
+export const afterFailedPoll = (previous: IndexerLiveness, lastError: string): IndexerLiveness =>
```
agronmurtezi, line 306: **A single failed poll opens the sync gate.** `afterFailedPoll` overwrites a proven `Behind` with `Unavailable`, which does not gate. A node timeout, or the indexer answering `block: null`, therefore lets a pending `waitForSyncedState()` resolve over a stale indexer, and a fresh, empty indexer can report fully synced after its first poll.

Suggested fix: keep the previous verdict when it was `Behind` or `WrongNetwork` (recording the failure alongside it if the count is wanted) rather than replacing it. No test currently starts from `Behind` and fails a poll; one would pin this down.

packages/unshielded-wallet/src/v1/SyncProgress.ts
```diff
@@ -28,7 +40,11 @@ export interface SyncProgress extends SyncProgressData {
 export const SyncProgress: SyncProgressOps = {
   isCompleteWithin(data: SyncProgressData, maxGap: bigint = 50n): boolean {
     const applyLag = BigInt(Math.abs(Number(data.highestTransactionId - data.appliedId)));
-    return data.isConnected && applyLag <= maxGap;
+    // `applyLag` only shows whether the wallet has kept up with the indexer; both of its terms come from the indexer, so
+    // it cannot reveal an indexer that is itself stale. That is what the liveness verdict adds: `Behind` blocks because
+    // staleness is proven, and `Unknown` blocks because it is not yet ruled out — the first verdict takes seconds, and
+    // an ungated `Unknown` would let start-up race past the check over exactly the stale view it exists to catch.
+    return data.isConnected && applyLag <= maxGap && !IndexerLiveness.blocksSyncCompletion(data.indexerLiveness);
```
agronmurtezi, line 47: **Custom sync services can never report synced.** `isCompleteWithin` now blocks on the `Unknown` default, and the only writers that move a wallet off `Unknown` are the default service's `livenessUpdates` and the simulator capability's hand-written `Skipped`. A wallet built through the public `V1Builder.withSync` with a service that lacks `livenessUpdates` compiles, but `waitForSyncedState()` and `facade.isSynced` hang silently with no error or log. The JSDoc at `Sync.ts:213` says `createSyncProgress` defaults to `Skipped`; it defaults to `Unknown`.

Suggested fix: have the variant write `Skipped` once when the service exposes no `livenessUpdates`, so the decision lives where the service is inspected, and correct the JSDoc.

packages/unshielded-wallet/src/v1/RunningV1Variant.ts
```diff
@@ -118,10 +121,58 @@ export class RunningV1Variant<TSerialized, TSyncUpdate> implements Variant.Runni
...
+    return SubscriptionRef.updateEffect(this.#context.stateRef, (state) =>
+      pipe(this.#v1Context.syncCapability.applyUpdate(state, update), EitherOps.toEffect),
+    );
+  }
+
+  /** Exponential backoff with jitter, capped at two minutes — shared by both sync streams. */
+  static #retrySchedule(): Schedule.Schedule<Duration.Duration, unknown> {
+    return pipe(
+      Schedule.exponential(Duration.seconds(1), 2),
+      Schedule.map((delay) => {
```
agronmurtezi: **The backoff cap is a no-op.** `Schedule.map` transforms the schedule's output value, not its recurrence interval, so both the indexer stream and the newly wired liveness feed back off 1 s, 2 s, 4 s… without bound and stop retrying once the interval exceeds `2^31 - 1` ms. Verified with `Schedule.run`: retry 8 outputs 120000 ms but the real interval is 128000 ms; retry 12 is 34 min; retry 23 never fires. Pre-existing on `main`, but this PR extracts the schedule, applies it to a second stream, and adds the comment claiming a two-minute cap.

Suggested fix: `Schedule.exponential('1 second').pipe(Schedule.union(Schedule.spaced('2 minutes')), Schedule.jittered)`, or `Schedule.modifyDelay`.

packages/unshielded-wallet/src/v1/RunningV1Variant.ts
```diff
@@ -130,24 +181,20 @@ export class RunningV1Variant<TSerialized, TSyncUpdate> implements Variant.Runni
...
-
-            return Duration.millis(Math.min(delayWithJitter, Duration.toMillis(maxDelay)));
-          }),
+      // The flag is written true on every progress update and nowhere else goes false, so without this it latches: a
+      // wallet whose subscription just died would sit through the retry backoff — or an entire indexer outage — still
+      // holding isConnected: true, a caught-up cursor, and its last liveness verdict, and so still reporting itself
+      // synchronized. The liveness check cannot catch that case: the indexer itself may be healthy while this wallet's
+      // subscription is dead. The connection flag is the one truthful signal, and the stream failing is the one place
+      // that knows it.
+      Stream.tapError(() =>
```
agronmurtezi, line 194: **A clean subscription end still latches `isConnected: true`.** This `tapError` fires only on failure. When the indexer sends a graphql-ws `complete` (which it does when its internal `UnshieldedUtxoIndexed` stream ends), the sink calls `emit.end()`, the stream finishes cleanly, neither `tapError` runs, `Stream.retry` never engages, and the background fiber is gone while `isStrictlyComplete()` keeps returning `true` and new transactions are never applied. The silent exit is pre-existing, but the changeset now states that `isConnected` clears when the subscription drops, which is not true for this path; the new test covers only a failing service.

Suggested fix: treat an unexpected `complete` as a failure so the retry and the flag reset both run, and cover it with a test.

packages/capabilities/src/liveness/livenessReads.ts
```diff
@@ -0,0 +1,290 @@
...
+     *   once.
+     *
+     *   Uninterruptible, because the poll fails fast: when the sibling indexer read fails, the in-flight node read is
+     *   interrupted. The client's acquire is uninterruptible on its own, so without this the interrupt would land
+     *   between the acquire completing and the cache write — stranding a fully-built client on the sync-lifetime scope
+     *   with the cache still empty, once per poll for as long as the indexer stays down. Deferring the interrupt past
+     *   the write costs at most the build's own ten-second connection bound.
+     */
+    const client = (nodeURL: URL): Effect.Effect<LivenessNodeReader, NodeClientError.NodeClientError> =>
+      Effect.uninterruptible(
```
agronmurtezi, line 191: **The 10 s read bound and 20 s poll timeout are soft.** `Effect.timeoutFail` awaits the interrupted loser, this region is `Effect.uninterruptible`, and the client build's `acquireRelease` is uninterruptible too. Inside `make()` only `ApiPromise.create` is raced; the following `await api.once('disconnected')` has no timer and is capped only by ws's default 30 s close timeout. Measured: `timeoutFail(100ms)` over `Effect.uninterruptible(sleep 2s)` returned at ~2 s. A node that completes the handshake and then goes half-open can publish its verdict ~40 s after the tick, delaying the next one. The comments at lines 30–31 and 180–181 claiming a read cannot outlive the tick are therefore inaccurate.

Suggested fix: bound the disconnect wait inside `make()` as well, and either drop `uninterruptible` here or document the real worst case.

packages/node-client/src/effect/PolkadotNodeClient.ts
```diff
@@ -54,34 +66,92 @@ export class PolkadotNodeClient implements NodeClient.Service {
...
+      // `autoConnectMs` keeps its default deliberately. Passing `false` does not mean "connect once without retrying" —
+      // it means "do not connect at all", leaving `ApiPromise.create` waiting on a connection nobody started. The bound
+      // below is what limits the wait; the provider's own retry behaviour is left alone.
+      const provider = new WsProvider(config.nodeURL.toString());
+
+      const created = ApiPromise.create({
+        // @ts-expect-error -- exactOptionalPropertyTypes cause an incompatibility here
+        provider,
+        // Surfacing a connection error rather than retrying past it, but only for a caller that asked to be bounded.
+        throwOnConnect: isBounded,
```
agronmurtezi: **Bounded connect fails on the first socket error.** `throwOnConnect: true` makes `ApiPromise.create` resolve `isReadyOrError`, which rejects on the first provider `error` event (any socket error, e.g. `ECONNREFUSED`). A finite `reconnectionTimeout` therefore means "fail within milliseconds, no retry" here, while `ensureConnection` with the same value retries until the deadline, and the changeset says the timeout "bounds the initial connection". A node restarting for two seconds produces an immediate `Unavailable`, where the old `throwOnConnect: false` path would have connected on the 2.5 s retry inside the window.

Suggested fix: keep `throwOnConnect: false` and let the `Promise.race` timer be the bound. The `boundedConnect` unit test mocks `create` as a pure hang, so error-then-success is untested.

packages/node-client/src/effect/PolkadotNodeClient.ts
```diff
@@ -187,34 +288,73 @@ export class PolkadotNodeClient implements NodeClient.Service {
...
-            cause: error,
-          }),
-      ),
-      Effect.ensuring(this.#release()),
+    return Effect.acquireUseRelease(
+      this.#register(),
+      () =>
+        pipe(
+          this.ensureConnection(),
+          Effect.andThen(() => Effect.promise(() => this.api.rpc.chain.getBlock(this.api.genesisHash))),
```
agronmurtezi: **`getGenesis()` still produces defects.** This `Effect.promise` turns an RPC rejection into `Cause.Die`, which bypasses the `mapError` two lines below and any `catchTag('ConnectionError')` in the caller, contradicting the changeset's "node connection failures are typed errors, no longer defects" and the `tryPromise` pattern this PR applied in `getFinalizedBlock()` and `ensureConnection()`. All lifecycle tests mock `getBlock` as resolving.

Suggested fix: `Effect.tryPromise`, plus a lifecycle test where `getBlock` rejects.

packages/capabilities/src/liveness/livenessService.ts
```diff
@@ -0,0 +1,300 @@
...
+
+  state(): Stream.Stream<IndexerLiveness.IndexerLiveness> {
+    // `changes` already replays the current value before subsequent ones, under the ref's own semaphore. Prepending a
+    // separate `get` would emit the first verdict twice — and read it outside that semaphore.
+    //
+    // `Stream.changes` then drops consecutive equal verdicts. Every poll writes unconditionally, and each write fans
+    // out into the wallet's full state stream — without this, a healthy idle wallet re-notified every subscriber once
+    // per poll, forever. Verdicts are structural data, so a lengthening outage still reports: `Unavailable`'s failure
+    // count climbs, which makes consecutive verdicts unequal.
+    return this.#state.changes.pipe(Stream.changes);
```
agronmurtezi: **This dedup cannot work in steady state.** `InSync`, `Behind` and `Ahead` carry `indexerHeight`/`finalizedHeight`, which advance about five blocks per 30 s poll, so every poll is a structurally new verdict. A healthy idle wallet therefore emits a new `CoreWallet`, a `StateChange.State` plus `ProgressUpdate`, a new `UnshieldedWalletState` and a new `FacadeState` every 30 s with unchanged balances, and nothing downstream deduplicates. The unit test uses `fixedReads(1_000n, 1_000n)`, so it only proves dedup of byte-identical verdicts.

Suggested fix: dedup on `_tag` (or tag plus a lag bucket) with `Stream.changesWith`, and test with advancing heights.

packages/node-client/src/effect/PolkadotNodeClient.ts
```diff
@@ -92,73 +162,104 @@ export class PolkadotNodeClient implements NodeClient.Service {
...
-      this.ensureConnection(),
-      Effect.tap(() =>
-        Effect.sync(() => {
-          this.#activeOperations += 1;
-        }),
-      ),
+  /** Deregisters one call, disconnecting the shared socket when it was the last one in flight. */
+  #deregister(): Effect.Effect<void> {
+    return SynchronizedRef.updateEffect(this.#activeCalls, (active) =>
+      active === 1 ? Effect.promise(() => this.api.disconnect()).pipe(Effect.as(0)) : Effect.succeed(active - 1),
```
agronmurtezi: **`#deregister` does not wait for the socket to close.** `api.disconnect()` returns while the socket is still `CLOSING`; `make()` awaits `api.once('disconnected')` for exactly this reason, but this path does not. The next call sees a stale `isConnected: true`, skips `connect()`, and its probe is dropped on the dying socket, then sleeps `reconnectionDelay`, reconnects, and typically fails one more pre-open probe, costing two or more seconds of the liveness read's 10 s budget.

Suggested fix: await the `disconnected` event here as `make()` does, ideally through one shared helper.

.changeset/indexer-liveness-cross-check.md
```diff
@@ -0,0 +1,35 @@
...
+The wallet now verifies "synced" against the chain instead of the indexer's self-report: it polls a node's finalized
+head (every 30 seconds by default) and, once, checks that both endpoints name the same genesis block. The node is
+`nodeClientConnection` on the configuration, falling back to `relayURL`; a wallet naming neither is not checked. The
+result is an `IndexerLiveness` verdict on `SyncProgress`: `Behind`, `Unknown` and `WrongNetwork` block completion; the
+other four (`InSync`, `Ahead`, `Unavailable`, `Skipped`) do not. Tune with `livenessConfiguration` and
+`livenessPollInterval`. This detects staleness, not withholding.
+
+### Fixes
+
+- `isConnected` on the sync progress clears when the indexer subscription drops; it previously latched `true`.
```
agronmurtezi: **The gate and the `isConnected` fix exist only in the unshielded wallet.** The shared `abstractions` `SyncProgress` used by shielded and dust is untouched, and neither sibling `RunningV1Variant` clears `isConnected`, so `facade.dust.waitForSyncedState()` (used on its own in `facade/src/index.ts`) and any standalone shielded or dust wallet still resolve synced over a stale indexer, and keep `isConnected: true` through an indexer outage. This Fixes entry reads as SDK-wide.

Suggested fix: scope the entry to `wallet-sdk-unshielded-wallet`, and track the shielded/dust gate as a follow-up.

packages/unshielded-wallet/src/v1/test/sync.test.ts
```diff
@@ -0,0 +1,275 @@
...
+
+describe('makeLivenessUpdates', () => {
+  it('should seed the stream from the verdict it is given, so a reconnect does not erase what was known', async () => {
+    // `updates()` is rebuilt on every indexer-stream retry, under `Stream.retry`. Starting each new liveness stream at
+    // `Unknown` would clear a `Behind` verdict on every reconnect and let `waitForSyncedState` resolve over the stale
+    // view the check had already caught.
+    const behind = IndexerLiveness.Behind({ indexerHeight: 900n, finalizedHeight: 1_000n, lag: 100n });
+
+    const first = await Effect.runPromise(
+      Stream.runHead(Option.getOrThrow(makeLivenessUpdates(withNode, behind))).pipe(Effect.scoped),
```
agronmurtezi: **This unit test opens real network connections.** It calls `makeLivenessUpdates(withNode, behind)` without the `makeReads` stub every sibling test injects, so `makeDefaultLivenessReads` opens a WebSocket to `ws://localhost:9944` and an HTTP request to the indexer URL on the first tick. With nothing on 9944 it passes in ~35 ms and logs an `API-WS: disconnected` line; with a dev node listening it took ~10 s, the full `NODE_CONNECTION_TIMEOUT`, and passes only because `testTimeout` is 30 s. `.claude/rules/testing.md` requires `*.test.ts` to be free of network access.

Suggested fix: pass `() => Effect.succeed(fixedReads(...))` as the third argument.

packages/facade/test/livenessWiring.test.ts
```diff
@@ -0,0 +1,85 @@
...
+describe('indexer liveness wiring through the facade', () => {
+  it('should hand the unshielded wallet a configuration the liveness check can read a node from', async () => {
+    // The check was originally reachable only by setting `nodeClientConnection`, which nothing set — so no wallet built
+    // through the facade ever ran it, while the release notes implied otherwise. Nothing caught that, because every other
+    // facade test runs in simulation mode, whose sync service never consults a node at all. This asserts the wiring
+    // itself: whatever the facade passes to the unshielded initializer must yield an endpoint to check against.
+    const seed = crypto.randomBytes(32);
+    const configurations: DefaultConfiguration[] = [];
+
+    await WalletFacade.init({
```
agronmurtezi, line 64: **This unit test builds a real node client.** `WalletFacade.init` without a `submissionService` override eagerly constructs the default submission service, which calls `PolkadotNodeClient.make({ nodeURL: ws://localhost:9944 })` with an infinite `reconnectionTimeout`, leaving a `WsProvider` reconnect loop running every 2.5 s for the rest of the worker. Adding `facade.stop()` would hang, because `close()` awaits a Deferred that never completes when nothing listens.

Suggested fix: `submissionService: () => fake`, as `submission.test.ts` does.

packages/docs-snippets/src/snippets/indexer-liveness.no-net.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  // The loudest verdict: the two endpoints disagree about the genesis block, so one of them points at another chain
+  // and every balance shown is suspect. This pins until the configuration is fixed — it cannot heal on its own.
+  WrongNetwork: ({ indexerGenesisHash, nodeGenesisHash }) =>
+    `The indexer and the node are on different networks (indexer genesis ${indexerGenesisHash}, ` +
+    `node genesis ${nodeGenesisHash}). Check the configured endpoints.`,
+});
+
+// The verdict lives on the wallet's state, so it is read wherever that state is read.
+export const describeUnshieldedLiveness = (state: UnshieldedWalletState): string =>
```
agronmurtezi, line 64: **This snippet breaks the undeployed lane.** The snippet runner auto-discovers it (`readdirSync` + `it.each(snippetFiles.noNet)` + `toMatchSnapshot`), but no entry was added to `__snapshots__/test-snippets.undeployed.test.ts.snap`, and the file only exports functions so its stdout is empty. `e2e-tests-local.yml` runs `yarn turbo test-undeployed` unfiltered with `CI=true`, where vitest refuses to write new snapshots, so that lane fails; `ci.yml` avoids it only via `-t @smoke`. Locally the run writes an empty snapshot, which is a dirty tree and a meaningless assertion.

Suggested fix: add a printing main body and commit the snapshot, or exclude the file from the runner.

packages/node-client/src/effect/NodeClient.ts
```diff
@@ -17,18 +17,67 @@ import { type SerializedTransaction } from '@midnightntwrk/wallet-sdk-abstractio
...
+  /** The height of the finalized block. */
+  readonly height: bigint;
+};
+
 export interface Service {
   sendMidnightTransaction(
     serializedTransaction: SerializedTransaction.SerializedTransaction,
   ): Stream.Stream<SubmissionEvent.SubmissionEvent, NodeClientError.NodeClientError>;
   getGenesis(): Effect.Effect<Genesis, NodeClientError.NodeClientError>;
+  getFinalizedBlock(): Effect.Effect<FinalizedBlock, NodeClientError.NodeClientError>;
```
agronmurtezi, line 63: **New public interface methods lack JSDoc.** These two methods are the change the changeset flags as BREAKING for implementers, yet they carry no documentation while the accessor functions below them do. `.claude/rules/functional-style.md` requires description, `@param`, `@returns` and `@example` on public APIs; an implementer currently has to reverse-engineer finalized-vs-best head, the `0x`-prefixed hash, and the interleaving guarantee from `PolkadotNodeClient`.

packages/abstractions/src/IndexerLiveness.ts
iangregsondev: Agreed and fixed.

packages/unshielded-wallet/src/v1/SyncProgress.ts
iangregsondev: Agreed and fixed!

packages/unshielded-wallet/src/v1/RunningV1Variant.ts
iangregsondev: Agreed and the same issue has been found for shielded/dust (existing on main).

And issue has been crated separately to fix those - https://github.com/midnightntwrk/midnight-wallet/issues/742

iangregsondev: Issue for unshielded is addressed here.

packages/unshielded-wallet/src/v1/RunningV1Variant.ts
iangregsondev: Agreed and fixed.

packages/capabilities/src/liveness/livenessReads.ts
iangregsondev: Agreed and fixed.

packages/node-client/src/effect/PolkadotNodeClient.ts
iangregsondev: Agreed and fixed.

packages/node-client/src/effect/PolkadotNodeClient.ts
iangregsondev: Agreed and fixed.

packages/capabilities/src/liveness/livenessService.ts
iangregsondev: Agreed and fixed.

packages/node-client/src/effect/PolkadotNodeClient.ts
iangregsondev: Agreed and fixed.

.changeset/indexer-liveness-cross-check.md
iangregsondev: Agreed and a new ticket has been created tot track the other work

https://github.com/midnightntwrk/midnight-wallet/issues/743

packages/unshielded-wallet/src/v1/test/sync.test.ts
iangregsondev: Agreed and fixed.

packages/facade/test/livenessWiring.test.ts
iangregsondev: Agreed and fixed.

packages/docs-snippets/src/snippets/indexer-liveness.no-net.ts
iangregsondev: Fixed.

packages/node-client/src/effect/NodeClient.ts
iangregsondev: Added missing JSDOCs

packages/capabilities/src/liveness/livenessService.ts
```diff
@@ -0,0 +1,309 @@
...
+}>;
+const GenesisCheck = Data.taggedEnum<GenesisCheck>();
+
+export class LivenessServiceImpl implements LivenessService {
+  readonly #state: SubscriptionRef.SubscriptionRef<IndexerLiveness.IndexerLiveness>;
+  readonly #reads: LivenessReads;
+  readonly #configuration: LivenessConfiguration;
+  // A plain `Ref` read at the top of a poll and written mid-poll — not a race, because polls run strictly one at a
+  // time: `startPolling` awaits each poll before taking the next tick.
+  readonly #genesisCheck: Ref.Ref<GenesisCheck>;
```
kapke, line 160: I think Deferred is a better fit for a write-once primitive.

iangregsondev: Done — `#genesisCheck` is now a `Deferred<GenesisCheck>`; the `Unverified` variant and the unreachable match branch went with it.

packages/capabilities/src/liveness/livenessService.ts
```diff
@@ -0,0 +1,299 @@
...
+        IndexerLiveness.sameGenesis(indexerGenesisHash, nodeGenesisHash)
+          ? GenesisCheck.SameChain()
+          : GenesisCheck.Mismatch({ verdict: IndexerLiveness.WrongNetwork({ indexerGenesisHash, nodeGenesisHash }) }),
+      ),
+      Effect.tap((check) => Deferred.succeed(this.#genesisCheck, check)),
+    );
+  }
+
+  /** Reads both heights and reduces them to a verdict under the configured tolerances. */
+  #compareHeights(): Effect.Effect<IndexerLiveness.IndexerLiveness, LivenessReadError> {
```
kapke: This can't be just height comparison, but also a block hash comparison.
In short - if the heights are equal (rare), both indexer and node need to report the same block hash, otherwise we have a mismatch.
If they are different, the one that is ahead needs to be able to resolve matching block by hash, otherwise we have a mismatch. 

If these checks are not carried - it becomes trivial for the indexer to pass the checks and still cheat.

packages/unshielded-wallet/src/v1/RunningV1Variant.ts
```diff
@@ -118,10 +122,53 @@ export class RunningV1Variant<TSerialized, TSyncUpdate> implements Variant.Runni
...
+   *
+   *   A service with no feed gets a verdict too. The progress defaults to `Unknown`, which gates completion until a first
+   *   verdict arrives — and for such a service none ever would, so the wallet could never report itself synchronized.
+   *   This is the one place that inspects the service, so the absence is turned into `Skipped` here, once. The write is
+   *   unconditional: a verdict restored with a serialized state came from a check this wallet no longer has.
+   */
+  #startLivenessInBackground(): Effect.Effect<void, never, Scope.Scope> {
+    const livenessUpdates = this.#v1Context.syncService.livenessUpdates;
+
+    return livenessUpdates === undefined
```
kapke: why can it be undefined?

kapke: Generally speaking - I much more prefer an always existent implementation which does nothing, and whole abstraction accomodating that, than a possibility of absence. The absence might hurt in multiple places, whereas always present implementation just changes the wiring.

packages/capabilities/src/liveness/livenessService.ts
iangregsondev: Agreed, going to fix this now.

iangregsondev: Fixed

packages/unshielded-wallet/src/v1/RunningV1Variant.ts
iangregsondev: Agreed. Made it required — livenessUpdates is no longer optional.

PR #668, opened 2026-08-18
refactor(dust): restructure variants — v1 = ledger-v8 (restored), v2 = ledger-v9

agronmurtezi:
Phase 6, dust increment (1 of 2): the shielded v1/v2 restructure recipe (#652) applied to the dust wallet. Stacked on #656. Behavior-preserving: production `DustWallet` still registers only the v9 variant; all existing tests pass with only mechanical renames.

## What changed

- **Rename**: `src/v1/` (ledger-v9 code, misnamed) → `src/v2/` with `V1*`→`V2*` identifier renames; blanket rename verified against a full token census first (141 `V1`/`v1` tokens, all dust's own; post-rename census matched exactly). Unlike shielded, dust keeps tests outside `src/`, so the rename scope includes `packages/dust-wallet/test/`.
- **Restore**: `main`'s ledger-v8 dust variant as `src/v1/` (honest `V1*` names, `@midnight-ntwrk/ledger-v8` dependency added). Serialization is rename-safe (snapshots hold publicKey/protocolVersion/networkId/offset/state blob — no variant naming).
- **Consumers repointed** (10 files, scoped edits — unshielded's genuine `V1` names untouched): testkit, facade test helpers, e2e, and the published `dust-fast-sync` docs snippet. `wallet-sdk` exposes **both** public subpaths, `dust/v1` = v8 and new `dust/v2` = v9, mirroring the shielded decision.

## Where dust genuinely differs from the shielded template (documented, not hidden)

- **The v1/v2 twins are NOT byte-parallel and cannot be**: ledger-v8's `DustLocalState` lacks `updateGenerationTreeFromEvidence`, `commitmentTreeFirstFree`, `generatingTreeFirstFree` and `nullifiers` — all used by the v9-only projections fast-sync work (#617). `src/v1` is therefore `main`'s tree verbatim (event-replay sync only, no `SyncSchema.ts`).
- Only one of shielded's three restore fixes applies (`activationRange` stub). The V8-simulator requalification was needed but **TypeScript was silent about it** — the simulator twins are structurally similar enough that v9 `Event` objects flowed into v8 WASM without a type error; caught by tests. No proving stub needed (dust never imports capabilities' proving).

## Verification

[...]

PR #669, opened 2026-08-18
feat(dust): fork boundary wiring, cross-version migration, and the crossing proof

agronmurtezi:
Phase 6, dust increment (2 of 2): the fork wiring + migration + proof pattern (#655 + #656) applied to the dust wallet. Stacked on #668. TDD: the red proof commit precedes every implementation commit; **20 red → 143 passing** (121 → 143 tests).

## What this delivers

- **Boundary wiring (D2/D3), both variants**: pure split/annotate helpers in dust's `Sync.ts`; monotone version recording on dust's core state; `activationRange` stored and threaded; healing emission. **Dust-specific D3 rule**: dust's sync schema gains an optional `protocolVersion` — an item WITHOUT a version applies normally and leaves the state version untouched (the deployed indexer doesn't serve the field yet; the selection-set + codegen stays deferred per the plan of record). Capability matrix covers untagged items and mixed batches.
- **Migration seam + cross-version instance**: fresh dust state, keys/network carried as plain data, **cursor parked at the fork** (the confirmed ids-continue semantics). Replaces the old `TODO` `migrateState` that returned `prevState`. Dust-specific: `DustParameters` are version-specific WASM objects that cannot cross — both migrations take them as builder configuration (optional, defaulting to initial parameters; no call site breaks). `withMigration` takes context per-method rather than as the builder's `TContext` (shielded's form makes the builder invariant in `TContext` and breaks dust's partially-configured chains — documented in place).
- **Restart wiring (D1)**: `DustWallet` retains the start aux in a private `Option`, registers the activation watcher once, re-dispatches `startSyncInBackground(retainedAux)` on the new variant; `stop()` clears. `DustSecretKey.fromSeed` parity across v8/v9 verified (seed retention works for dust too).
- **The crossing proof** (`test/…` unit tier): a real dust chain (4 Night rewards, then one registration per Night UTXO — one `dustInitialUtxo` event per block, which is what lets one fixture drive both the batched and block-granular capabilities) → fork → `VersionChange` → migration captured with parked cursor → fresh post-fork state → **dust re-discovered from the replayed timeline and compared UTXO-for-UTXO** (not just by count) → both negatives (within-range bump; fresh post-fork wallet). Mutation-verified: park→reset fails 2 tests (`expected 0n to be 4n`); corrupting a replayed event's payload fails 3.

## Honest limits (in docstrings and the changeset, not hidden)

- **No closing spend**: dust is spent as the fee half of a transaction a *chain* must validate; an indexer replay is an event log, not a ledger. Commitment-root equality stands in its place.
- **No WASM-translation fidelity link — deliberately unit-tier only**: ledger-v8 and v9 encode a `dustInitialUtxo` identically apart from the header tag, so the replay uses the *same bytes* where shielded re-mints *equivalent* coins. The proof opens by asserting that byte-equivalence, so a future ledger encoding change retires the model loudly. `turbo.json`/CI integration gates are correctly untouched.
- **Finding — the projections fast-sync path cannot do boundary handling**: a projections update is a folded snapshot with no per-item version, and `BlockData` carries none. It accepts `activeRange` for signature parity and ignores it — a wallet syncing exclusively via projections will not hand over at a fork. Pinned by a dedicated test and called out in the changeset; resolving it needs indexer-side versioning on projections (same ask family as the deferred selection-set work).

## Verification

Branch tip: `yarn dist --force` 17/17 · `typecheck --force` 37/37 · `lint` 58/58 · `test:unit --force` 53/53 · dust suite 143 passed · cross-package typecheck · Effect diagnostics clean · `format:check` · changeset status clean. All commits GPG-signed.

### DIFF
--- a/.changeset/dust-fork-boundary-wiring.md
+++ b/.changeset/dust-fork-boundary-wiring.md
@@ -1,0 +1,51 @@
+---
+'@midnightntwrk/wallet-sdk-dust-wallet': minor
+---
+
+Make the dust wallet able to follow the chain across a hard fork: sync now recognises the protocol version the indexer
+reports, hands over at the variant boundary, and starts the next ledger version's variant on a fresh state.
+
+**Behavioural change: wallet state now tracks the reported protocol version.** Until now nothing wrote the indexer's
+`protocolVersion` into the wallet state, so it stayed pinned at its initial value and a protocol version change could
+never be signalled. It is now recorded on every applied batch that reports one. The value only ever increases — a source
+that briefly reports an older version (a reconnect replaying from an earlier cursor, say) cannot drag the wallet back
+below a boundary it has already crossed.
+
+**Breaking for custom `withSync` authors: `SyncCapability.applyUpdate` takes a third argument.** It is now
+`applyUpdate(state, update, activeRange)`, where `activeRange` is the half-open protocol version range the running
+variant owns. Anything the source reports at or beyond its end belongs to a later variant and must be left unapplied for
+that variant to fetch. Implementations that ignore the parameter keep their present behaviour; the exported
+`splitAtVersionBoundary` and `annotateVersion` helpers implement the rule for anyone who wants it.
+
+**`SyncEventsUpdateSchema` gains an optional `protocolVersion`.** Optional because dust's subscription does not select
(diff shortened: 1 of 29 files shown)


agronmurtezi:

PR #671, opened 2026-08-19
refactor(unshielded): restructure variants — v1 = ledger-v8 (restored + signer retrofit), v2 = ledger-v9

agronmurtezi:
Phase 6, unshielded increment (1 of 2): the v1/v2 restructure applied to the unshielded wallet, **including the v8 async-signer retrofit** — the known expensive part of the whole Phase-6 line. Stacked on #669. Production `UnshieldedWallet` still registers only the v9 variant; no assertion weakened anywhere.

## What changed

- **Rename**: `src/v1/` (ledger-v9 code) → `src/v2/` with `V1*`→`V2*` renames, census-verified across BOTH test trees (unshielded keeps tests in `src/v1/test/` *and* top-level `test/`).
- **Restore**: `main`'s ledger-v8 tree as `src/v1/` + `@midnight-ntwrk/ledger-v8` dependency.
- **The signer retrofit**: `main`'s v8 signing was two synchronous `Either`-returning methods on `TransactingCapability` with an internal segment loop. The current architecture requires the async `SigningService`/`SignSegment` split. The retrofit gives v1 its own `Signing.ts` (v8 twin, minus v9-only scheme validation), the pure halves (`collectSignableData`/`attachSignatures`) in `TransactionOps`, `withSigning`/`withSigningDefaults` on `V1Builder`, and routing through `RunningV1Variant`. ECDSA/`SchemeConsistency` stay out of v1 (v9-only); the bridge-claim `SyncSchema` lines back-ported. Two discoveries beyond the plan: top-level `src/KeyStore.ts` is v9-typed (v9 signatures are `{tag, value}`, v8's are bare hex), so v1 got its own `KeyStore.ts` (main's verbatim + `signDataAsync`); and the `Simulator` re-export needed `V8` requalification.
- **Restored-test adaptations — exactly one file, two cases**: the two "when signing" tests in `v1/test/transacting.test.ts` now drive `SigningService.sign` and unwrap an `Effect` instead of an `Either`; both assertions byte-identical.
- **8 external `/v1` consumers repointed** (wallet-integration-tests alone has four); `wallet-sdk` exposes both `unshielded/v1` (v8) and new `unshielded/v2` (v9) — the standing per-wallet decision.

## Verification

Branch tip standalone: `yarn dist --force` 17/17 · `typecheck --force` 37/37 · `lint` 58/58 · 138/138 package tests (93 before, +45 restored) · cross-package typecheck · Effect diagnostics clean apart from the known-accepted `deterministicKeys` on `WalletError.ts` · `format:check` · changeset status clean. All commits GPG-signed.


agronmurtezi:
**Restore-discipline correction (2026-08-19, user-directed):** an inventory pass against the v2 line found three portable items the retrofit had missed, now ported into v1: the deserialization **key↔address consistency guard** (`assertKeyAddressConsistency` — a real gap: a spliced snapshot previously loaded silently and a malformed key escaped as a wasm trap), `extractOwnInputs` returning `Utxo`s owned by the derived address rather than the verifying key (inert today, wrong-valued), and a signer-rejection→`SignError` coverage test. ECDSA/`SchemeConsistency`/tagged-key schemas remain justified v9-only exclusions (verified: ledger-v8 has no `SignatureKind`; keys and signatures are bare strings).

agronmurtezi:

PR #672, opened 2026-08-19
feat(unshielded): per-message fork boundary, structural migration, and the crossing proof

agronmurtezi:
Phase 6, unshielded increment (2 of 2) — and the last per-wallet increment of the hard-fork line. Stacked on #671. TDD: 16 red tests (+2 unloadable files) before implementation; 185/185 passing at the tip.

## What this delivers

- **Per-message boundary semantics (D2, unshielded rules)** — unlike shielded/dust's batched split, unshielded sync is per-message: a transaction at/after the boundary is **not applied at all** (no UTXO change, no `appliedId` bump, no tx-history put) — only the version annotation is written, and the next variant re-fetches it from the unchanged cursor. Progress messages never touch the version. Note: unshielded's `protocolVersion` is **required** on the transaction schema (unlike dust's optional), so there is no "absent means in-range" case.
- **Structural cross-version migration** — unshielded state is public UTXO data, so `migrateState` carries it field-for-field into the new variant (no fresh-state, no replay wait), with the cursor parked. The builder gains the migration seam (`withMigration` in dust's method-level-context form).
- **Keyless restart (D1)** — unshielded is watch-only; the activation watcher (previously absent entirely) re-dispatches `startSyncInBackground()` with no retained aux.
- **The crossing proof**: the load-bearing chain of assertions is that the boundary transaction is applied **exactly once, by the new variant** — each link asserted separately — plus field-for-field UTXO carry, the parked cursor asserted on the captured migration before sync touches it, both negatives, and a lemma that both ledger versions derive the same address from the same secret. Mutation-verified four ways: drop the carry (3 tests fail), advance the cursor by one (the boundary tx is lost), apply the boundary tx in the old variant (cursor moves), apply it there with the cursor parked (it arrives twice).

## A real bug the proof caught (and the unit tests could not)

`PreviousLedgerWallet.state` typed its UTXOs as `Iterable<UtxoLike>` — but the real state holds Effect `HashMap`s, whose iteration yields `[key, value]` **entries**, so the migration was carrying malformed pairs. Nothing caught it: the runtime casts into `migrateState` (TypeScript never checked) and the migration unit tests built fixtures from plain arrays that satisfy the type honestly. The fork proof failed end-to-end, the type now names `HashMap` (still version-agnostic — it's an Effect type), fixtures build through the real `UnshieldedState`, and the stronger type immediately caught the same mistake in the test factory. This is the unshielded instance of the silent-structural-compatibility hazard the dust increment flagged.

## No analogue, stated plainly

There is no re-framing/re-minting of ledger bytes and no integration companion: unshielded state is public and the wire format is JSON, so there is no ledger encoding for a translation to be faithful to. `turbo.json` and the CI integration workflow are untouched. Both variants are nonetheless genuinely ledger-v8 and ledger-v9 with their own modules.

## Verification

Branch tip (rebased onto #669's CI-timeout fix before push; dust 21/21 + unshielded 19/19 + typecheck 37/37 re-verified after the rebase): `yarn dist --force` 17/17 · `typecheck --force` 37/37 · `lint` 58/58 · `test:unit --force` 53/53 · 185/185 package tests · cross-package typecheck 19/19 · Effect diagnostics clean apart from the known-accepted `deterministicKeys` · `format:check` · changeset status clean. All commits GPG-signed.

### DIFF
--- a/.changeset/shielded-v1-v2-restructure.md
+++ b/.changeset/shielded-v1-v2-restructure.md
@@ -13,11 +13,12 @@ Restructure the shielded wallet for the coming hard fork: the variant directorie
 `RunningV1Variant`→`RunningV2Variant`, and so on). Imports of the old names from `./v1` will no longer resolve — switch
 the subpath to `./v2`.
 
-`./v1` now holds the restored pre-fork ledger-v8 variant, byte-for-byte the `3.x` line's shielded internals with its
-honest `V1` names, kept for wallets that must sync pre-fork history across the fork. This makes
-`@midnight-ntwrk/ledger-v8` a runtime dependency of the package: consumers of the `./v1` subpath load a second ledger
-WASM module, which matters for browser bundle size. The `Simulator` namespace exported from `./v1` is the ledger-v8
-simulator twin only, where it previously re-exported the whole simulation entry point.
+`./v1` now holds the pre-fork ledger-v8 variant with its honest `V1` names, kept for wallets that must sync pre-fork
+history across the fork: the `3.x` line's shielded internals restored, and then given this line's fork wiring, so both
+variants carry the same boundary splitting, migration seam and healing emission (see the boundary-wiring change). This
+makes `@midnight-ntwrk/ledger-v8` a runtime dependency of the package: consumers of the `./v1` subpath load a second
+ledger WASM module, which matters for browser bundle size. The `Simulator` namespace exported from `./v1` is the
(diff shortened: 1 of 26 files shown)


agronmurtezi:
Consolidated for review: replaced by #690 — same head branch, identical commits, based on `feat/fork-state-translation`, carrying the whole three-wallets slice (#652→#672) as one net diff. This PR is base-locked by GitHub's PR-stack feature, hence a replacement rather than a retarget.

PR #675, opened 2026-08-19
feat(abstractions,runtime): version-dispatch primitives for the dual-ledger API

agronmurtezi:
First increment of the dual-ledger public API redesign (plan of record: workspace-local `api-redesign.md`; work packages WP-1–4). Stacked on #672. No public-surface flows change yet — this PR builds the version-dispatch primitives every later increment (restore routing, versioned codecs/proving/pending, two-variant registration) selects on.

## What's here

- **WP-1 — protocol-version-keyed registry** (`abstractions`): `ProtocolVersion.Registry<T>` with `makeRegistry` / `makeRegistryFromActivations` (overlap-rejecting, `Either`), and total `select` / `selectEntry` returning `Option`. Half-open ranges, same semantics as the existing `withinRange`.
- **WP-2 — variant resolution** (`runtime`): `Variant.selectByRange(variants, version)` and `BaseWalletClass.variantFor(version)` resolve which registered variant owns a `ProtocolVersion`, deriving boundaries exactly the way the runtime derives `VariantContext.activationRange` (implemented on the WP-1 registry). Typed miss (`Option`), never throws.
- **WP-3 — `ProtocolState.variantTag`**: the runtime now hands over the tag of the variant that produced each state emission (`RuntimeState` types it as the union of registered tags), so downstream per-emission capability selection needs no casts. `ProtocolState.getEquivalence` compares the tag strictly.
- **WP-4 — configuration namespacing**: `withVariant(sinceVersion, builder, configuration)` overload — a self-configured variant drops out of the builder's configuration intersection (all-self-configured ⇒ `build()` takes no configuration). On top of that, each wallet package now **owns** its `DefaultShieldedConfiguration` / `DefaultDustConfiguration` / `DefaultUnshieldedConfiguration` instead of aliasing the v2 variant's configuration type.

## Breaking notes (loud in the changesets)

- `ProtocolState` gains a **required** `variantTag` — anything constructing one (including tests) must supply it. Runtime consumers only reading states get the extra field additively.
- The wallet packages' `Default*Configuration` are no longer type-aliases of `DefaultV2Configuration` (same declared shape today; the alias identity is the break). `DefaultV1/V2Configuration` remain exported from `./v1` / `./v2` unchanged.

## Notable finding

The plan's claim F3 ("production variant configurations are structurally identical **and ledger-free**") is half-stale: dust's variant configs carry `dustParameters?: DustParameters` from their respective ledgers. One package-owned config works only because v8 and v9 `DustParameters` are structurally identical — that coincidence is now **pinned by `dust-wallet/src/test/configuration.test.ts`**, so a future divergence surfaces as a compile error instead of a wallet that cannot be built for one variant.

## Tests & gate

Red-first TDD throughout (each `feat`/`refactor` commit is preceded by its failing-spec commit): 18 registry tests, 5 `ProtocolState`, 7 `selectByRange`, 14 builder (`variantFor` / `variantTag` / self-configuration, incl. type-level), 3× package config guards. Gate: dist 17/17, typecheck 37/37, lint 58/58, unit 53/53 tasks (899 passed / 22 skipped), effect-language-service clean on all changed files (the two `Runtime.ts` hits pre-exist on the base branch), `format:check` + `changeset:check` green.

### DIFF
--- a/.changeset/package-owned-wallet-configuration.md
+++ b/.changeset/package-owned-wallet-configuration.md
@@ -1,0 +1,32 @@
+---
+'@midnightntwrk/wallet-sdk-shielded': major
+'@midnightntwrk/wallet-sdk-dust-wallet': major
+'@midnightntwrk/wallet-sdk-unshielded-wallet': major
+---
+
+**`DefaultShieldedConfiguration`, `DefaultDustConfiguration` and `DefaultUnshieldedConfiguration` are no longer aliases
+of their head variant's builder configuration.** Each package now declares its own configuration type. The declared
+shape is unchanged, so applications that pass a configuration object literal — which is all of them — need to change
+nothing.
+
+What breaks is the alias identity. Code that relied on these being the *same type* as `DefaultV2Configuration` no
+longer holds; in particular a package-owned configuration will not silently acquire a field that the v2 variant's
+builder configuration gains. If you were passing a package configuration straight into a variant builder, or vice
+versa, that still compiles today only because the shapes coincide, and the packages now assert that coincidence rather
+than assume it: each package has a type test requiring its configuration to stay interchangeable with what **both** its
+variants are built from, so a divergence surfaces as a compile error in the SDK instead of as a wallet that cannot be
+built for one of its variants. `DefaultV1Configuration` and `DefaultV2Configuration` are unchanged and still exported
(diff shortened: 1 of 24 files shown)


agronmurtezi:

PR #676, opened 2026-08-19
feat(runtime,shielded,facade): wallet-layer start material and version-routed restore

agronmurtezi:
Second increment of the dual-ledger public API redesign (WP-5a/b, WP-6, WP-7 runtime half + facade). Stacked on #675. This PR builds the wallet-layer machinery for starting and restoring across variants; the production two-variant registration flip is deliberately NOT here (next PR — see below).

## What's here

- **WP-5a — `StartMaterial`** (`runtime`): what a wallet retains so *any* variant can start sync — `FromSeed { seed }` or `ForVariants { byTag }` (the latter is the generalized both-keys product from the key-decision: a partial set is a typed miss, never wrong-ledger keys). `auxFor(material, tag, deriveFromSeed)` resolves per activation.
- **WP-5b — `startAux` variant capability** (`shielded`): both variants gain `startAux.fromSeed(seed)` deriving their **own** ledger's `ZswapSecretKeys` (plan axiom: the seed is the only key material that crosses the fork). Required on the builder — `withStartAuxDefaults()` in `withDefaults()`, `withStartAux(cap)` for custom chains, resolved at build time with a hard error rather than silent wrong-ledger keys. `withSync` drops a stale derivation, mirroring `withMigration`.
- **WP-6 — restore routing, complete**: `Restore.peekProtocolVersion` (permissive envelope peek) → `variantFor(version)` → new `BaseWalletClass.startAtVariant(class, variant, state)`. A snapshot whose *readable* version no registered variant owns fails with typed `UnsupportedSnapshotVersionError`; a legacy envelope (version absent, unreadable, or payload not a JSON object) falls back to the head variant, leaving real diagnosis to the deserializer. **Serialized format unchanged** — `protocolVersion` was already a required snapshot field.
- **WP-7 (runtime half)**: `RuntimeState` is now a *distributed* union, so `variantTag` is a genuine discriminant — narrowing on it narrows the state type with zero casts (pinned by test). Facade gains additive `FacadeState.protocolVersion` (per wallet) and `activeProtocolVersion` (min of the three).

## Design note: variant-addressed, not tag-addressed, start

`HList.Find` cannot recover a state type from a runtime-derived tag (it collapses to `never`), so instead of weakening the tag-addressed `start`, restore routing hands the **resolved variant** to `startAtVariant` — `StateOf<TVariant>` distributes over the union, callable from a `variantFor` result with no cast at the call site (one justified internal cast at the `Runtime.init` boundary, documented).

## Deliberately deferred (atomic trio, next PR)

The wallet class still retains start-aux directly (WP-5c), state wrappers are still v2-pinned (WP-7b), and `ShieldedWallet(config)` still registers one variant (WP-8). Those three are only *observable* together with two variants registered, and WP-8 drags a ~20-site required-`forkVersion` sweep across 8 packages — landing them as one dedicated PR keeps both PRs honest and reviewable. `src/test/forkWallet.ts` and its remaining workarounds stand until then; the fork-simulation regression tests pass here **unmodified**.

## Plan claim F2 — confirmed, with a caveat

Every scalar the shielded state wrapper projects is a `string` alias in both ledgers, and the coin shapes are byte-identical declarations (`diff` of the two `.d.ts` ranges); the only version-bound member is `CoreWallet.state` + `serialize()` input — exactly as claimed. Caveat carried into WP-7b: identical shape ≠ identical declared type, so capability and state must be selected together inside a `variantTag`-narrowed branch; pairing a V2 capability with a union state would type-check via method bivariance while being wrong.

## Tests & gate

Red-first throughout: `startMaterial` (6), `startAux` v2+v1 mirrors (3+3), `restore` (9), facade `protocolVersion` (4), +4 builder tests. Gate: dist 17/17, typecheck 37/37, lint 58/58, unit 53/53 tasks (928 passed / 22 skipped), effect-language-service clean on changed files (the two `Runtime.ts` hits pre-exist on base), `format:check` + `changeset:check` green. Changesets: runtime minor, shielded **major** (hand-composed builder chains must add `withStartAux`/`withStartAuxDefaults`), facade minor.

### DIFF
--- a/.changeset/facade-protocol-version.md
+++ b/.changeset/facade-protocol-version.md
@@ -1,0 +1,15 @@
+---
+'@midnightntwrk/wallet-sdk-facade': minor
+---
+
+`FacadeState` gains two additive readings of where the wallets are in the protocol timeline.
+
+`protocolVersion` reports the version each of the three wallets has reached. `activeProtocolVersion` is the lowest of
+them: the three follow the same chain but not in lock-step — each recognises a protocol version change when its own
+synchronization reaches it — so around a fork they disagree for a while, and a transaction needs all three. The lowest
+is the version every wallet is known to understand.
+
+`WalletProtocolVersions` and `lowestProtocolVersion` are exported for callers that want the same rule over versions they
+hold themselves.
+
+Purely additive: nothing consumers read today changes shape.
--- a/.changeset/shielded-start-aux-and-restore-routing.md
+++ b/.changeset/shielded-start-aux-and-restore-routing.md
@@ -1,0 +1,33 @@
+---
+'@midnightntwrk/wallet-sdk-shielded': major
+---
+
+**Both shielded variants now derive their own key material from a seed, and restoring a snapshot routes on the protocol
+version it declares.**
+
+**Breaking for anyone composing a `V1Builder`/`V2Builder` chain by hand: `withStartAux` is required.** Each variant
+carries a `startAux` capability — given the seed the application started the wallet from, it produces the key material
+its own ledger version's sync expects. This is what makes a seed the only key material a wallet needs to hold in order
+to cross a protocol boundary: key objects belong to one ledger's runtime and cannot be re-handed to the variant on the
+other side, even though the public keys they yield are identical.
+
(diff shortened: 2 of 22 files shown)


agronmurtezi:

PR #677, opened 2026-08-19
feat(shielded): retain start material and require the chain's fork version

agronmurtezi:
Third increment of the dual-ledger public API redesign (WP-5c + the two-variant config contract and its full sweep). Stacked on #676. The behavioural registration flip is deliberately the NEXT PR — this one lands everything the flip needs, so that PR can be almost purely behavioural.

## What's here

- **WP-5c — the wallet class adopts `StartMaterial`**: `ShieldedWallet` retains `StartMaterial` instead of raw key objects. `startWithSeed` retains the seed (usable by ANY variant it later migrates to); `start(keys)` accumulates key objects per variant tag — the incrementally-built both-keys product from the key decision — and a retained seed supersedes them. A key-object-started wallet crossing into a variant it holds nothing usable for fails with a typed `MissingStartAuxError` carrying that variant's tag — never silent wrong-ledger keys. Behaviour-preserving for a single-variant wallet: the four existing retention/activation tests pass unmodified.
- **Required `forkVersion`** on `DefaultShieldedConfiguration` (no default — the mainnet fork constant is an open external ask), plus the full compiler-driven sweep: 26 construction sites across facade (whose `DefaultConfiguration` intersection inherits the requirement), testkit, e2e-tests, wallet-integration-tests, docs-snippets, test-website. Central factories (`testkit/environment.ts`, `e2e-tests/test-fixture.ts`) define it once; docs-snippets carry the inline literal with a comment so they stay copy-pasteable standalone. `DefaultV1/V2Configuration` deliberately do NOT gain the field — the fork version belongs to the wallet layer; neither variant knows another exists (pinned by type-level tests).

## Why 2000000 (`V9_NATIVE_FORK_VERSION`)

Nothing in the repo pinned what a real chain reports as `protocolVersion`, and the flip's correctness depends on it (head variant = v8; migration to v9 happens only when sync observes a version ≥ `forkVersion`). Measured empirically against the exact standalone environment the live suites use (midnight-node 2.0.0-rc.3 + indexer-standalone, `infra/compose/docker-compose-dynamic.yml`), two independent ways: the `dustCommitmentMerkleTreeUpdate` HTTP query and a live `zswapLedgerEvents` subscription event both report **`protocolVersion: 2000000`** — the scaled node runtime version (2.0.0 → 2_000_000), not a small ordinal. `forkVersion: 1` would also work on v9-native chains but is unsafe as a copy-paste source: on a genuinely pre-fork chain (1.x node, version ≥ 1) it would route ledger-v8 bytes into the v9 variant. 2000000 is correct in both worlds, so every swept site uses the shared named constant.

## Changesets

shielded **major** (required `forkVersion` — every constructor site; what the value means; that the mainnet constant is TBD), facade **major** (the intersection makes every facade config literal break), testkit **major** (exported `WalletConfiguration` requires the field; suites using `getWalletConfig()` are unaffected). Also in passing: `shielded-wallet/README.md` refreshed (was stale independently — ledger-v7 refs, missing `txHistoryStorage`, v3 indexer paths).

## Tests & gate

WP-5c red-first (typed-miss specs); the config threading is honestly type-driven, not test-driven — registration is still single-variant, so there is no behaviour to specify; `yarn typecheck --force` drove the sweep to completion. Gate: dist 17/17, typecheck 37/37, lint 58/58, unit 53/53 tasks (930 passed / 22 skipped), shielded fork-simulation integration 3/3 (no --force, cached WASM), effect-language-service 0 errors on all 26 changed files, format + changeset checks green.

### DIFF
--- a/.changeset/shielded-required-fork-version.md
+++ b/.changeset/shielded-required-fork-version.md
@@ -1,0 +1,49 @@
+---
+'@midnightntwrk/wallet-sdk-shielded': major
+'@midnightntwrk/wallet-sdk-facade': major
+'@midnightntwrk/wallet-sdk-testkit': major
+---
+
+**`DefaultShieldedConfiguration` gains a required `forkVersion`. Every site that builds a shielded — or facade —
+configuration must add it.**
+
+`forkVersion` is the protocol version at which a chain hands over from the pre-fork ledger version to the post-fork one.
+The shielded wallet is being made able to register a variant either side of that boundary, and the boundary itself is a
+property of the chain an application points at, not of the SDK — so the SDK cannot supply it. It is required rather than
+defaulted on purpose: a wrong value does not degrade gracefully, it decides which ledger version reads the chain.
+
+```ts
+import { ShieldedWallet, V9_NATIVE_FORK_VERSION } from '@midnightntwrk/wallet-sdk-shielded';
+
+const wallet = ShieldedWallet({
+  networkId,
+  indexerClientConnection,
+  txHistoryStorage,
+  forkVersion: V9_NATIVE_FORK_VERSION,
+});
+```
+
+**`V9_NATIVE_FORK_VERSION` (`2000000`) is the value for a ledger-v9-native chain**, and it is measured rather than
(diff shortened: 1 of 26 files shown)


agronmurtezi:

PR #678, opened 2026-08-19
feat(shielded)!: a genuinely two-variant production wallet

agronmurtezi:
Fourth increment of the dual-ledger public API redesign — the shielded lane's payoff: **`ShieldedWallet(config)` now registers a variant on either side of the fork**, and the fork-crossing proofs run through the shipped wallet instead of a test-only factory. Stacked on #677.

## What's here

- **`CustomForkingShieldedWallet(configuration, preFork, postFork)`** + `ForkingShieldedWalletClass`; `ShieldedWallet(config)` is now the forking wallet (v1 since `MinSupportedVersion`, v2 since `config.forkVersion`), each variant registered with its **own** configuration via the self-configured `withVariant`. `CustomShieldedWallet`/`CustomizedShieldedWallet*` are untouched and stay single-variant. The wallet-level parameter is the narrower `ForkingShieldedConfiguration` (`{networkId, forkVersion}`) — `DefaultShieldedConfiguration` satisfies it, and custom compositions no longer invent an indexer URL they don't use.
- **Unplanned but load-bearing — cross-ledger event decoding** (own commits, own changeset): measured that each ledger's `Event.deserialize` rejects the other's bytes (`expected header tag 'midnight:event[v9]:', got 'midnight:event[v14]'`), while the sync service decoded eagerly on arrival. Without a fix the flip is a paper flip: a pre-fork wallet on a v9-native chain fails its first batch and retries forever instead of observing the version and handing over. `EventsSyncUpdate` now carries raw hex and each variant decodes only the slice it actually applies — in both twins, pinned with specs feeding each variant the *other* ledger's real event bytes. Breaking for custom sync capability authors.
- **`forkWallet.ts` is deleted**, its three workarounds genuinely dissolved: config collision → per-variant self-configuration; start-aux across the boundary → retained `StartMaterial` (`startWithSeed` + `requireDerivedAuxFor`); the deferred post-fork source → an ordinary field of the post-fork variant's own configuration. The fork-simulation tests (unit + integration) were rewired with exactly two wiring edits each — import path and calling `wallet.shielded.transferTransaction(...)` through the shipped API — **zero assertion changes** (diff-verified).
- **The temporary pre-fork transacting seam**: `balanceTransaction`/`transferTransaction`/`initSwap` on the v1 branch fail with typed `PreForkTransactingUnsupportedError {operation}`, pinned against a synced coin-holding pre-fork wallet. JSDoc + changeset state loudly: TEMPORARY, closed by version-routed proving (WP-11 + Flow 2/3), must not survive to GA — mainnet is pre-fork until the fork. One deliberate exception: `revertTransaction` is a pre-fork **no-op** (pinned), because the facade reverts all three wallets together on submission failure, reverting builds nothing, and while the seam holds the pre-fork variant cannot have produced the transaction being reverted — the alternative was weakening an existing facade spec.
- **Start behavior pinned** (`forkStart.test.ts`, ≥/<-`forkVersion` semantics, no chain-version literals): a chain already past `forkVersion` → v1 boots, applies **nothing**, hands over (keys crossed, fresh state), v2 re-discovers the coins — the accepted spurious start migration. A chain below `forkVersion` → syncs on v1 with seed-derived v8 keys and stays there.

## Notes for reviewers

- **Cross-release restore**: a snapshot declaring a version below `forkVersion` now routes to the pre-fork variant; a *never-synced* snapshot from an earlier beta (declares 0) will fail to deserialize there. Within-release snapshots round-trip. Documented in its own changeset.
- **Empty-timeline chains**: a wallet on a chain that has produced no shielded events never observes a protocol version, so it stays pre-fork and refuses transaction building. This promotes WP-8b (probe chain version at start) from optimization to a likely GA requirement — tracked in the plan.
- The e2e/undeployed validation of the v9-native production path is **this PR's CI** (smoke e2e + the integration matrix run the swept suites fresh).
- Dust and unshielded have the same eager-decode shape in their sync paths — to check before their own flips.

## Tests & gate

Red-first verified with the implementation stashed (fork tests fail on the missing module; decode tests fail on the missing `raw` field). Gate: dist 17/17, typecheck 37/37, lint 58/58, unit 53/53 tasks (shielded **180 passed / 22 skipped** across 25 files), fork-simulation integration 3/3 (no --force), effect-language-service 0 findings on all 14 changed files, format + changesets green. Changesets: shielded majors (two-variant flip + decode change + the seam wording), plus the named cross-release restore case.

### DIFF
--- a/.changeset/shielded-bound-state-projections.md
+++ b/.changeset/shielded-bound-state-projections.md
@@ -1,0 +1,26 @@
+---
+'@midnightntwrk/wallet-sdk-runtime': minor
+'@midnightntwrk/wallet-sdk-shielded': major
+---
+
+Two changes that let one wallet read state produced by either of two variants.
+
+**`StartMaterial.requireDerivedAuxFor` / `StartMaterial.seedOf` (runtime).** A wallet's public API speaks one ledger
+version's key objects, but the variants either side of a protocol boundary do not agree on that type.
+`requireDerivedAuxFor` serves the odd one out from a retained seed, failing with `MissingStartAuxError` when the wallet
+holds key objects instead — there is nothing to convert, because key objects belong to one ledger version's runtime.
+`seedOf` exposes the seed arm directly.
+
+**Breaking (shielded): `ShieldedWalletState` no longer exposes `capabilities` or `services`.** It now holds projections
+already bound to the variant that produced the emission, rather than a capability set and a state side by side. The two
+variants' capability types are structurally identical, so a capability belonging to one type-checks against a state
+belonging to the other and would be silently wrong at runtime; binding them at the point where the producing variant is
+known takes the pairing out of the type system's hands. `ShieldedWalletState.mapState` is replaced by
+`ShieldedWalletState.fromVariant(variant, emission)`, which is generic over the state type so that the pairing is
+checked.
+
+**Everything observers read is unchanged** — `balances`, `totalCoins`, `availableCoins`, `pendingCoins`,
+`coinPublicKey`, `encryptionPublicKey`, `address`, `progress`, `protocolVersion` and `serialize()` keep their names,
+types and meaning, and their tests pass unmodified. The one widening is `state`, now the union of the two variants' core
+states (`ShieldedCoreState`) — the single member the ledger version surfaces on, everything else being version-agnostic
+plain data.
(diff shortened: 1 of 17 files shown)


agronmurtezi:

PR #679, opened 2026-08-20
feat(capabilities,facade,dust)!: version-routed codecs, pending transactions, and proving

agronmurtezi:
Fifth increment of the dual-ledger public API redesign — the services lane (WP-9/10/11). Stacked on #678. Everything here selects through the one `ProtocolVersion.Registry` primitive from #675; no second dispatch mechanism exists.

## What's here

- **WP-9 — version-switched codecs**: new `LedgerParametersCodec` namespace (`capabilities/codecs`); every site that decoded indexer hex with a hardcoded ledger is either routed or accounted for (full survey table in the increment report): capabilities' validation `blockData`, dust v1 sync, dust v2 `BlockDataSchema` (wire schema now carries raw hex + `protocolVersion`, decoded at its own version). The indexer-client block selections now fetch `Block.protocolVersion`.
- **WP-10 — versioned pending + orphan-on-fork**: `VersionedTransactionTrait` = a registry of `TransactionTrait`s; each pending item is stamped with the `protocolVersion` it was authored for (additive envelope — old envelopes deserialize with the head trait and are never orphaned). `TransactionResult` gains a real `ORPHANED_BY_FORK` arm carrying `{authoredFor, chainNow}`. The facade watches the three wallets' state versions (not `facade.state()` — that would feed back the pending set orphaning writes to), and when the observed version crosses out of an item's epoch, the existing auto-revert runs and history records `orphaned-by-protocol-upgrade`. Pinned end-to-end with real wallets + real history storage.
- **WP-11 — version-keyed proving (P1)**: `VersionedProvingService` routes on the version a transaction was **built** for — all three `BalancingRecipe` arms gain a required `protocolVersion` stamp, because the fork can land between balancing and proving, making prove-time chain version the wrong key. Config: `provingServers: [{sinceVersion, url}]`; the existing `provingServerUrl` maps cleanly to a one-entry registry (what it was implicitly claiming) so every current config site compiles and runs unchanged; the loud break is the service type — a custom injected prover now takes the version (`singleVersionProvingServiceEffect` wraps a single-version backend). v8 local WASM proving is expressed as *absence* (the bundled prover registers only from the version its key material covers) → typed `UnsupportedProvingVersionError`.

## Two discoveries reviewers should know

- **External ask #4 was never external.** `protocolVersion: Int!` has been on the indexer's `type Block` all along (`indexer.gql:51`), and `DustLedgerEvent.protocolVersion` was already selected by the `DustNullifierTransactions` subscription — only our selection sets omitted it. The comment in dust v1 sync claiming this "waits on confirmation from a deployed indexer" was stale. This also makes the WP-8b start-probe cheap.
- **Unshielded has no eager-decode problem** (its sync schemas are structured records, no ledger imports). Dust has the shape shielded's flip had to fix, in four precisely-located places, with the delta to shielded's `readEvent` model mapped — the dust flip increment starts from a survey, not a hunt.

## Honest seams (documented in code + changesets)

- The facade's trait registry registers the **same v9 trait on both sides of the fork** — honest today (the facade only carries v9 transaction objects) and load-bearing (the two epochs are what make orphaning fire); a genuine v8 trait needs the Flow 2/3 transaction union.
- `finalizeTransaction(tx)` without an explicit version falls back to the observed chain version — the wrong key by this PR's own reasoning, but the bare `UnprovenTransaction` has nowhere to carry a stamp; the optional version parameter is the mitigation and `finalizeRecipe` always routes on the recipe's stamp. Properly closed by Flow 2/3's handles.
- One existing test's **fakes** were repaired (`facade/test/submission.test.ts`): `vi.mockObject` drops accessors, so the doubles had no `state` for the version watcher to subscribe. A documented helper restores the real accessor; zero assertions changed (diff-verified).

## Tests & gate

43 new tests (codec 6, blockData 3, dust schemas +7, versioned trait 12, orphan-on-fork e2e 3, versioned proving 9+3), red-first (one disclosed fixture correction in the codec spec, assertions unchanged). Gate: dist 17/17, typecheck 37/37, lint 58/58, unit 53/53 tasks (**985 tests / 120 files**), shielded fork-simulation integration 3/3, effect-language-service clean on all 36 changed files (all remaining hits pre-existing on untouched lines), format + changesets green. Changesets: capabilities majors, facade majors, dust-wallet major, indexer-client minor — prover-client deliberately none (nothing in that package changed; the plan's expectation was stale).


agronmurtezi:

PR #680, opened 2026-08-20
feat(dust)!: a genuinely two-variant production dust wallet

agronmurtezi:
Sixth increment of the dual-ledger public API redesign — the dust flip, applying the wallet-layer shape the shielded lane defined (#676–#678). Stacked on #679.

## What's here

- **Lazy decode (the flip's prerequisite)** — dust had the eager-decode deadlock shape in four places, two of which reach back past the boundary *by construction* (the nullifier lookup runs from block 0): sync events in both twins, matched nullifier-transaction events, nullifier-block `LedgerParameters`, collapsed-tree updates and dtime insertion paths. All now cross the schema boundary as raw hex and are decoded only by the variant that applies them (shielded's `readEvent` model), red-first with the other ledger's real bytes — verified live that the ledgers refuse each other's payloads (`midnight:ledger-parameters[v5]:` vs `[v8]:`). The nullifier path deliberately *skips* what a version cannot read (`tryReadEvent`), because a prefix match over-delivers by design. Also: `DustLedgerEvents` now selects `protocolVersion`, and the stale comment claiming that awaited indexer-team confirmation is deleted.
- **The flip**: `CustomForkingDustWallet(configuration, preFork, postFork)`; `DustWallet(config)` registers v1 since `MinSupportedVersion` and v2 since required `config.forkVersion`, each variant self-configured. `CustomDustWallet` unchanged, still single-variant. The fork-crossing proof (`forkSimulation.test.ts`) rewired to the production factory with a **one-line diff** (import path) — zero assertion changes. Start behavior pinned in `forkStart.test.ts` (13 tests, ≥/<-`forkVersion` semantics): v9-native chains boot v1, apply nothing, hand over, re-discover dust via replayed events; pre-fork chains stay on v1.
- **Per-variant `DustParameters` rebuild** (`asPreForkDustParameters`): the three scalar rates cross, the WASM object does not — the wrong-parameters hazard the old test factory could only flag in a comment is now unrepresentable by construction.
- **Temporary pre-fork seam** (`PreForkDustTransactingUnsupportedError`, same GA wording as shielded): the seven v9-typed transaction-building/fee methods are gated on the v1 branch. Deliberately NOT gated: `revertTransaction` (facade reverts all three wallets together; the pre-fork variant cannot have booked anything), `splitNightUtxosForDustRegistration` (plain arithmetic over plain data — gating would break pre-fork fee estimation for nothing), and all of sync/state/`balance(date)`/serialize/restore/migration.
- **A state-encapsulation leak found and fixed**: `DustWalletState` exposed `capabilities`/`services` publicly typed at v2 — and the facade reached through it. Both members removed; the one used projection (`splitNightUtxos`) is now a bound method, and the state's version union surfaces on `state` alone (the F2 pattern, verified field-by-field for dust).

## The finding to read: projections fast-sync × the flip

**A two-variant dust wallet can never use projections fast-sync, and WP-8b does not fix that.** The pre-fork variant has no projections path at all (a permanent, user-decided exclusion — v8 lacks the four required APIs), so the shipped `DustWallet(config)` always boots into event replay; a start-at-current-version probe would remove the spurious migration but still not make fast-sync reachable — fast-sync inherently means composing a **single post-fork variant**. That is exactly what `docs-snippets/dust-fast-sync.ts` and the projections e2e already do via `CustomDustWallet`; both are unaffected, the snippet now states the trade, and the whole finding is pinned by two tests rather than prose.

## Sweep

Near-zero: the facade's `DefaultConfiguration` intersection already requires `forkVersion` via shielded (same field, same type — verified: zero facade call sites changed). The dust-specific sweep is the testkit's separate dust config builder (`DustWalletConfiguration` → package-owned `DefaultDustConfiguration`, `getDustWalletConfig()` supplies `V9_NATIVE_FORK_VERSION`) — hence the testkit major.

## Tests & gate

28 new tests (lazy decode 5+10, forkStart 13) + the rewired proof; disclosed non-mechanical repairs with byte-identical assertions (a structural fake made unusable by lazy decode was split into `matchedDustSpends` + re-pinned coverage). Gate: dist 17/17, typecheck 37/37, lint 58/58, unit 53/53 tasks (dust **184 tests / 25 files**), shielded fork-simulation integration passing, effect-language-service clean on all changed files, format + changesets green. Changesets: dust-wallet major, facade major, testkit major, indexer-client minor.


agronmurtezi:

PR #681, opened 2026-08-20
feat(unshielded)!: a genuinely two-variant production unshielded wallet

agronmurtezi:
Seventh increment of the dual-ledger public API redesign — the unshielded flip, completing the set: **all three wallets now register a variant on either side of the fork.** Stacked on #680.

## What's here

- **The flip**: `CustomForkingUnshieldedWallet(configuration, preFork, postFork)`; `UnshieldedWallet(config)` registers v1 since `MinSupportedVersion` and v2 since required `config.forkVersion`, each variant self-configured; `CustomUnshieldedWallet` unchanged. The fork-crossing proof rewired with **zero assertion changes** (wiring only — the proof previously named a pre-built pre-fork *state*; it now drives the production `startWithPublicKey` identity path, which is why the diff is a few lines rather than dust's one).
- **The honest survey result**: unshielded retains no key material at all (sync reads the state's own public address), so `StartMaterial` deliberately does NOT apply — no machinery was forced where nothing derives. What IS version-bound is the start identity: `startWithPublicKey` takes the v9 `{tag, value}` key while v8 keys are bare hex. Resolution, pinned by tests: a **schnorr** key narrows losslessly and starts on v1 (migration re-widens it); an **ecdsa** key starts directly at the post-fork variant via `startAtVariant`, stamped with `forkVersion` — narrowing it would create a v8 wallet claiming an identity it does not have (ecdsa derives a different address), corrupting restore and signing. This is the first production use of the restore-routing `startAtVariant` primitive outside `restore()`.
- **State encapsulation**: the same `capabilities`/`services` public leak dust had (here with zero external readers) is removed; projections are bound per emission in the `variantTag`-narrowed branch. One load-bearing departure from the dust template: the capability requirement is narrowed with `Pick<>` to exactly the projected methods, because the whole `KeysCapability` cannot be satisfied two-variant (`getPublicKey` returns `string` on v1 vs `SignatureVerifyingKey` on v9) — the narrowing keeps the wallet buildable while keeping the unprojectable member out of reach.
- **Temporary pre-fork seam** (`PreForkUnshieldedTransactingUnsupportedError`, same TEMPORARY/MUST-NOT-SURVIVE-GA wording): the eight v9-typed balance/transfer/swap/sign/rotate methods are gated on the v1 branch; `revertTransaction` is deliberately a pre-fork no-op (consistent with shielded and dust — the facade reverts all three wallets together, and the pre-fork variant booked nothing). Sync, state, balances, serialize/restore, and migration work fully on both branches.
- **Start pins** (13 tests, ≥/<-`forkVersion` semantics): v9-native start = boot v1, apply nothing, migrate via **empty structural carry** (`appliedId === 0n` both sides — unshielded carries UTXOs with an unchanged cursor, unlike shielded/dust's fresh-state replay), v2 syncs the whole chain; pre-fork start = stays v1 in the bare-hex v8 signer world with the shared address. Plus 9 restore-routing tests mirroring shielded's.
- **Eager decode confirmed absent** (the #679 survey held: no ledger import in either twin's wire schemas) — checked, not inherited.

## Sweep

Facade churn: **zero** (diff-verified empty against facade/apps/docs-snippets — the `DefaultConfiguration` intersection collapsed onto the `forkVersion` field shielded and dust already require). Touched: testkit's `restoreUnshieldedWallet`, five e2e fixtures (all reading `forkVersion` from their environment's one definition), the wallet-integration-tests config, unshielded's own test utils and README. One barrel collision resolved: unshielded's restore module exports as `UnshieldedRestore` (dust already claimed the unqualified names in the umbrella).

## Tests & gate

22 new tests + the rewired proof; one disclosed post-red edit (a snapshot-peek assertion re-routed through the `peekProtocolVersion` production function — equivalent-or-tighter, lint-driven). Gate: dist 17/17, typecheck 37/37, lint 58/58, unit 53/53 tasks (unshielded **211 tests / 22 files**), shielded fork-simulation integration passing (cache + direct vitest run), effect-language-service 0 findings on all 17 changed files, format + changesets green. Changesets: unshielded-wallet major, testkit patch-level touch.

### DIFF
--- a/.changeset/unshielded-two-variant-fork-flip.md
+++ b/.changeset/unshielded-two-variant-fork-flip.md
@@ -1,0 +1,71 @@
+---
+'@midnightntwrk/wallet-sdk-unshielded-wallet': major
+'@midnightntwrk/wallet-sdk-testkit': patch
+---
+
+`UnshieldedWallet(configuration)` is a two-variant, fork-crossing wallet, and `forkVersion` is required.
+
+The shipped unshielded wallet now registers one variant either side of a protocol boundary — the pre-fork ledger version
+from the minimum supported version, the post-fork one from `configuration.forkVersion` — and follows the chain across
+it. The crossing itself is unchanged and still proven by `test/forkSimulation.test.ts`, which was rewired onto the
+shipped factory with **no assertion changed**: the test-only two-variant builder it used to stand on has been dissolved
+into `test/forkHarness.ts`, which is now observation and simulated infrastructure around the real wallet.
+
+Unshielded's crossing is a **structural carry**, not a fresh state plus replay. Its UTXOs are public ledger data the
+wallet holds as plain records, so every one of them crosses field for field, along with the identity, the network and
+the parked cursor — and the wallet never passes through a state in which it has forgotten what it owns. The boundary
+transaction is applied exactly once, by the new variant: the old one records its version and refuses every part of it,
+which leaves the cursor pointing just before it for the new variant to re-fetch from.
+
+**`forkVersion` is required on `DefaultUnshieldedConfiguration`, without a default.** A wrong value does not degrade —
+it decides which ledger version reads the chain. `@midnightntwrk/wallet-sdk-shielded` publishes `V9_NATIVE_FORK_VERSION`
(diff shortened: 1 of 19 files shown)


agronmurtezi:

PR #682, opened 2026-08-20
feat(abstractions,wallet-sdk): the transaction handle and the named ledger subpaths

agronmurtezi:
Eighth increment of the dual-ledger public API redesign — the Flow 2/3 primitives. Stacked on #681. This PR deliberately stops at the primitives boundary: its survey found the downstream cascade (the `secretKeys` drop, the three seam closures, the facade rewiring) is strictly ordered with no gate-green intermediate, and three prep refactors belong ahead of it (see "What comes next").

## What's here

- **`WalletTransaction<Stage>`** (`abstractions` — ledger-free, reaches the umbrella root): the opaque carrier handle. A class with a private `#carried` field and private constructor — nominal in TS, payload unreachable; `adopt(stage, tx, version)` seals author output in; `unwrapWithin(handle, range)` is the ONE unwrap primitive, taking a `ProtocolVersion.Range` rather than a version, because enforcement means "same side of the boundary", not stamp equality (a pre-fork chain reports many versions; exact-stamp stays expressible as a one-wide range). `ProtocolVersionMismatchError {authoredFor, accepted, stage}` and `WireFormatError` are typed.
- **Stages are three, not four**: `'Unproven' | 'Unbound' | 'Finalized'` — `ProvenTransaction` appears nowhere in the repo; `UnboundTransaction` *is* the proven-not-bound stage. Consumer aliases: `UnprovenTx`, `UnboundTx`, `FinalizedTx`, `AnyTx` (covariant stage).
- **Wire envelope**, minimal and versioned: `{ wireFormat: 1, protocolVersion, stage, transaction: hex }` — `wireFormat` is separate from `protocolVersion` so a reader can distinguish "unknown envelope" from "unknown ledger version"; `fromWire` takes the decoder from the caller (this package holds no ledger) and converts throws into typed errors. Deliberately conservative — connector coordination is an open external ask.
- **`wallet-sdk/ledger/v8` and `wallet-sdk/ledger/v9`** author subpaths (real dist targets, publint-clean), with root-barrel ledger-freedom verified two ways: zero ledger imports in the barrel source, and a test asserting no ledger symbol appears in the root's runtime exports. The v9-only-authoring-throws-until-the-fork consequence is documented in both subpath headers, the changeset, and a new README "Authoring transactions" section. The package gained its first test runner.
- The handle is pinned against **real WASM objects of both ledgers** (`v8/v9.Transaction.fromParts`), which is what validates its structural constraint against the actual classes.

## What comes next (the survey's findings, ahead of the seam closures)

1. `ValidationService.validateTx` is a closed v9 union with no type parameter — the only subsystem with no version story; every balance path calls it.
2. `UnboundTransaction` is defined identically in four places (three exported publicly) and must collapse to one owner.
3. `AnyTransaction` is a dust-internal v9 union re-exported through facade public signatures and needs an owner.

Also corrected from the plan: `FacadeState.pending` does have a consumer (`dust-sponsorship.ts` reads `pending.all.length`), and the carrier blast radius is 47 consumer files, not ~33.

## Tests & gate

26 new tests (handle 19, subpaths 7), red-first (one disclosed test-defect fix after green: a fixture-closure identity comparison replaced with strictly more precise value assertions). Gate: dist 17/17, typecheck 37/37, lint 58/58, unit 54/54 tasks (the new wallet-sdk runner joins), shielded fork-simulation integration 3/3, effect-language-service clean on all changed files, format + changesets green, publint clean. Per-wallet totals unchanged (shielded 180/dust 184/unshielded 211) — this PR touches none of them; all three temporary seams remain in place for the closure increment.

### DIFF
--- a/.changeset/wallet-sdk-ledger-subpaths.md
+++ b/.changeset/wallet-sdk-ledger-subpaths.md
@@ -1,0 +1,23 @@
+---
+'@midnightntwrk/wallet-sdk': minor
+---
+
+**Both ledger versions now ship from the umbrella package, at `./ledger/v8` and `./ledger/v9`.** An application that
+authors its own transactions has always had to import a ledger package directly — depending on something this package
+does not promise, and pinning a version by hand in its own `package.json`. Now it imports the one package it already
+depends on, and the import line says which ledger version's rules the bytes follow.
+
+The subpaths are named for the **ledger** version rather than a variant ordinal, because that is what an author is
+choosing. Seal what you build with `WalletTransaction.adopt('Unproven', tx, version)` to hand it back to the wallet.
+
+**Neither is re-exported from the root barrel.** These are WebAssembly modules, and an application that only carries
+transactions should not pay for a ledger it never names; having to ask for one by name is what keeps that true.
+
+**Which one to import is a property of the chain, not of this release.** Both are shipped side by side because both are
+real: a chain is pre-fork until it forks, and mainnet is pre-fork until then. An authoring path that only ever imports
+`./ledger/v9` compiles cleanly and **fails at run time on every chain that has not yet crossed**, with a
+`ProtocolVersionMismatchError` — the wallet refuses a transaction built for a version it is not acting at. Read the
+protocol version from the wallet's state and author against the version the chain is on. The check is a run-time one by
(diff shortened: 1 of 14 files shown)


agronmurtezi:

PR #683, opened 2026-08-20
feat(capabilities,facade): version-routed validation and one owner per transaction type

agronmurtezi:
Ninth increment of the dual-ledger public API redesign — the three plumbing refactors the #682 survey put ahead of the seam-closure cascade. Stacked on #682. All three are green intermediates; the seams themselves remain untouched.

## What's here

- **Version-routed `ValidationService`** — mirrored on the proving shape: `ValidationServices` = a `ProtocolVersion.Registry` of per-version validators, `makeVersionedValidationServiceEffect` selecting on the version a transaction was **authored** for (stamp, not chain), typed `UnsupportedValidationVersionError` miss, `singleVersionValidationServiceEffect` compat wrapper. Existing types gained parameters **defaulting to the current ledger**, so zero call sites changed — the changeset is additive/minor. (Survey correction along the way: `validateTx` has exactly ONE caller — the facade's `validateTransaction`; no balance path calls it.)
- **A genuine ledger-v8 validator, shipped, not deferred**: `makePreForkValidationServiceEffect` really calls v8's `LedgerState.blank`/`wellFormed` and is pinned against real v8 transactions (pass, wrong network, expired TTL, fetch given/failed). This corrects a plan asymmetry: "no v8 proving path exists" is true for proving, but v8 *validation* is fully constructible today. **The one remaining gap is wiring, stop-and-reported for the closure increment**: nothing registers the v8 validator yet because `defaultLedgerParametersCodecs` holds a single open-ended v9 entry — closure task #1 is a v8 codec entry + routing the block-data fetch on the block's reported version.
- **`UnboundTransaction` collapsed — it was three, not four**: the shielded and unshielded-v2 duplicates re-export capabilities/proving's declaration (the dependency-graph owner; shielded's was a dead alias); unshielded-v1's is `@midnight-ntwrk/ledger-v8`-typed — a genuinely DIFFERENT type, kept with a comment, and pinned non-assignable via a compiler-enforced `@ts-expect-error`. Public entry points keep exporting the same names (verified in emitted `.d.ts`).
- **`AnyTransaction` gets an owner**: the facade was using a dust-*variant-internal* type in five public signatures while exporting nothing — a facade-only app had to import from `dust-wallet/v2`. The facade now exports its own `AnyTransaction` (deliberately an alias, so it cannot drift from what the methods accept), with the closure-increment pointer on the declaration.

## Tests & gate

12 new tests across 4 files — the two validation suites red-first behavioural (routing, stamp-not-chain, typed miss, real v8 `wellFormed` semantics), the two type pins compiler-enforced and honestly disclosed as such (structurally-identical duplicates cannot produce a runtime red). Zero type casts added. Gate: dist 17/17, typecheck 37/37, lint 58/58, unit 54/54 tasks, shielded fork-simulation integration passing, effect-language-service 0 findings on all 11 changed files, format + changesets green. Changesets: capabilities minor (additive routing + the v8 validator), facade minor (additive `AnyTransaction`), shielded/unshielded patches (same names, one declaration).

### DIFF
--- a/.changeset/facade-owns-the-transactions-it-accepts.md
+++ b/.changeset/facade-owns-the-transactions-it-accepts.md
@@ -1,0 +1,22 @@
+---
+'@midnightntwrk/wallet-sdk-facade': minor
+---
+
+The facade now exports `AnyTransaction`, the type its own public signatures already took.
+
+`calculateTransactionFee`, `estimateTransactionFee`, `revert`, `revertTransaction` and
+`BalancingRecipe.getTransactions` all name this type, but it was reachable only as
+`@midnightntwrk/wallet-sdk-dust-wallet/v2`'s `AnyTransaction` — a variant-scoped internal of a package a facade-only
+app need not depend on at all, and one with a differently-typed pre-fork twin under `/v1`. Annotating a variable, or
+writing a wrapper around any of those methods, meant importing from that subpath.
+
+Additive: `import { type AnyTransaction } from '@midnightntwrk/wallet-sdk-facade'`. Existing imports from the dust
+subpath keep working — the facade's name is an alias to that same type, deliberately so it cannot drift from what the
+methods accept.
+
+**Ownership decision.** The facade names it, rather than dust promoting it to a documented public export, because the
+facade has a single entry point and was exporting nothing for a type in five of its public signatures, while dust's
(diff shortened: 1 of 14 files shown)


agronmurtezi:

PR #684, opened 2026-08-20
feat(facade,wallets)!: pre-fork transacting, carried on version-stamped handles

agronmurtezi:
Tenth increment of the dual-ledger public API redesign — the seam-closure cascade. Stacked on #683. **Pre-fork transacting is now real**: all three temporary `PreFork*TransactingUnsupportedError` seams are deleted (their must-not-survive-GA promise fulfilled), and the facade carries opaque version-stamped handles end to end.

## What's here

- **The three seams are gone.** Each wallet's v1 branch builds through its own variant `Transacting` with its own derived keys; every result travels as a `WalletTransaction` handle stamped with the **epoch floor** of the variant that built it (the one value guaranteed to sit inside the epoch every routing decision keys on). The 18 refusal pins became 40 working-path pins across the three `forkStart` suites (8/16/16) — including "spends what it holds, so a pre-fork transfer is a real one" (a genuine `v8.Transaction` with real inputs), "pays a fee out of the dust it actually holds" (v8 `LedgerParameters` in the recipe's block data), cross-boundary refusals naming both versions, and ecdsa refusals.
- **`secretKeys` parameters are dropped** from transfer/balance/swap across the wallets and facade (~70 deletions swept through e2e/docs-snippets/testkit; two now-unrepresentable "wrong secret key" tests removed). The wallets' retained `StartMaterial` was already the source of truth.
- **The facade speaks handles**: `UnprovenTx`/`UnboundTx`/`FinalizedTx`/`AnyTx` across every transaction signature (full table in the PR description's first commit). Internally there is ONE unwrap point — `accept<T>()` within the current epoch — and the facade's own merge/bind/identifiers go through four minimal structural types both ledgers satisfy, so per-version dispatch exists at exactly one place (throwaway-signature authoring for registration estimates). `finalizeTransaction` takes a handle; the stamp is authoritative; the observed-version fallback is deleted.
- **Enforcement (user decision 4)**: at every facade entry point (submit, finalize, merges, dust-registration signing) a handle from the other side of the boundary is refused with `ProtocolVersionMismatchError`. Two deliberate exceptions, documented: `validateTransaction` routes on the transaction's OWN stamp (well-formedness is the producing ledger's question — a later fork cannot change the answer), and `revertTransaction` treats a mismatch as "nothing of mine to release" (the facade reverts all three wallets together on failed submission).
- **Validation is fully wired** (closing #683's stop-and-report): the default codec registry splits at `forkVersion` (v8 below, v9 from it), block data is decoded at the block's REPORTED version, and the genuine v8 validator is registered below the boundary. Pinned: a pre-fork transaction validates through real v8 `wellFormed`; cross-epoch offers refuse.
- **The pending registry carries a genuine ledger-v8 trait** below the boundary — each per-ledger trait lifted onto handles (`overHandles`), unwrapping at its epoch, adopting on deserialize, disowning the other epoch's handles. The plan's "pending v8 txs are silently orphaned today" is now false. `FacadeState.pending` kept its shape (only the transaction type changed) — the dust-sponsorship snippet needed zero adaptation.
- **A minimal internal signature adapter** (`capabilities/signatures`: lower/lift between v8 bare-hex and v9 `{tag, value}`, typed `UnsupportedSignatureKindError`) — in scope by necessity (dust registration and the unshielded signer carry these shapes); this is NOT the Flow-7 public `Signature` change, which stays in the next increment.

## Findings for reviewers

- **v8 and v9 `LedgerParameters` are mutually assignable** (structurally identical), so `AnyLedgerParameters` is a statement of intent and the codec routing has to be *right* rather than merely well-typed — hence the decode-at-reported-version pins. The `Transaction` types are nominally distinct, so those unions are real.
- Pre-fork **proving** remains typed-unavailable without a registered pre-fork proving server (external ask: proof-server-8) — building, balancing, validating, and fee calculation all work; three facade unit suites inject a pre-fork mock prover for exactly this reason.
- Simulator-backed facade tests now use `forkVersion: MinSupportedVersion` — a single-ledger simulator has never had two epochs, and pretending otherwise made those facades nominally pre-fork while every object was v9.
- Two harness-limit pins are stop-and-reported honestly rather than faked: dust's `attachDustRegistration` can only act on an intent the *unshielded* wallet builds, and unshielded's `rotateUtxos` cannot complete on plain sync fixtures — each pin asserts the true reachable behavior.

## Tests & gate

Red-first throughout (the pin flips written first and confirmed red against the still-gated implementation, then gate + old pin deleted together). Gate: dist 17/17, typecheck 37/37, lint 58/58, unit 54/54 tasks (shielded 182 / dust 187 / unshielded 215 / facade 46 / capabilities 182), shielded fork-simulation integration 3/3, effect-language-service 0 findings on all changed files, format + changesets green. 58 files changed; sweep resolved to 28 consumer files actually needing edits (the survey's 47 included files that turned out to import only version-stable names). Changesets: three wallet majors + facade major + capabilities minor.

### DIFF
--- a/.changeset/abstractions-epoch-of.md
+++ b/.changeset/abstractions-epoch-of.md
@@ -1,0 +1,8 @@
+---
+'@midnightntwrk/wallet-sdk-abstractions': minor
+---
+
+`ProtocolVersion.epochOf(version, forkVersion)` names the range of protocol versions on the same side of a protocol
+boundary as a given version — what the SDK means by "the same ledger version made these bytes". Everything that routes
+on a version is really asking which epoch it belongs to, so the boundary is computed in one place and the two ends of
+that question, the wallet stamping a transaction and the caller unwrapping it, cannot disagree.
--- a/.changeset/capabilities-validate-and-lower.md
+++ b/.changeset/capabilities-validate-and-lower.md
@@ -1,0 +1,26 @@
+---
+'@midnightntwrk/wallet-sdk-capabilities': major
+---
+
+Validate a transaction at the version it was authored for, on a chain with a protocol boundary.
+
+BREAKING CHANGE — `defaultLedgerParametersCodecs` is now a function of the chain's fork version and returns a registry
+**split at it**: a block reported from before the boundary is read with the previous ledger version's deserializer and
+one from the boundary with the current one, so the parameters handed to a validator are always ones its own
+`LedgerState` accepts. `makeDefaultBlockDataFetcher`'s configuration gains a required `forkVersion` for the same
+reason.
+
+`makeDefaultValidationServices` and `makeDefaultVersionedValidationService` register a validator either side of the
+boundary — the shipped pre-fork validator is no longer merely constructible but actually reachable — and route on the
+version a transaction was authored for.
+
+BREAKING CHANGE — `makeDefaultValidationService` and `makeDefaultValidationServiceEffect` take their block data as
+`AnyLedgerParameters`, the union naming what a fetcher spanning a boundary can yield. The two ledger versions'
+`LedgerParameters` are structurally identical, so this is a statement of intent rather than something the compiler
+enforces; the distinction is nominal at run time, which is why the routing has to be right rather than merely
+well-typed.
+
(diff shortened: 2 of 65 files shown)


agronmurtezi:

PR #685, opened 2026-08-20
feat(facade,wallets,hd)!: the ledger-free public surface — seeds, tokens, statuses, one vocabulary

agronmurtezi:
Eleventh and final substantive increment of the dual-ledger public API redesign — Flows 1, 4, 5, 6, 7. Stacked on #684. **After this PR, an application imports a ledger package for exactly one reason: hand-authoring transactions** (via `wallet-sdk/ledger/v8|v9` + `adopt`, by design).

## Flow 1 — keys & start (user decision 1, implemented)

- **`WalletSeeds.fromMasterSeed`** (`packages/hd` — it owns BIP32; no new dependency): one name for the master-seed → per-wallet-seeds walk that six hand-rolled copies were writing across testkit/e2e/snippets. Pinned differentially against the hand-rolled walk plus golden hex vectors. *Honest gap: the spec-reference vectors treat per-wallet seeds as given, so this pins current behavior — adding master-seed derivation to the spec + vectors is a follow-up.*
- **Facade `start(seeds)`** is the primary start (one discriminated parameter, `WalletSeeds | FacadeKeysByEpoch`); the escape hatch `start({v8, v9})` requires BOTH epochs' key objects per the decision.
- **`startWithSecretKeys`/`startWithSecretKey` deleted on the forking classes** — replaced by `startWithKeys({v8, v9})` with both required; every old call site migrated (map in the increment report). Deliberate boundary: the single-variant `Custom*Wallet` classes KEEP their single-key starts — one variant means one key object is the whole truth (the fast-sync dust composition depends on it); the shipped `ShieldedWallet`/`DustWallet` are the forking classes, so the public break is real.
- **Seed hygiene, literal**: seed-accepting starts derive BOTH epochs' keys eagerly and drop the seed reference with the calling frame; key objects are released at `stop()`. Retention is a precise per-epoch product (`{preFork: Option<v8.X>, postFork: Option<v9.X>}`) rather than the homogeneous `StartMaterial` map — zero casts, models both-required directly. JSDoc states plainly that JS/WASM release references without zeroization guarantees.
- `DustParameters` → optional plain scalars on every start path.

## Flow 4 — tokens & protocol phase

- SDK-owned `TokenType = string` (documented unbranded exception) + **`Token.night`** + `parseTokenType`. **One constant, deliberately**: the dual-ledger pin proves `nativeToken().raw` ≡ `shieldedToken().raw` ≡ `unshieldedToken().raw` ≡ 64 zeros on BOTH ledgers — the raw type does not distinguish forms (the API being called does), so a `Token.shieldedNative` alias would assert a distinction that doesn't exist. Zero `nativeToken()` calls remain in production source; the last two snippets using `shieldedToken()` now say `Token.night` (adjudicated during verification), making every non-authoring snippet ledger-free.
- Additive **`FacadeState.protocol: ProtocolPhase`** — `Settled {version}` when all three wallets are in the same epoch, `Crossing {from, to, behind}` otherwise (within-epoch version spread is ordinary sync, not a crossing). Renamed from the plan's `ProtocolState` (root-barrel collision).

## Flow 5 — `tryRestore`

Additive Either-returning restore on the three forking classes; `restore` = `getOrThrow(tryRestore(...))`, one implementation. Typed per-package errors (`ShieldedRestoreError` etc.). One disclosed test-design correction, coverage strictly increased.

## Flow 6 — pending statuses

`FacadeState.pending: readonly PendingTransaction[]` — `{transaction, submittedAt, authoredFor, status}` with tagged `Submitted | Confirmed | Rejected | Orphaned {authoredFor, chainNow}` — a pure projection over the existing machinery (no persistence change). Both real consumers adapted (the plan's zero-consumers claim was false: the dust-sponsorship snippet, plus one facade test).

## Flow 7 — the umbrella

Root barrel exports seeds/tokens/handles/protocol/signing; the ledger-freedom test extended over all of it. **Public `Signature`/`SignatureVerifyingKey` adopt v9's `{tag, value}` shape** in `abstractions/Signing` — the internal #684 adapter now states its current-shape side in these terms (one definition; `import type` only, no WASM) — and `UnsupportedSignatureKindError` is promoted to the root with an identity pin. v9-typed signers compile unchanged, as promised.

## Tests & gate

Red-first throughout; deletion map old-test → new-test complete; forkStart suites preserved untouched (8/16/16). Gate: dist 17/17, typecheck 37/37, lint 58/58, unit 54/54 tasks (shielded 186 / dust 191 / unshielded 219 / facade 61 / abstractions 76 / hd 17 / wallet-sdk 17), fork-simulation integration 3/3, effect-language-service 0 findings on all 57 changed files, format + changesets green. Changesets: facade + shielded + dust majors (deleted starts, start surface, pending reshape), hd minor, abstractions/capabilities/wallet-sdk per impact.

### DIFF
--- a/.changeset/facade-pending-statuses.md
+++ b/.changeset/facade-pending-statuses.md
@@ -1,0 +1,24 @@
+---
+'@midnightntwrk/wallet-sdk-facade': major
+---
+
+**BREAKING:** `FacadeState.pending` is now `readonly PendingTransaction[]` with a tagged `PendingStatus`, in place of
+the pending set the services work in.
+
+Each entry carries the transaction handle, when it was submitted, the protocol version it was authored for, and one of
+four statuses: `Submitted`, `Confirmed`, `Rejected` (with the segments the chain reported), and `Orphaned` — which
+carries `authoredFor` and `chainNow`.
+
+`Orphaned` is deliberately its own arm rather than another rejection. A rejection is the chain's verdict: the node saw
+the transaction and refused it. An orphaned transaction has no verdict at all and never will, because its bytes were
+authored under a protocol version the chain has moved past and nothing can include them afterwards. The wallet has
+already unbooked the coins and recorded the rejection either way; what differs is what an application can tell a user,
+and whether resubmitting the same bytes could ever help. The distinction existed underneath already — it was only
+reachable by comparing an indexer-shaped status string against a constant the SDK did not export.
+
+The persistence story is unchanged: pending transactions remain session-scoped and are orphaned on a fork.
+
+```diff
+- rx.filter((s) => s.pending.all.length === 0)
++ rx.filter((s) => s.pending.length === 0)
+```
--- a/.changeset/facade-protocol-phase.md
+++ b/.changeset/facade-protocol-phase.md
@@ -1,0 +1,12 @@
+---
+'@midnightntwrk/wallet-sdk-facade': major
+---
+
+**BREAKING:** `FacadeState`'s constructor takes the fork version as a fifth argument.
+
+Adds `FacadeState.protocol`, additively: a tagged reading of whether the three wallets are settled on one side of the
+protocol boundary or still crossing it. Three protocol versions cannot say this on their own — a difference within one
+epoch is ordinary synchronization, not a crossing — and around a fork the three wallets disagree for a while because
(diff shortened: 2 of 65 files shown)


agronmurtezi:

PR #686, opened 2026-08-20
feat(wallets,capabilities)!: ask the chain where it is before choosing a variant to start at

agronmurtezi:
Twelfth increment of the dual-ledger public API redesign — WP-8b, promoted from post-GA optimization to a GA correctness item. Stacked on #685.

## Why promoted

Two-variant wallets boot the head (pre-fork) variant and learn the chain's version only from observed events. That meant (a) every start on a v9-native chain paid a spurious v8-boot → migrate → v2-resync dance, and (b) — the sharper hazard #684 created — a wallet on a chain that has shown it **no events** stays in the pre-fork epoch and now silently *builds* ledger-v8 transactions against a v9 chain, failing only at the node. The probe pins the correct epoch from the first moment.

## What's here

- **`chainVersionProbe`** (`capabilities/chainVersion`): reuses the existing `BlockHash` query (which already selects `protocolVersion` — zero indexer-client changes), `Option`-shaped core (`absent`, not zero, for a chain with no block), 5-second ceiling, typed `ChainVersionUnavailableError`.
- **Wired into the three forking wallets' start paths** (`startWithSeed`/`startWithKeys`/`startWithPublicKey` — never `restore`, whose snapshot version is authoritative): probe → `variantFor` → `startAtVariant` on that variant's own fresh state, annotated inside its activation range (the same construction unshielded's ecdsa start already used). Default factories wire the probe from the config's existing `indexerClientConnection`; `CustomForking*` callers get an optional field — absent means today's behavior exactly.
- **Best-effort by design**: no probe, probe rejected, probe timed out, and unclaimed version all collapse — deliberately undistinguished — to the head-variant boot, pinned per wallet alongside the probed paths. Offline-first apps and simulators cannot newly fail to start.
- **The breaking part, framed honestly**: the forking classes' start methods return `Promise<Wallet>` — the variant must be settled before the runtime exists, so there is no synchronous way to achieve "the pre-fork variant never activated". Blast radius was small: wallet factories were already typed `MaybePromise`, so every `(config) => Wallet(config).startWithSeed(...)` closure is unchanged; direct callers gained `await`. Single-variant `Custom*` starts stay synchronous.
- **Not claimed**: the v8 WASM module still loads at import time (value import + eager both-epoch key derivation) — the probe removes the migration and the epoch error, not the load; making the load lazy is separate work.

## Pins (second sanctioned pin-flip of the line, mapped old → new in the report)

Per wallet: the old boot-v1 → migrate pin is **kept verbatim** under "no probe", joined by a probe-failed sibling asserting the identical fallback dance; new probed pins assert direct v2 start with **no hand-over at all**; pre-fork chains pin to v1; and the **empty-timeline pair** encodes the sharper hazard — an unprobed wallet believes it is pre-fork and refuses the chain's own transactions, a probed one is in the chain's epoch. Dust's fast-sync pin is *restated, not weakened*: still reachable only through a single-variant composition, probe or no probe. forkStart totals: shielded 13, dust 21, unshielded 22 (red-first confirmed: 13/20/22 failing before implementation).

## Gate

dist 17/17, typecheck 37/37, lint 58/58, unit 54/54 tasks (shielded 191 / dust 196 / unshielded 225 / capabilities 186 / facade 61), fork-simulation integration 3/3, effect-language-service 0 findings on all 30 changed files, format + changesets green. Live suites verified: configs already carry indexer connections, so they get the probe — and stop paying the spurious migration — for free. Changesets: three wallet majors (async starts + new default behavior + fallback semantics), capabilities minor.

### DIFF
--- a/.changeset/start-at-current-chain-version.md
+++ b/.changeset/start-at-current-chain-version.md
@@ -1,0 +1,54 @@
+---
+'@midnightntwrk/wallet-sdk-capabilities': minor
+'@midnightntwrk/wallet-sdk-shielded': major
+'@midnightntwrk/wallet-sdk-dust-wallet': major
+'@midnightntwrk/wallet-sdk-unshielded-wallet': major
+'@midnightntwrk/wallet-sdk-testkit': patch
+---
+
+A wallet that spans a protocol boundary asks the chain where it is before choosing a variant to start at.
+
+A two-variant wallet used to learn the chain's protocol version only from the events it observed, so every start began
+on the pre-fork variant. That cost two things. On a chain already past the boundary it paid a hand-over per start: boot
+the pre-fork variant, apply nothing, migrate, re-sync. And on a chain that had shown the wallet **no events at all** it
+was not a cost but a wrong answer — the wallet stayed pre-fork indefinitely, and since transacting now works on either
+side of the boundary, it built transactions with the ledger version the chain does not run, which fail at the node
+rather than at the wallet.
+
+The wallet now asks. `chainVersionProbe` answers which protocol version the chain is on; the variant that owns it is
+resolved through `variantFor` and started with a fresh state of its own, annotated with the version so it begins inside
+its own activation range. On a v9-native chain a default wallet is therefore on the post-fork variant from the first
+moment, with the right epoch before a single event has arrived — no hand-over, and no migration emitted.
+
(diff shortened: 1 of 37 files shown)


agronmurtezi:

PR #687, opened 2026-08-20
docs(changesets,claude) + test: the wrap-up audit — release notes, debts, spec vectors

agronmurtezi:
Thirteenth and final increment of the dual-ledger public API redesign — the WP-16/17 wrap-up audit over the complete line (#675–#686). Stacked on #686.

## Changesets as release notes (14 corrections)

Nothing from this beta line has been released, so the pending changesets must read correctly in aggregate at the next cut. Fourteen described superseded states in the present tense — each corrected to carry the release's real arc: the temporary transacting seams are introduced AND closed within this release (no `PreForkTransactingUnsupportedError` exists in the published surface — verified zero occurrences); `startWithSecretKeys` guidance no longer tells readers to migrate onto something the same release deleted; the spurious start migration is the *fallback*, not the default, now the probe exists; `FacadeState.pending`'s shape, `finalizeTransaction`'s arity, the wired v8 validator, and the fork-split codec defaults all state what actually ships. One release-policy flag for reviewers (deliberately unchanged): two `minor` bumps technically break *constructors* of `ProtocolState`/`VariantContext` — consistently reasoned house-rule ("consumers read, don't construct"), loudly documented, but worth a deliberate eye.

## The three debts, closed

- **Forking `restore` composition pins** (×3 wallets, `forkRestore.test.ts`): a snapshot from each epoch restores through the *shipped forking class* onto the correct running variant, with the held value surviving — both epochs in one file, so a router that always answers head (or always last) fails half. Survey confirmed this closed a real gap: nothing anywhere asserted the composed round trip.
- **`ProtocolPhase` Crossing wiring pin** (facade, 5 tests): real shipped wallets with injected state streams through the real `WalletFacade.init` — pins per-slot version reads (three *different* versions, so slot mix-ups fail), within-epoch drift = `Settled`, one/two laggards in fixed `behind` order, live transitions, and `Crossing.from === activeProtocolVersion`. Discrimination verified empirically by mutating a slot read and watching exactly the right tests fail.
- **The spec now owns `WalletSeeds.fromMasterSeed`**: a new "Per-wallet seeds" section in `docs/spec/Specification.md` (filling a gap the document itself left — it twice said seeds derive "at certain path" without saying which), a reference implementation in `packages/spec-reference` over `@scure/bip32`, and a new `seedDerivation.json` vector file (24 entries). **The substantive result: the reference implementation — sharing zero code with the SDK — reproduces `packages/hd`'s hand-written golden hex exactly**, and the existing key/address vector files are byte-identical after regeneration. `hd`'s tests now pin against the vectors (53 tests); the golden-hex pin retired as subsumed.

## A CI gap found and fixed

`packages/spec-reference` was the only package defining `test` but not `test:unit` — so **its vector verification never ran in CI** (the workflow runs `yarn test:unit`; nothing invokes bare `test`). The script is added; the spec vectors are now part of the gate (29 tests).

## CLAUDE.md (WP-17, corrections only — every referenced path verified)

The stale `ledger-v7` references become the dual-ledger reality (both `.d.ts` paths, the handle's `unwrapWithin`, version-travels-as-data); the scope-hyphen footgun is documented at first mention (`@midnight-ntwrk` vs `@midnightntwrk` — one character, two npm orgs); the three dead `dapp-connector-reference` paths point at real current exemplars (`parseTokenType` for parse-don't-validate; three real transaction-inspection test files); a new "v1/v2 Twin Convention" section states the twins discipline and its one permanent exclusion; the docs-snippets list gains its 10 missing files.

## For reviewers, not covered by any changeset

- **Spec drift, deliberately left to the spec owner**: the spec's role table calls role 4 "Metadata"; `packages/hd` repurposed it as `Roles.EcdsaUnshielded` (released). The new section describes roles 0/2/3 by name and states the different-scheme principle without asserting role 4's meaning, keeping the document internally consistent; the vectors cover role 4 by path, so nothing is unpinned.
- Vector files remain hand-copied into consuming packages (`addresses.json` precedent, now `seedDerivation.json` too) — documented in spec-reference's README with a vector→section table; still a manual step.
- One genuinely new diagnostics hit on the line (`LedgerTranslation.ts:27` `deterministicKeys`) follows the repo-wide short-tag convention every other error uses; left consistent, reported.

## Gate

dist 17/17 · typecheck 37/37 · lint 58/58 · unit **55/55 tasks, 147 files, 1246 tests** (the known dust parallel-load flake surfaced once during independent verification and passed clean on re-run) · fork-simulation integration 3/3 · spec-reference 29 vector tests now gating · effect-language-service swept across all 350 line-changed files with only the accepted pre-existing hits · format + changesets green. Regression nets untouched and passing: all six fork suites, facade orphan/submission, runtime.

### DIFF
--- a/.changeset/dust-two-variant-fork-flip.md
+++ b/.changeset/dust-two-variant-fork-flip.md
@@ -8,9 +8,9 @@
 
 The shipped dust wallet now registers one variant either side of a protocol boundary — the pre-fork ledger version from
 the minimum supported version, the post-fork one from `configuration.forkVersion` — and follows the chain across it.
-The crossing itself is unchanged and still proven by `test/forkSimulation.test.ts`, which was rewired onto the shipped
+The crossing itself is unchanged and still proven by `src/test/forkSimulation.test.ts`, which was rewired onto the shipped
 factory with **no assertion changed**: the test-only two-variant builder it used to stand on has been dissolved into
-`test/forkHarness.ts`, which is now observation and simulated infrastructure around the real wallet.
+`src/test/forkHarness.ts`, which is now observation and simulated infrastructure around the real wallet.
 
 **`forkVersion` is required on `DefaultDustConfiguration`, without a default.** A wrong value does not degrade — it
 decides which ledger version reads the chain. `@midnightntwrk/wallet-sdk-shielded` publishes `V9_NATIVE_FORK_VERSION`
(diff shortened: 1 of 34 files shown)


agronmurtezi:

PR #688, opened 2026-08-21
feat(api)!: public API for the dual-ledger wallets

agronmurtezi:
## What's in this PR

The public API for the dual-ledger wallets. Applications no longer import a ledger package.

- **Version routing** — `ProtocolVersion.Registry` selects the right implementation (variant, codec, prover, validator) for a given protocol version.
- **Transactions carry their version** — `WalletTransaction` records which protocol version built each transaction; using it on the wrong side of the fork fails with `ProtocolVersionMismatchError`. Apps that build their own transactions import `wallet-sdk/ledger/v8` or `/v9` explicitly.
- **Both variants active** — all three wallets register both variants and switch at `configuration.forkVersion` (now a required setting; a 2.x chain reports 2000000). Transacting works on both sides of the fork.
- **Starting** — from a seed (`WalletSeeds.fromMasterSeed` → `facade.start(seeds)`), which works across the fork; or from key objects for both ledger versions. `startWithSecretKeys` is removed. On start, wallets ask the chain its version and begin on the matching variant.
- **Fork behavior** — proof servers are configured per version (`provingServers`); transactions left pending at the fork are marked orphaned and their funds released; facade state reports whether the wallets are settled on one version or mid-crossing.
- **Release notes and spec** — the pending changesets are rewritten to describe what this release actually ships; seed derivation is added to the wallet specification with test vectors that now run in CI; new tests cover restore and the crossing status.
- **New snippet** — `hard-fork-support.ts`, a runnable example of all of the above.

## How to review

1. `packages/docs-snippets/src/snippets/hard-fork-support.ts` — the whole API in ~180 lines.
2. `packages/abstractions` (`ProtocolVersion`, `WalletTransaction`) and `packages/hd` (`WalletSeeds`).
3. One wallet package end to end, then the other two.
4. `packages/facade` and `packages/wallet-sdk` (what applications see), then the changesets.

## Open points for reviewers

- Two `minor` bumps change constructor signatures of `ProtocolState` and `VariantContext`. Rationale: consumers read these objects, they don't construct them. Flagged for a deliberate call.
- The spec calls HD role 4 "Metadata"; the released `packages/hd` uses it as `Roles.EcdsaUnshielded`. The new spec section leaves role 4 unnamed — for the spec owner to resolve.
- API ergonomics noticed while writing the snippet (none blocking): `provingServers` cannot distinguish "pre-fork proving intentionally unavailable" from a missing entry; there is no facade-level `tryRestore`; the facade and per-wallet key-object shapes nest differently; filtering pending transactions by status doesn't narrow the type; error messages print `MaxSupportedVersion` as a raw number.


packages/capabilities/src/signatures/signing.ts
```diff
@@ -0,0 +1,43 @@
+// This file is part of MIDNIGHT-WALLET-SDK.
```
kapke, line 1: This feels more like "capabilities" than "abstractions".

packages/abstractions/src/WalletTransaction.ts
```diff
@@ -0,0 +1,337 @@
...
+/**
+ * How far along the building of a transaction a handle is.
+ *
+ * @remarks
+ *   The three points at which a transaction changes hands between an application and the SDK: built but not proved,
+ *   proved but not yet bound to its own contents, and finalized — proved, signed and bound, which is the only shape the
+ *   network takes. Both ledger versions have all three, and the words mean the same thing on either side of a protocol
+ *   boundary, which is why the stage can be plain data on a version-agnostic handle.
+ */
+export type TransactionStage = 'Unproven' | 'Unbound' | 'Finalized';
```
kapke: This type should come from the schema to prevent accidental decoupling.

packages/abstractions/src/WalletTransaction.ts
```diff
@@ -0,0 +1,337 @@
...
+   *   about which ledger version produced the transaction, and the SDK holds the caller to it: a handle stamped for one
+   *   side of a protocol boundary is refused by a wallet acting on the other side, with a
+   *   {@link ProtocolVersionMismatchError}. Until the fork, that means a transaction built with the post-fork ledger
+   *   version is refused at run time, however well it type-checks.
+   * @param stage How far along the building of the transaction is.
+   * @param transaction The transaction, of whichever ledger version built it.
+   * @param protocolVersion The protocol version that ledger version serves.
+   * @returns The sealed handle.
+   */
+  static adopt<TStage extends TransactionStage>(
```
kapke, line 180: Claude highlighted it's quite an insecure API because the caller provides version information.

I have another concern though - if the underlying transaction type is `Serializable`, how it can assert anything useful to the caller? Shouldn't it adopt similar API style for construction and operating on it to the runtime, so that there are disptatchers for different branches?

packages/dust-wallet/src/ForkingDustWallet.ts
```diff
@@ -0,0 +1,1028 @@
...
+    /**
+     * Starts background synchronization on whichever variant is current, and keeps the key for the next one.
+     *
+     * @remarks
+     *   The key is the post-fork ledger version's, so it is retained against that variant alone. A wallet still on the
+     *   pre-fork variant is started from what it retained instead — which a wallet built from a seed can always answer,
+     *   and a wallet built from a key object cannot.
+     * @param secretKey The post-fork ledger version's dust secret key.
+     */
+    start(secretKey: ledger.DustSecretKey): Promise<void> {
```
kapke, line 665: Shouldn't it take keys for both branches instead?

This likely also applies to other wallets.

packages/dust-wallet/src/ForkingDustWallet.ts
```diff
@@ -0,0 +1,1028 @@
...
+
+    /**
+     * Runs one synchronization step on whichever variant is current.
+     *
+     * @remarks
+     *   Available on both sides of the boundary: the pre-fork variant is stepped with a key derived from what the wallet
+     *   retained, exactly as background synchronization is.
+     * @param secretKey The post-fork ledger version's dust secret key.
+     */
+    stepSync(secretKey: ledger.DustSecretKey): Promise<void> {
```
kapke, line 704: ditto

packages/facade/src/index.ts
```diff
@@ -1343,51 +1783,74 @@ export class WalletFacade {
...
       this.dust.revertTransaction(tx),
     ]).then(async () => {
-      await this.pendingTransactionsService.clear(tx as unknown as ledger.FinalizedTransaction);
-      const key = revertTxHistoryKey(tx);
+      // Reverting is total over the stages: a transaction at any of them may have booked coins, and the pending set
+      // recognises only the finalized ones. Narrowing rather than casting is what says so.
+      await this.pendingTransactionsService.clear(tx as FinalizedTx);
+      // A transaction of the other epoch has no history entry of this session's to land on, so there is nothing to
+      // key and nothing to record — the same reason the wallets treat it as nothing of theirs to release.
+      const key = Option.match(Either.getRight(WalletTransaction.unwrapWithin<Carried>(tx, this.currentEpoch())), {
```
kapke: It can't only be current epoch checked here. Tx submitted _before_ the fork, if failing, may report so _after_ the fork.

kapke (commented):
Tremendously important to ensure - despite all the new wrappings, restoration of existing v1 wallet should just work

packages/dust-wallet/src/ForkingDustWallet.ts
```diff
@@ -0,0 +1,1028 @@
...
+   * @remarks
+   *   Short, because what is being bought is small: the alternative to an answer is the hand-over this wallet has always
+   *   done, which costs one migration and no correctness on a chain that produces events. It is a ceiling rather than a
+   *   typical cost — an unreachable indexer refuses a connection long before this — and it exists so that a probe which
+   *   neither answers nor fails cannot hold a start open indefinitely.
+   */
+  const probeTimeout = Duration.seconds(5);
+
+  /** Where a start begins, when the chain answered for it. */
+  type ProbedStart = Readonly<{
```
kapke, line 437: An idea from Claude, in line with #657 - before commanding the version switch, confirm it with the node first.

packages/dust-wallet/src/ForkingDustWallet.ts
kapke, line 1: I don't like the naming. It should stay "DustWallet", whereas the "DustWallet" should indicate it's a wrapper for a single variant.

packages/capabilities/src/signatures/signing.ts
agronmurtezi: Agreed and moved in 3b08ddea: it now lives at `capabilities/src/signatures/signing.ts`, beside `preForkSignatures.ts` — its only consumer, the lifting/lowering between the two ledgers' signature shapes. Nothing inside `abstractions` used it beyond the barrel line. The public surface is unchanged: `import { Signing } from '@midnightntwrk/wallet-sdk'` resolves as before (now pinned by a root-surface test rather than incidental), and the release note names the new home. No compat re-export from abstractions — that would have inverted the dependency.

packages/abstractions/src/WalletTransaction.ts
agronmurtezi: Done in 20857efd: `TransactionStage` is now `Schema.Schema.Type<typeof TransactionStageSchema>` — the schema leads and the type is read off it, so the envelope and the type cannot drift (the old annotation pinned the schema to the type but not the reverse). The refusal of an unknown stage was already pinned by `walletTransaction.test.ts` ("refuses an envelope naming a stage that is not one").

packages/abstractions/src/WalletTransaction.ts
agronmurtezi: You're right that `adopt` itself checks nothing — that is the recorded design decision (authors are runtime-check-only) rather than an oversight, and here is what actually stands behind it. A handle whose *stamp* is on the wrong side is refused at every entry point (`accept`, validation, each wallet's build/submit paths — all via `unwrapWithin` → `ProtocolVersionMismatchError`). A handle whose stamp *lies* passes those range checks and then fails loudly in the ledger itself, in both directions: deserialize rejects on the versioned header tag (`expected header tag 'midnight:transaction[v9](…)'`), and any cross-ledger method call fails wasm-bindgen's instance check (`expected instance of Transaction`) — measured both on the shipped ledgers. Nothing silently misreads; the handle is routing data, not a security boundary.

On hardening `adopt` with a round-trip deserialize under the claimed version: measured at ~0.06 ms per call as the floor (scales with transaction size). Of the ~60 non-test `adopt` sites, nearly all are the SDK's own wallets stamping transactions they just built with a literal next to the variant that built them — the check would cost the most where it can never fire, including `seal` on every facade-built transaction and `deserialize`, which would re-parse what it just parsed. A dispatcher/registry construction is the stronger shape in principle, but `abstractions` deliberately imports no ledger, so the entries would have to be injected — which is exactly what `fromWire`'s caller-supplied `decode` already is. Happy to add the round-trip check behind the door where lies can actually enter (`adopt` called from application code) if you think it's worth the parse — say the word.

packages/dust-wallet/src/ForkingDustWallet.ts
agronmurtezi: The both-keys form exists and populates both: `startWithKeys(DustKeysByEpoch)` retains pre- and post-fork keys, and seeds derive both. The single-key `start(secretKey)` retaining only the post-fork slot is documented as deliberate in four places (both wallets' files, `StartMaterial.ts`, and the release note): a lone key object belongs to one ledger version's runtime, so only a seed or the two-key form can cross a boundary.

Your instinct found a real hole one door over, though: a wallet **restored from a pre-fork snapshot** retains no keys at all, and `start(v9Key)` on it *rejects* with `MissingStartAuxError` — so a direct `DustWallet(...).restore` caller (the facade exposes no restore) cannot bring a restored wallet across without re-starting from a seed. Also true for shielded (unshielded is watch-only and exempt), and `startWithKeys` currently has zero test coverage. We've logged it as a design decision to make — widen `start`, add an instance-level `startWithKeys`, or take keys at `restore` (where the material is actually missing) — rather than patching it inside this review.

packages/dust-wallet/src/ForkingDustWallet.ts
agronmurtezi: Same answer as the thread just above (at the `#retainKeys` write): the two-key form is `startWithKeys`, the single-key `start` is deliberately post-fork-only — and the genuinely missing piece is keys on the restore path, logged as a design decision.

packages/facade/src/index.ts
agronmurtezi: Confirmed real, and fixed in 8720a653 — thank you, this one was subtle. Orphaning did not cover it: the facade advances its observed version *before* `orphanBeyond` marks the item, so at the moment a verdict lands the current epoch has already moved, `unwrapWithin` returns Left, and the `gotRejected` history write was silently skipped — the submission's own pending entry stayed `pending` forever. The revert path now reads the history key at the transaction's **own** epoch (`epochOf(tx.protocolVersion)`), not the facade's current one. A red-first suite (`lateVerdictAcrossFork.test.ts`) drives all three wallets across the boundary and pins both: a chain-reported failure arriving after the crossing lands on the entry the submission wrote, and an orphaned transaction's `orphaned-by-protocol-upgrade` reason is recorded once the wallets have crossed. (The pending-set clear itself was never broken — only the history write.)

packages/dust-wallet/src/ForkingDustWallet.ts
agronmurtezi: Worth keeping on the table as defense in depth. Two properties shipped since this comment that narrow the risk: the recorded version is monotonic (a reconnect replaying older history cannot drag a wallet back across a boundary), and the version watcher (#702) that triggers a switch on a quiet chain adopts a version only after confirming, against the indexer, that the wallet is provably caught up on event ids. A node-side second opinion before migrating would guard against a misbehaving indexer specifically — parked in the follow-ups plan alongside the #657 line of work rather than wiring a second data source into the hand-over now.

packages/dust-wallet/src/ForkingDustWallet.ts
agronmurtezi: The public names already carry the semantics you're asking for: applications construct `DustWallet(config)` (the shipped both-variants wallet), `CustomDustWallet(config, builder)` (explicitly the single-variant composition — `dust-fast-sync.ts` documents the contrast in exactly those words), and `CustomForkingDustWallet` (the general two-variant form). There is no exported `ForkingDustWallet` binding — it is a file name and the running-instance *type*; `type DustWallet` and `function DustWallet` merge in the public module, and "Forking" appears in zero docs, changesets or snippets. So the rename would be file/type-internal. If those names grate, the contained fix is renaming the *files and instance types* (never the public factories) — we'd rather do that as a mechanical follow-up across all three wallets than mid-review. The missing doc comments on `CustomDustWallet`/`CustomUnshieldedWallet` (only the shielded one documents its single-variant contract) we'll add either way.

agronmurtezi:
> Tremendously important to ensure - despite all the new wrappings, restoration of existing v1 wallet should just work

It does, and it is tested. `restore`/`tryRestore` route on the protocol version the snapshot itself declares (`Restore.ts` → `variantForSnapshot`) — the chain probe is never involved. Serialized states round-trip unchanged on this line (the wallets changeset states it, and the serialization suites hold round-trip tests per variant), and `forkRestore.test.ts` covers a pre-fork snapshot restoring onto the pre-fork variant and syncing on. The one deliberate addition on `./v1`: deserialization now rejects a snapshot whose address does not derive from its verifying key — snapshots written by released wallets are unaffected (called out in the changeset).

agronmurtezi:

PR #689, opened 2026-08-21
feat(runtime,capabilities): hard-fork test harness and real v8-to-v9 state translation

agronmurtezi:
## What's in this PR

The groundwork for hard-fork support: runtime changes, test infrastructure, and the state translation the wallet PRs build on. Wallet behavior does not change yet.

### Variant activation in the runtime

The wallet runtime hosts one wallet implementation ("variant") per protocol-version range. It now derives each variant's range from the registered variants themselves, and notifies a variant when it becomes the active one — the mechanism a wallet uses to switch implementations when the chain changes protocol version.

### Both ledgers in one process

`@midnight-ntwrk/ledger-v8` and `@midnightntwrk/ledger-v9` are separate WASM modules, and hard-fork support requires loading both at once. A test proves they load and run side by side in the same process.

### Fork simulation for tests

The in-memory chain simulator used by tests now exists for both ledger versions, and a new fork simulator runs a chain on the v8 ledger up to a chosen protocol version, then continues it on the v9 ledger. Later PRs use it to test wallets crossing a fork end to end.

### Real v8-to-v9 state translation

A new package, `@midnightntwrk/wallet-sdk-state-translation`: a Rust crate compiled to WebAssembly that wraps the ledger's own state-translation code, plus the TypeScript loader that calls it. The fork simulator feeds the v8 chain's final state through it, so a simulated post-fork chain starts from the translated pre-fork state — the same thing a real fork does — rather than from an empty ledger.

Integration tests check that the v9 ledger accepts a translated state, that every coin commitment accumulated on the v8 chain is still present afterwards, and that a v9 transaction applies on top of the result.

This is the first Rust in the repository. A CI job builds the WASM artifact, verifies it by translating a real state, and passes it to the integration tests. Local builds need rustup, wasm-bindgen 0.2.104 and binaryen — see `packages/state-translation/wasm/README.md`. Nothing else in the repo needs Rust; regular builds and unit tests are unaffected.

## How to review

- `packages/state-translation` — the crate, the loader and their tests are the bulk of the diff.
- `packages/capabilities/src/simulation` — the per-version simulators and the fork simulator.
- `packages/runtime` — the activation changes.


packages/capabilities/turbo.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": ["//"],
+  "tasks": {
+    "test:integration": {
+      // forkStateTranslation.integration.test.ts runs the real v8-to-v9 translation, which is a WASM artifact built from
+      // packages/state-translation/wasm. Declared so turbo builds it first rather than leaving the ordering to CI —
+      // which is why running any integration test in this package needs a Rust toolchain.
+      "dependsOn": ["$TURBO_EXTENDS$", "@midnightntwrk/wallet-sdk-state-translation#build:wasm"]
```
kapke: It feels like a wrong place to put the translation. I would expect the translation's build to be set up in a way, that it requires the own wasm build instead, to not expose build details of a package with others.

packages/state-translation/turbo.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": ["//"],
+  "tasks": {
+    "test:integration": {
```
kapke, line 11: Why not own build requires build:wasm? That would make it behave much, much more predictably.

packages/state-translation/turbo.json
```diff
@@ -0,0 +1,9 @@
...
+      // These import the built WASM statically, so the artifact has to exist before they run.
+      "dependsOn": ["$TURBO_EXTENDS$", "build:wasm"]
```
kapke: Shouldn't test also require verify:wasm for completeness?

packages/capabilities/turbo.json
agronmurtezi: Fixed in 6a2e4223: capabilities no longer names another package's build task. It now depends on state-translation's own `artifacts` gate, which owns both `build:wasm` and `verify:wasm` — so the build details live with the package that builds them.

packages/state-translation/turbo.json
agronmurtezi: Deliberate, and now written down in the file itself: the repo invariant is that `dist`, `typecheck`, `lint` and `test:unit` need no Rust (CLAUDE.md, Rust toolchain section) — the WASM is loaded only by the integration tier, so hanging it on `build` would put cargo in front of every `yarn dist` for all 22 packages. The committed `.d.ts` is what lets dist typecheck against an artifact it never builds. The predictability you're after is addressed one level up instead: the artifact tasks are now owned by this package behind a single `artifacts` gate (6a2e4223), so consumers depend on one stable, package-local name.

packages/state-translation/turbo.json
agronmurtezi: Yes — done in 6a2e4223. `test:integration` now routes through the `artifacts` gate, which depends on both `build:wasm` and `verify:wasm`, so completeness holds by construction rather than by a second list that could drift. One discovery while wiring it: `verify:wasm` was never declared in the root turbo.json at all (a `dependsOn` naming it would not have built a graph) — it is now, with its real edges (`build:wasm` and `dist`, since the verify script imports the built loader).

agronmurtezi:

PR #690, opened 2026-08-21
feat(shielded,dust,unshielded): dual-ledger variants and hard-fork migration for all three wallets

agronmurtezi:
## What's in this PR

Each wallet package — shielded, dust, unshielded — now contains two implementations of the same wallet, one per ledger version, and can follow the chain across a hard fork.

### The two variants

- `src/v2` is the ledger-v9 variant: the existing production code, renamed from `src/v1` (`V1Builder` → `V2Builder`, and so on).
- `src/v1` is the ledger-v8 variant: the pre-fork wallet code restored from main, importing `@midnight-ntwrk/ledger-v8`.

The two directories are intentionally near-identical apart from the ledger they import. Anything importing the packages' `/v1` subpaths now gets the v8 variant, with the v9 variant at `/v2`; the root entry points are unchanged.

### Crossing the fork

Each wallet watches its sync stream for the first update at or past the fork's protocol version. From that point it applies nothing further, and the runtime moves the wallet onto the other variant:

- **Shielded and dust** rebuild their state on the new ledger. Private state cannot be copied between ledger versions, so the new variant starts fresh and re-discovers the wallet's coins from the chain.
- **Unshielded** state is public UTXO data, so it is copied over field by field and syncing resumes from the same position.

### The fork tests

For each wallet, a test drives a simulated chain across a fork — with transactions before, at and after the boundary — and verifies the wallet ends up on the new variant with its funds and history intact. Each suite includes negative tests (dropping the carried state, moving the sync position, applying the boundary transaction on the old side) confirming a broken crossing would be caught.

## How to review

1. `packages/shielded-wallet` in full — the other two wallets follow the same structure.
2. `packages/dust-wallet` and `packages/unshielded-wallet` — the differences are in how state crosses: rebuilt versus copied.
3. The `src/v1` directories are restored v8 code with a few small, documented edits — spot-checking them is enough.


packages/shielded-wallet/src/v1/Sync.ts
kapke, line 338: I think the payload's protocol version should be inspected first to asses whether the events should be parsed here or deferred to the next variant. This likely also matters to Dust. Less for unshielded one.

agronmurtezi: Implemented later in this stack (it lands in #688's slice). Events now travel undecoded — `EventsSyncUpdate` carries the raw hex plus the version the source reported, and its doc comment states the rule: the bytes are only read by the capability that knows the activation range, on the slice it is about to apply. `splitAtVersionBoundary` (shielded `src/v{1,2}/Sync.ts`) is the single place the boundary rule lives; everything from the first out-of-range event onwards is deferred for the next variant to re-fetch, and `applyUpdate(state, update, activeRange)` carries the range. Dust got the same treatment (its own `splitAtVersionBoundary` in `dust-wallet/src/v2/Sync.ts`); unshielded's payloads are version-independent JSON and annotate the version they were reported at. The fork suite proves the deferral — "hands over on the first batch, having applied nothing" in `forkStart.test.ts`.

agronmurtezi:

PR #698, opened 2026-08-28
fix(unshielded-wallet)!: stop a leaked booking duplicating a utxo and doubling the balance

### LINKED ISSUE #697: A leaked unshielded booking gets duplicated into both UTXO maps, and the balance doubles
# A leaked unshielded booking gets duplicated into both UTXO maps, and the balance doubles

Target: `@midnightntwrk/wallet-sdk-unshielded-wallet` (`packages/unshielded-wallet`), with a contributing gap in `@midnightntwrk/wallet-sdk-facade` (`packages/facade`).

Tested against `wallet-sdk-unshielded-wallet@3.1.0`, `wallet-sdk-facade@4.1.0`. Source line numbers are from `main` at `6e1050e`, where both are unchanged.

## Symptom

A wallet holding only unshielded NIGHT reports **exactly twice** its on-chain balance (6,000 displayed against 3,000 on chain) and keeps doing so across restarts. DUST registration then reports success without doing anything, and the register and deregister views both show "none available". No request reaches the proof server.

The persisted state has the same UTXO in both maps at once:

```json
"availableUtxos": [ { "intentHash": "421c4146…", "outputNo": 0, "value": "3000000000" } ],
"pendingUtxos":   [ { "intentHash": "421c4146…", "outputNo": 0, "value": "3000000000" } ]
```

Deleting the persisted state is the only recovery.

## The invariant

`availableUtxos` and `pendingUtxos` are disjoint by construction. `spend()` moves a UTXO between them — `unshielded-wallet/src/v1/UnshieldedState.ts:55-66`:

```ts
spend: (state, utxo) =>
  Either.gen(function* () {
    const hash = UtxoHash(utxo.utxo);
    if (!HashMap.has(state.availableUtxos, hash)) {
      return yield* Either.left(new UtxoNotFoundError({ utxo: utxo.utxo }));
    }
    return {
      availableUtxos: HashMap.remove(state.availableUtxos, hash),   // removed here
      pendingUtxos: HashMap.set(state.pendingUtxos, hash, utxo),    // added here
    };
  }),
```

Every balance accessor depends on that. `CoinsAndBalances.ts:41-67` sums each map independently (`getAvailableCoins`/`getPendingCoins` return the raw `HashMap` values — no subtraction), and `getTotalBalances` adds the two. `UnshieldedWallet` exposes the available side as its headline number, `unshielded-wallet/src/UnshieldedWallet.ts:68-70`:

```ts
get balances(): Record<ledger.RawTokenType, bigint> {
  return this.capabilities.coinsAndBalances.getAvailableBalances(this.state);
}
```

So a UTXO in both maps is counted twice by `getTotalBalances`, and is reported as spendable by `balances` while also being reported as booked — inside the SDK, before any consumer code runs.

## Defect 1 — `applyUpdate` re-adds a UTXO that is currently booked

`UnshieldedState.ts:98-120` unions the update's `createdUtxos` into `availableUtxos` with no check against `pendingUtxos`:

```ts
return {
  availableUtxos: HashMap.union(
    HashMap.removeMany(state.availableUtxos, update.spentUtxos.map((utxo) => UtxoHash(utxo.utxo))),
    HashMap.fromIterable(update.createdUtxos.map((utxo) => [UtxoHash(utxo.utxo), utxo])),
  ),
  pendingUtxos: HashMap.removeMany(state.pendingUtxos, update.spentUtxos.map((utxo) => UtxoHash(utxo.utxo))),
};
```

The indexer-backed sync capability (`Sync.ts:103-135`) feeds `createdUtxos` straight from the `unshieldedTransactions` subscription, so any replay of the transaction that created a currently-booked UTXO puts that UTXO back into `availableUtxos` while it is still in `pendingUtxos`.

**The SDK already has this guard elsewhere.** The simulator sync capability does exactly the check the indexer path omits, `Sync.ts:208-212`:

```ts
// Created: in simulator but not in wallet (neither available nor pending)
const createdUtxos = Array.from(simulatorUtxoMap)
  .filter(
    ([hash]) => !HashMap.has(state.state.availableUtxos, hash) && !HashMap.has(state.state.pendingUtxos, hash),
  )
```

`restore()` (`:50-53`) does not check disjointness either, so once written the duplicate survives every restart and no amount of resyncing clears it.

## Defect 2 — a booking is only released through the submit path

Inputs are booked during **balancing**, not submission — `Transacting.ts:346` and `:558` both call `CoreWallet.spendUtxos(...)` while building the offer. Release happens in exactly three places, all downstream of submission:

1. `facade.submitTransaction` reverts in its catch — `facade/src/index.ts:772-787`:

```ts
try {
  await this.pendingTransactionsService.addPendingTransaction(tx);
  const key = submitTxHistoryKey(tx);
  await this.#txHistoryStorage.gotPending({ ...key, submittedAt: this.clock.now() });
  await this.submissionService.submitTransaction(tx, 'Finalized');
  return identifiers.at(-1)!;
} catch (error) {
  await this.revert(tx);
  throw error;
}
```

2. The pending-transaction watcher reverts failures — `facade/src/index.ts:565-571`:

```ts
this.#pendingSubscription = this.pendingTransactionsService
  .state()
  .pipe(
    concatMap((pending) => PendingTransactions.allFailed(pending)),
    concatMap((item) => this.revert(item.tx)),
  )
  .subscribe();
```

3. `applyFailedUpdate` (`UnshieldedState.ts:122-141`) returns inputs on an on-chain `FAILURE`.

All three require the transaction to have reached submission, and (2) and (3) require an on-chain result: `allFailed` selects only items whose `result.status` is `FAILURE` or `PARTIAL_SUCCESS` (`capabilities/src/pendingTransactions/pendingTransactions.ts:67-74`). Nothing reaps a booking on a TTL or reconciles pending entries against chain state.

That leaves two windows where a booking leaks permanently:

- **Booked but never submitted.** Balancing succeeds, then proving fails — in our case a proof server running a mismatched ledger version — or the process dies. No pending entry was ever created, so no revert path can see it.
- **Submitted but never resolved.** The transaction is accepted for submission but never lands and never gets a result. Nothing expires it.

A caller *can* call `facade.revert(recipe)` itself, and that is the documented escape (`facade/src/index.ts:1345-1352`). But nothing in the SDK's surface signals that a booking is outstanding, so a caller that misses one — or a process that dies before it can — has no way back: the booking is unspendable and, via Defect 1, becomes a double-count on the next resync.

## How the two combine

1. UTXO `U` is in `availableUtxos`.
2. Balancing books it: `spend()` moves it to `pendingUtxos`. The balance is still correct — `getAvailableBalances` no longer counts it.
3. Proving fails. Nothing reverts (Defect 2), so `U` stays pending forever.
4. A later sync replays the transaction that *created* `U` — a restore from a cursor predating it, or a resubscription from an earlier `appliedId` — and Defect 1 re-adds `U` to `availableUtxos`.
5. `U` is now in both maps. `getAvailableBalances` counts it and `getPendingBalances` counts it, so `getTotalBalances` doubles. Consumers that add pending to available (to avoid flashing the balance to zero mid-booking) double it too. Coin selection that excludes booked UTXOs still finds nothing to spend — so the wallet simultaneously reports twice the money and nothing available.

## Reproduction

1. Fund a fresh wallet with unshielded NIGHT only — no shielded history keeps the failure isolated to this state machine.
2. Point it at a proof server running a different ledger version than the node, so proving fails deterministically.
3. Build a transaction that books an input (a transfer, or a DUST registration) and let proving fail. Do not call `facade.revert`.
4. Restart so the wallet restores from persisted state and resyncs from its cached cursor.
5. The persisted state now has one `intentHash`+`outputNo` in both `availableUtxos` and `pendingUtxos`, and the reported balance is doubled.

## Suggested fixes

**Defect 1** — give the indexer path the guard the simulator path already has:

```diff
       return {
+        // A created UTXO that is currently booked must not re-enter availableUtxos:
+        // the maps are disjoint by construction (see `spend`) and every balance
+        // accessor counts them independently.
         availableUtxos: HashMap.union(
           HashMap.removeMany(
             state.availableUtxos,
             update.spentUtxos.map((utxo) => UtxoHash(utxo.utxo)),
           ),
-          HashMap.fromIterable(update.createdUtxos.map((utxo) => [UtxoHash(utxo.utxo), utxo])),
+          HashMap.fromIterable(
+            update.createdUtxos
+              .filter((utxo) => !HashMap.has(state.pendingUtxos, UtxoHash(utxo.utxo)))
+              .map((utxo) => [UtxoHash(utxo.utxo), utxo]),
+          ),
         ),
```

`restore()` deserves the same treatment: dropping an entry present in both maps repairs already-persisted state on load instead of carrying it forward forever. That alone would have turned this from "unrecoverable without deleting the cache" into "self-heals on next start".

**Defect 2** — in increasing order of completeness:

1. Reconcile on sync: a pending UTXO that the indexer still reports as unspent, whose booking has no corresponding pending transaction, was never spent. This is the case Defect 1's guard cannot fix, because there is nothing to guard against — the booking itself is stale.
2. Expire bookings on a TTL, the way transactions already have one. A booking outliving its transaction's TTL cannot still be valid.
3. Surface outstanding bookings on the public API, so a consumer can at least detect and release them. Today `pendingCoins` shows the coins but not which transaction booked them or whether that transaction still exists.

## Note for consumers

Two things bit us that are worth stating plainly for anyone else on this API:

- Bookings are taken at **balance** time, so any failure between balancing and submission must be followed by an explicit `facade.revert(recipe)`. We had this right in our fee-estimation path (which reverts in a `finally`) and wrong in the transfer and registration paths, which is how we produced the leak.
- `balances` is the **available** map only, so a consumer that wants "what this wallet owns" has to add `pendingCoins` back — and that addition is what turns a duplicated UTXO into a visibly doubled balance. A consumer doing this should skip any pending UTXO whose `intentHash#outputNo` also appears in `availableCoins`, and warn: the duplicate is trivially detectable at read time, and would have saved us a two-hour investigation.

iangregsondev:
Related issue: #697

# Description

A wallet holding only unshielded Night reported exactly twice its on-chain balance, kept doing so across restarts, and simultaneously reported nothing available to spend. Deleting the persisted state was the only recovery.

Two defects combined. The indexer sync path re-admitted a created coin without checking whether it was still booked, so a resync from a cursor predating the creating transaction put that coin into `availableUtxos` while it was also in `pendingUtxos`. And a booking taken during balancing was released only through the submit path, so a transaction abandoned before submission — a proof server on a mismatched ledger version, in the reported case — left its coins booked forever. Both maps are persisted, so the corruption survived every restart and no amount of resyncing cleared it.

# Proposed Solution

ADR 0008 in this PR records the decision and the options weighed. The short version: a booking is not the caller's private intent, it is a record of something the wallet has already released into the world, so it must survive a restart — and therefore needs something that can eventually retire it.

Four mechanisms, smallest first, each useful without the ones after it.

The invariant is enforced where it broke. `applyUpdate` no longer admits a created coin that is still booked, the guard the simulator path already had, and loading a snapshot holding one coin in both maps keeps it on the pending side only, so state already corrupted in the field repairs itself on the next start.

A booked coin carries the TTL of the transaction it was booked for, and both sync capabilities release expired bookings as updates arrive. This is the only thing that releases a booking, so nothing else can free a coin some other transaction has since booked. It sweeps only once sync has caught up with the chain, because a replay from an earlier cursor can still be carrying the transaction that spent a booked coin, and it releases strictly after the TTL rather than at it, because the ledger accepts an intent while its TTL is at or after the block's timestamp. This bounds any abandoned transaction to its own lifetime with no help from the caller, and works for a wallet used without the facade.

A transaction that has been balanced but not yet proven is recorded as a reservation in the pending-transactions service: its identifiers, the ids of the coins it booked, and its TTL. Never the transaction itself, which carries key material. In-place balancing records only the coins the wallet moved onto the pending side, since the transaction it hands back may already carry inputs the caller put there. The record persists, so a booking held for a counterparty that has not answered yet survives a restart. Registering or clearing the real transaction drops the reservation standing in for it, and so does reverting the booking on any failure path.

Bookings restored from a snapshot are reconciled once sync reaches the chain tip: every transaction the address is party to has been applied by then, so a coin still booked was never spent — unless a reservation or a transaction being tracked says the spend is still out there. Both halves matter, because registering a transaction drops the reservation that stood for it, so from submission onwards the tracked transaction is the only thing holding its coins. Uncovered coins come back in seconds rather than waiting out a TTL that defaults to an hour.

Separately, and in the same class as the reported defect: balancing a transaction in place discarded the state coin selection produced, so it never booked the coins it selected. Two balance calls could hand the same coin to two transactions. It now books them like every other path.

Breaking, unshielded wallet: `UnshieldedState.spend`, `UnshieldedState.spendByUtxo`, `CoreWallet.spend` and `CoreWallet.spendUtxos` take the transaction's TTL as a required last argument. `UnshieldedState.restore` and `toArrays` exchange pending entries as `{ utxo, ttl, restored }` rather than bare UTxOs. `TransactingCapability` gains `revertUtxos` and `releaseRestoredPending`, so a custom implementation must add them. The public coin accessors are unchanged: `pendingCoins`, `availableCoins` and `balances` still report UTxOs.

New on the unshielded wallet: `revertUtxos(ids)` releases booked coins without the transaction that booked them, for a caller holding only a record of the ids.

Stored-format compatibility, worth a reviewer's attention. Both formats gained an optional member rather than a new version, so a store written by this release is still readable by an earlier one, which simply does not see the new member. The unshielded snapshot carries each booking's expiry; a snapshot written before expiries existed says nothing about when its coins were booked, so each booking is granted a full transaction lifetime from the moment it is loaded rather than being treated as already expired. The pending-transactions store carries its reservations the same way.

The commits building this are ordered so each test commit precedes the implementation it specifies. The last seven are the exception: they carry one round of review fixes, grouped by theme with their tests alongside, because those changes reach across the same files.

# Testing

Unit tests cover the state transitions, the snapshot round trip in both directions, both sync paths, the reservation collection and its persistence, the service poll, the facade registering and dropping reservations, and in-place balancing booking what it selects. Integration tests run against a real node and indexer under Docker, starting from a hand-written snapshot so the coins under test are ones the indexer never mentions: a snapshot holding one coin in both maps loads disjoint and stays disjoint on every emitted state, a booking whose transaction can no longer be accepted is released as updates arrive, and a restored booking something still accounts for stays booked until nothing does.

One path has no automated test, and I would rather say so than imply otherwise: the facade's trigger that calls the reconciliation when sync reaches the tip. The simulator never reports a chain tip, so the condition cannot be reached in that harness. The wallet behaviour behind it is covered against the real indexer; the subscription that calls it is not.

Worth a reviewer's eye: the sentence the reconciliation turns on, since all three of its simpler forms are wrong. Releasing restored bookings on sync completion unguarded would invalidate a swap whose counterpart has not answered; never releasing them leaves a coin stuck for up to an hour after sync has already proved nothing holds it; and guarding on reservations alone frees the coins of a transaction that was submitted and is still in the mempool, because registering it dropped its reservation. Related to that last point, the facade's default pending-transactions service starts empty, so a consumer that restores the wallet snapshot but not the pending store reaches the tip with nothing accounting for the spends its previous process submitted. That is stated in ADR 0008 rather than worked around here.

### DIFF
--- a/.changeset/fix-unshielded-utxo-duplication.md
+++ b/.changeset/fix-unshielded-utxo-duplication.md
@@ -1,0 +1,36 @@
+---
+'@midnightntwrk/wallet-sdk-unshielded-wallet': major
+'@midnightntwrk/wallet-sdk-capabilities': minor
+'@midnightntwrk/wallet-sdk-facade': minor
+---
+
+fix(unshielded-wallet): stop a leaked booking duplicating a UTxO and doubling the balance
+
+`availableUtxos` and `pendingUtxos` are disjoint by construction, and every balance accessor sums them independently, so
+a UTxO present in both is counted twice. Two defects combined to produce exactly that: the indexer sync path re-admitted
+a created UTxO that was still booked, and nothing released a booking taken while balancing when the transaction was
+abandoned before submission. Persisted, the result survived every restart.
+
+- `applyUpdate` no longer admits a created UTxO that is currently booked, the guard the simulator path already had, and
+  loading a snapshot holding one coin in both maps keeps it on the pending side only. State already corrupted in the
+  field repairs itself on the next start.
+- A booked coin now carries the TTL of the transaction it was booked for, and both sync capabilities release expired
+  bookings on every applied update. Past that instant the ledger rejects the transaction, so the booking cannot still be
+  valid. A pending entry restored from a snapshot written before this change is released by the first sweep.
+- A transaction that has been balanced but not yet proven is recorded as a reservation in the pending-transactions
+  service: its identifiers, the ids of the coins it booked, and its TTL, never the transaction itself, which carries key
+  material. The record persists, so a booking held for a counterparty that has not answered yet survives a restart. The
+  service's existing poll marks a reservation whose TTL has passed, and the facade then releases its coins.
+- Bookings restored from a snapshot are released once sync reaches the chain tip, unless a reservation still accounts
+  for them, returning an abandoned coin in seconds rather than at its transaction's TTL.
+- `UnshieldedWallet.revertUtxos(ids)` releases booked coins without the transaction that booked them, for a caller
+  holding only a record of the ids.
+
+BREAKING CHANGE: `UnshieldedState.spend`, `UnshieldedState.spendByUtxo`, `CoreWallet.spend` and `CoreWallet.spendUtxos`
+take the TTL of the transaction the coins are booked for as a required last argument. `UnshieldedState.restore` and
+`toArrays` exchange pending entries as `{ utxo, ttl, restored }` rather than bare UTxOs, and `TransactingCapability`
+gains `revertUtxos` and `releaseRestoredPending`, so a custom implementation of that interface must add them. The public
+coin accessors are unchanged: `pendingCoins`, `availableCoins` and `balances` still report UTxOs.
+
+The pending-transactions store gains a new format version carrying reservations. A store written before this change
+still loads; one written after it cannot be read by an earlier version of `@midnightntwrk/wallet-sdk-capabilities`.
--- a/.changeset/fix-unshielded-utxo-duplication.md
+++ b/.changeset/fix-unshielded-utxo-duplication.md
@@ -15,14 +15,19 @@ abandoned before submission. Persisted, the result survived every restart.
   loading a snapshot holding one coin in both maps keeps it on the pending side only. State already corrupted in the
   field repairs itself on the next start.
 - A booked coin now carries the TTL of the transaction it was booked for, and both sync capabilities release expired
-  bookings on every applied update. Past that instant the ledger rejects the transaction, so the booking cannot still be
-  valid. A pending entry restored from a snapshot written before this change is released by the first sweep.
+  bookings as updates arrive. This is the only thing that releases a booking, so two sweeps on different clocks cannot
+  free the same coin twice. It runs only once sync has caught up with the chain, because a replay from an earlier cursor
+  can still be carrying the transaction that spent a booked coin, and it releases strictly after the TTL rather than at
+  it, because the ledger accepts an intent while its TTL is at or after the block's timestamp.
 - A transaction that has been balanced but not yet proven is recorded as a reservation in the pending-transactions
   service: its identifiers, the ids of the coins it booked, and its TTL, never the transaction itself, which carries key
-  material. The record persists, so a booking held for a counterparty that has not answered yet survives a restart. The
-  service's existing poll marks a reservation whose TTL has passed, and the facade then releases its coins.
-- Bookings restored from a snapshot are released once sync reaches the chain tip, unless a reservation still accounts
-  for them, returning an abandoned coin in seconds rather than at its transaction's TTL.
+  material. In-place balancing records only the coins the wallet moved onto the pending side, not inputs the caller had
+  already put in the transaction. A reservation is dropped wherever its booking is reverted, and when its TTL passes.
+- Bookings restored from a snapshot are released once sync reaches the chain tip, unless a reservation or a transaction
+  being tracked still accounts for the spend, returning an abandoned coin in seconds rather than at its transaction's
+  TTL.
+- Balancing a transaction in place now books the coins it selects, so two balance calls can no longer select the same
+  coin.
 - `UnshieldedWallet.revertUtxos(ids)` releases booked coins without the transaction that booked them, for a caller
   holding only a record of the ids.
 
@@ -32,5 +37,7 @@ take the TTL of the transaction the coins are booked for as a required last argu
 gains `revertUtxos` and `releaseRestoredPending`, so a custom implementation of that interface must add them. The public
 coin accessors are unchanged: `pendingCoins`, `availableCoins` and `balances` still report UTxOs.
 
-The pending-transactions store gains a new format version carrying reservations. A store written before this change
-still loads; one written after it cannot be read by an earlier version of `@midnightntwrk/wallet-sdk-capabilities`.
+Both stored formats gained an optional member rather than a new version, so a store written by this release is still
+readable by an earlier one: the unshielded snapshot carries each booking's expiry, and the pending-transactions store
+carries its reservations. A pending entry restored from a snapshot that predates expiries is granted a full transaction
+lifetime from the moment it is loaded.
--- a/docs/decisions/0008-unshielded-bookings-persist-and-expire.md
+++ b/docs/decisions/0008-unshielded-bookings-persist-and-expire.md
@@ -56,29 +56,47 @@ the pending side only, so state already corrupted in the field repairs itself on
 duplicate forward.
 
 **2. A booking carries the TTL of the transaction it was taken for**, and both sync capabilities release expired
-bookings on every applied update — the indexer path against wall time, the simulator path against simulator time. Past
-that instant the ledger rejects the transaction, so the reservation cannot still be valid. This bounds the damage of any
-abandoned transaction to that transaction's own lifetime, with no help from the caller, and it is the floor for a wallet
-used without the facade. A snapshot written before bookings carried a TTL decodes at the epoch, so the first sweep
-releases it — which is the repair such a snapshot needs.
+bookings on every applied update — the indexer path against wall time, the simulator path against simulator time. Once
+that instant has passed the ledger rejects the transaction, so the reservation cannot still be valid. This bounds the
+damage of any abandoned transaction to that transaction's own lifetime, with no help from the caller, and it is the
+floor for a wallet used without the facade. It is also the only thing that releases a booking: everything below decides
+what to hold, never what to free, so one mechanism owns release and two sweeps on different clocks cannot free the same
+coin twice.
+
+Two details decide whether this is safe. It sweeps only once sync has caught up with the chain: while a replay from an
+earlier cursor is running, a coin can be booked here and already spent by a transaction the replay has not delivered,
+and releasing it then offers a coin that is gone. And a booking expires strictly after its TTL rather than at it,
+because the ledger accepts an intent while its TTL is at or after the block's timestamp. A snapshot written before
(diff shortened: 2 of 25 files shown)


iangregsondev:
We need to review the ADR 0008 (added in this PR) under the section 

Relationship to the shielded and dust wallets

You will a table, there are some inconsistencies between wallets and naming conventions. I think this is fine but I wouldn't like some confirmation on it.

Also, reservations are being exposed on the new Unshielded wallet, but we do not do this on Shielded and Dust.

docs/decisions/0008-unshielded-bookings-are-not-persisted.md
```diff
@@ -0,0 +1,175 @@
...
+- A booking represents a caller's intent to submit. That intent cannot survive the process that holds it.
+- The disjointness of the two maps is load-bearing for every balance figure the SDK reports, yet it is maintained by
+  convention across several call sites rather than by the type.
+- Corrupted persisted state was unrecoverable without deleting the cache — the worst possible failure mode for a wallet.
+- Coin selection must still refuse a coin that a submitted, unconfirmed transaction is spending, across a restart.
+- Consumers already carry one breaking change for #697; a second one later is worse than a wider one now.
+
+## Considered Options
+
+- Keep bookings persisted, and reap them: guard `applyUpdate`, repair `restore`, expire at the transaction's TTL, and
```
kapke: This is the path to be taken.

The reason is that booking is not only process state. It also reflects reality the wallet "left". Please note that proper tracking of transaction history is tightly connected to proper tracking of booked coins.

For example - let's take a case of wallet issuing a swap counterpart, which needs to sit in external system for unspecified amount of time. If, in the meantime, the wallet is restarted, it looses booking information, which may lead to involuntary invalidation of that swap.

kapke (changes requested):

docs/decisions/0008-unshielded-bookings-are-not-persisted.md
iangregsondev: Agreed.

PR #699, opened 2026-08-28
feat(shielded): carry the local state across the hard fork as bytes, instead of resetting

agronmurtezi:
## What's in this PR

A hard-fork rehearsal against a real node and a real indexer (1.0.1 → 2.1.0-beta.1) showed that the shielded wallet loses its coins when it crosses the fork. The shipped migration started the post-fork wallet empty and expected the indexer to replay pre-fork history as new-version events. The indexer does not replay anything: it continues its commitment-tree indices across the boundary, so the empty wallet's tree (at index 0) meets the first post-fork event (at the old tree's size) and the ledger rejects it — over and over. The wallet shows a zero balance and never recovers.

This PR makes the shielded wallet carry its local state across the fork instead.

**The crossing is a byte-level round trip.** The two ledger majors share the identical `zswap-local-state[v6]` serialization — verified against both real ledgers, down to byte-identical encodings (the transaction codec moved between the majors; this one did not). So the migration serializes the pre-fork wallet's local state and deserializes it with the post-fork ledger: coins at their Merkle indices, the tree position, and the outputs the wallet was still expecting all cross losslessly, with no network round-trip and no reliance on anything the indexer may or may not re-serve. The only thing that cannot cross this way is the wallet's coin-hash bookkeeping, which needs secret keys the migration deliberately does not hold — it is declared pending and resolved on the first synchronization update, which always carries the keys. Snapshots written mid-crossing carry the declaration and finish the job on restore; older snapshots load exactly as before.

**The guard.** Nothing forces a future ledger major to keep this codec, so a characterization test pins it against both real ledgers: the shared header tag, the faithful round trip, and a spend from a byte-crossed state accepted by a post-fork chain (with a negative control the chain rejects). If a future ledger moves the codec, that test goes red naming both tags, the migration fails loudly at the boundary rather than corrupting anything, and the migration seam takes a ledger-provided translation as the replacement — the seam was designed for exactly that hand-over.

**What proves it.** The fork-simulation suite models what the real indexer actually does (the post-fork chain announces nothing) and asserts: the migrated state holds the coins at the moment of migration; a post-fork payment landing exactly where the pre-fork tree ended applies cleanly — the precise scenario that wedged on real infrastructure; a crossed coin can be spent on the post-fork chain; and the outputs a wallet was still expecting survive the crossing — an assertion that fails against the replay-era design. The integration tier runs the same proof against the ledger team's real v8-to-v9 state translation: the byte-crossed wallet's tree root equals the truly-translated chain's, and the chain accepts its spend. A permanent negative documents the failure mode this replaces, down to the exact ledger error.

**One breaking edge.** `PreviousLedgerWallet` — the shape a wallet presents to be migrated across the boundary — now requires the previous wallet's local state, described by the one capability the crossing needs: it can serialize itself. The previous variant's own state satisfies this as-is, so wallets built through the SDK are unaffected; only code that hand-builds that input has to add the field.

Dust and unshielded crossings are unchanged: the chain itself wipes and replays dust state at the fork, and unshielded state is public UTXO data carried over field for field.

## How to review

1. `packages/shielded-wallet/src/v2/Migration.ts` — the crossing and the design note it replaces.
2. `packages/shielded-wallet/src/v2/test/byteCrossing.test.ts` — the codec guard: what the two ledgers share, and what happens the day they stop sharing it.
3. `packages/shielded-wallet/src/v2/Sync.ts` — coin hashes resolving on the first keyed update.
4. `packages/shielded-wallet/src/test/forkSimulation.test.ts` — the crossing contract, the drill regression, the expected-outputs assertion, and the negative that shows why the carry exists.


agronmurtezi:

PR #701, opened 2026-08-31
fix(capabilities): fresh wallets start where their unread history starts, not at the tip

agronmurtezi:
## What's in this PR

A fresh wallet — one starting from a seed with no saved state — reads the chain's history from the very beginning. Before this change it asked the indexer which protocol version the chain is on **now** and started on the variant that owns that version. On a chain that has already hard-forked, that is the post-fork variant, which then receives the pre-fork history as bytes its ledger cannot deserialize: the shielded wallet fails on every historical event and ends up empty. We hit this live in a fork rehearsal.

The fix is one question asked differently: the default probe now reads the protocol version of the chain's **first block** instead of its latest one, because a fresh wallet needs the variant that can read the first event it is about to fetch. A chain that has been post-fork since its genesis (our test networks) still starts directly on the post-fork variant with no hand-over, exactly as before. A chain with real pre-fork history routes the fresh wallet to the pre-fork variant, which reads its coins and then crosses through the carrying migration — the path proven by the state-carry PR under this one.

Nothing else moves. Restores never used the probe (a snapshot names its own version), every consumer of the probe's answer was traced and none relied on tip semantics, and the no-answer fallback (start pre-fork) is unchanged. Applications that supply their own `chainVersionProbe` should note the question it must answer is now "where does the timeline start", not "where is the tip" — the type's documentation says so.

One known gap is left deliberately open here and closed by the next PR in the stack: on a forked chain with no traffic, a wallet routed to the pre-fork variant has no event to learn the new version from. That is the version watcher's job.

## How to review

1. `packages/capabilities/src/chainVersion/chainVersionProbe.ts` — the one-line semantic change and the re-stated docs.
2. `packages/capabilities/src/chainVersion/test/chainVersionProbe.test.ts` — the probe now provably asks about block height 0.
3. `packages/shielded-wallet/src/test/forkStart.test.ts` — the new contract: a fresh wallet on a forked chain with pre-fork coins ends up post-fork holding them, and the negative twin showing tip-routing leaves it empty.


agronmurtezi:

PR #702, opened 2026-08-31
feat(shielded,unshielded,dust): a version watcher so quiet wallets still notice a fork

agronmurtezi:
## What's in this PR

A wallet learns the chain's protocol version only from the events it receives, and the indexer's progress messages carry no version. So on a chain with no traffic after a hard fork, a wallet never finds out the fork happened: it sits on the pre-fork variant indefinitely, and everything built on top stays routed to the pre-fork version. We watched this live in a fork rehearsal — the wallet only crossed once somebody sent a transaction.

This PR gives every wallet a **version watcher**: on the indexer-backed sync path, a small periodic check (default every 30 seconds, configurable) reads the tip's protocol version and feeds it into the same observation path events use. Recording an out-of-range version is already what triggers the hand-over, so the watcher adds no new migration machinery — only the missing signal.

The heart of the change is the safety gate, because adopting a version too early would be worse than the disease: migration parks the sync cursor where it stands, and any *unread* pre-fork event would afterwards reach the wrong ledger's decoder — and unread coins would miss the carry. The watcher therefore adopts a version only when the wallet is provably caught up, established per wallet against the indexer's actual semantics:

- **Shielded and dust** (global event streams): the tip is read first, then a one-message probe of the event subscription reports the highest event id; the signal is adopted only when that id is at or below the wallet's own cursor. Tree indices are deliberately not used — an event can advance the stream without touching a tree.
- **Unshielded** (address-scoped): a one-message probe of the wallet's own subscription, opened one past its cursor — at that position a caught-up wallet can only receive the progress frame (the indexer emits one immediately, for any address), and a transaction frame arriving at all is proof of unread history. Opening at the cursor itself would re-deliver the already-applied boundary transaction forever; the tests pin this.

A failed or timed-out check is a skipped tick, never treated as caught up, and never disturbs the sync stream. Both variants of all three wallets get the watcher; the pre-fork variants are the ones that need it for this fork. Dust's projections-based fast sync is excluded and its existing known-limitation note says so — it has no event cursor for the gate to compare against. One documented edge: a chain with literally zero dust history produces no dust signal, and such a wallet crosses on its first dust event instead.

Proven end to end on the live fork rehearsal: fork enacted, **zero transactions sent**, and all three wallets crossed within one watcher tick — the shielded wallet arriving with its coins and exact balances. Each package also carries a regression test driving a real running pre-fork variant over a silent stream: without the watcher it times out, with it the hand-over fires.

## History note: the attempted progress-frame adoption (last three commits)

The indexer team shipped `4.4.0-rc.2-25da0487`, which puts the chain-tip `protocolVersion` on `UnshieldedTransactionsProgress` frames — enough to delete the unshielded watcher outright, which commits `1a18f24c` + `b2a6969d` did (schema/image pin + adoption, net −466 lines). CI's smoke e2e then failed, and the cause is the image, not the SDK: that build sits on midnight-ledger `9.1.0.0-rc.4` ("Ledger 8 → 9 hardfork + DUST migration"), while this repo's wallets prove with ledger-v9 `1.0.0-rc.3` and the e2e node (`2.0.0-rc.3`) runs the rc.3 ledger. The first fee-paying wallet transaction is accepted by the node and then **crashes the indexer** on re-apply ("dust spend proof failed to verify … disagreement on dust state on the declared time") — reproduced locally. Commit `618783ee` reverts the adoption byte-for-byte; it re-lands by reverting that revert once an indexer image carries the fields on an rc.3-compatible ledger, or the platform aligns on rc.4 end to end.

## How to review

1. `packages/shielded-wallet/src/v2/Sync.ts` — the reference implementation: the `VersionSignal` arm, the tick, the gate.
2. `packages/indexer-client/src/graphql/subscriptions/ZswapEventTip.ts` (+ `DustLedgerEventTip`, `UnshieldedTransactionTip`) — the one-message probes.
3. `packages/unshielded-wallet/src/v1/test/versionWatch.test.ts` and the dust/shielded twins — the crosses-on-the-watcher-alone regressions.
4. The unshielded probe-position reasoning lives in `UnshieldedTransactionTip`'s and the sync file's remarks — it is the subtlest piece.


agronmurtezi:

PR #706, opened 2026-09-01
feat(unshielded): release orphaned bookings at the fork crossing, instead of stranding them

agronmurtezi:
Found during the #688 review pass, planned and drill-reasoned since: **a UTXO booked by an in-flight transaction crossed the hard fork still booked, and nothing could ever un-book it.** Funds permanently unavailable, surviving serialization.

**The two halves that contradicted each other.** The cross-ledger migration deliberately carries pending UTXOs field-for-field. But un-booking runs through `ForkingUnshieldedWallet.revertTransaction`, which dispatches on the *active* variant — and the post-fork arm cannot unwrap a pre-fork handle, so it resolves as a no-op whose doc comment claimed "a transaction built on the other side of the boundary cannot have booked any of this variant's UTxOs". The migration made that claim false. No other path releases: the wallet keeps no pending-transaction store of its own, so no internal TTL exists, and the facade's fork-orphaning auto-revert (`orphanBeyond` → `ORPHANED_BY_FORK` → revert) lands on exactly the arm that no-ops.

**The fix: the crossing releases every carried booking into the available set.** A booking exists to prevent double-spending while its transaction might still land; the transaction codec moved at the boundary, so a pre-fork transaction can never be included in a post-fork block — the booking's reason expires at the boundary itself.

**Why the release is exact rather than heuristic.** The hand-over fires only after the pre-fork variant has applied the complete pre-fork timeline: the timeline path migrates on the first beyond-range event, and the quiet path's version signal is gated on there being no unread events. So a pending transaction that actually landed pre-fork has already confirmed and cleared its own bookings before the crossing — whatever is still booked at the boundary is definitively orphaned. Defense-in-depth if an unapplied pre-fork event ever slipped through anyway: `applyUpdate` removes confirmed spends from **both** maps, so a released-then-confirmed UTXO converges as if it had never been released.

**Observable change.** At the moment of crossing, the pending balance drops to zero and the available balance rises by the same amount; the total is unchanged. The facade's orphan auto-revert and any app-level revert with a retained pre-fork handle now correctly find nothing to release.

**What proves it.** Written red-first against the unfixed code (all failed with the booked UTXOs stuck in the pending set): the rewritten migration unit test in both twins ("releases previously booked UTXOs into the available set" — replacing "carries pending UTXOs as pending", which pinned the bug), and three whole-wallet fork-simulation tests: the regression (transfer built pre-fork, fork crossed, inputs available again), the now-true no-op (reverting the pre-fork handle after the crossing resolves void with nothing to touch), and a snapshot round-trip with nothing booked. A fourth whole-wallet test is a characterization that passed immediately, pinning the exactness argument: a booked UTXO whose transaction confirmed before the boundary is not resurrected.

**Two test-infrastructure changes worth a look.** The fork timeline's fixture UTXOs used intent hashes the real v8 ledger rejects (`Invalid character 'i'`), which made whole-wallet transfers impossible to build; they are now deterministic hex. One pre-existing test leaned on that unreadability — it asserted pre-fork routing via the ledger's rejection string, with a comment admitting the rotation "cannot be completed here". It now asserts the same routing claim on a completed rotation: the built handle unwraps within the pre-fork epoch to a `v8.Transaction`, the idiom of the test beside it.

## How to review

1. `packages/unshielded-wallet/src/v2/Migration.ts` — the release and the reasoning in the remarks (`v1/Migration.ts` is the twin mirror).
2. `packages/unshielded-wallet/src/test/forkSimulation.test.ts` — the regression, the now-true no-op, the snapshot round-trip, and the confirmed-before-boundary characterization.
3. `packages/unshielded-wallet/src/v2/test/migration.test.ts` (+ twin) — the rewritten contract.
4. `packages/unshielded-wallet/src/ForkingUnshieldedWallet.ts` — the revert doc comment, now true.
5. `packages/unshielded-wallet/src/test/{forkTimeline,forkHarness,forkStart.test}.ts` — the fixture fix and the rewritten routing test.

**Stack:** based on `feat/fork-version-watcher` (#702). The shielded analog of this bug (carried `pendingSpends` with no ledger API to release them) is a known, separately-tracked issue blocked on a ledger ask; dust is unaffected (its crossing carries no dust state).


agronmurtezi:

PR #707, opened 2026-09-01
feat(shielded,dust): a restored wallet can be started with material that crosses the fork

agronmurtezi:
Found during the #688 review pass as a hardening note, investigated to its real severity: **a wallet restored from a pre-fork snapshot had no valid start path at all.**

**The dead end.** Snapshots deliberately carry no key material, so `restore`/`tryRestore` produce an instance holding nothing. The only instance start, `start(v9Keys)`, fills only the post-fork key slot — on a pre-fork snapshot the active variant is v1, so it rejects with `MissingStartAuxError`: *"Start it from a seed, or hand it both versions' keys."* Both named remedies existed **only as statics**, and statics build fresh wallets: taking that path discards the restored, synced state the caller restored precisely to keep. Post-fork snapshots worked; pre-fork snapshots could not be started with anything.

**The fix: the instance mirrors the static surface.** `startWithSeed(seed)` retains material every variant derives its own keys from; `startWithKeys({ v8, v9 })` retains both key objects. Both share `start()`'s existing tail (retain → register the activation watcher once → resume sync on the current variant), factored into one private method. The rejection message now names the two instance remedies.

**Why `start(v9Keys)` stays.** It is the shared `ShieldedWalletAPI`/`DustWalletAPI` member the facade wires (`facade.start` → `shielded.start(keys.shielded)`), and the steady-state method: variant activation is forward-only, so once a wallet is past the fork — or restored from a post-fork snapshot — v8 keys are dead weight, and requiring `ShieldedKeysByEpoch` forever would force every application to import the pre-fork ledger (a second multi-megabyte WASM) to build keys that can never be used. `startWithSeed` remains the primary recommendation: no ledger imports, answers for every variant, this fork and the next.

**What proves it.** Written red-first: the two headline tests failed at typecheck (`TS2339: Property 'startWithKeys' does not exist`) and at runtime (`restored.startWithKeys is not a function`), and the rejection-message test failed on content. After implementation, everything is green. Three characterizations passed immediately, which is the point of them:

- restore of a **post-fork** snapshot + `start(v9Keys)` syncs (the working case, now pinned);
- the **static `startWithKeys` escape hatch crosses the fork** — it had zero test coverage until now;
- the runtime's activation-watcher containment: a failing watcher does not stop the runtime, does not stop other watchers on the same activation, and is logged.

Scope, verified: unshielded is exempt (watch-only sync, no start aux); the facade is unaffected (it has no restore); the single-variant `Custom*` wrappers don't cross forks.

**Reviewer-attention items, flagged deliberately:**
- One behavioural nuance in `start()`: its post-fork branch now resolves keys through the retained-keys path (`#resumeSyncOn(v2, postForkAux)`) instead of using the passed object directly — identical material, and what makes the tail shareable.
- The fork harnesses gained a `startFrom: 'seed' | 'keys'` knob so the static-hatch characterization can build through it.
- Dust harness subtlety: the pre-fork wire is a single-consumer queue, so the snapshot-writing wallet is stopped before the restored one starts (`Scope.close` stays idempotent).
- The `hard-fork-support` docs snippet's restore section now starts the restored wallet with `startWithSeed`, with a comment on which start serves which snapshot.

## How to review

1. `packages/shielded-wallet/src/ForkingShieldedWallet.ts` — the two instance methods, the shared `#startSynchronizing` tail, the updated messages (`ForkingDustWallet.ts` is the mirror).
2. `packages/shielded-wallet/src/test/restoredStart.test.ts` + dust twin — the headline flows and the pinned rejection.
3. `packages/{shielded-wallet,dust-wallet}/src/test/forkStart.test.ts` — the static-hatch crossing characterizations.
4. `packages/runtime/src/test/runtime.test.ts` — the watcher-containment pin.
5. `packages/docs-snippets/src/snippets/hard-fork-support.ts` — the restore flow.

**Stack:** based on `feat/unshielded-orphan-release` (#706). Follow-up recorded: once the v8 epoch is history, `startWithKeys`/`*KeysByEpoch` become deprecation candidates.


agronmurtezi:

PR #709, opened 2026-09-02
build(ledger): align the wallet on the ledger-v9 rc.4 line (node 2.1.0-beta.1, indexer 25da0487, proof-server 9.0.0-rc.7)

agronmurtezi:
Stacked on #707 (`feat/restored-wallet-cross-keys`). Draft: the rc.4 platform alignment, as one reviewable unit.

## What

Three commits:

1. **Re-land the progress-frame adoption** (`e0fbbde4`, a revert of #702's revert `618783ee`): unshielded reads the protocol version off the indexer's `UnshieldedTransactionsProgress` frames (indexer ≥ `4.4.0-rc.2-25da0487`) and drops its version watcher. Net −466 lines.
2. **Move the wallet onto `@midnightntwrk/ledger-v9` `1.0.0-rc.4`** (`656c7d34`) with node `2.1.0-beta.1` in the e2e compose. rc.4's `.d.ts` is byte-identical to rc.3's — nothing in the TypeScript surface moves; the change is inside the WASM.
3. **Prove against proof-server `9.0.0-rc.7`** (`4dd3338c`) — every local prover pin (e2e compose, prover-client and wallet-integration-tests fixtures, shared testcontainers helper).

## Why the prover pin is the load-bearing commit

Ledger rc.4 (`a5a01fc7`, "Improve dust circuit and validation") changed the dust `spend` **circuit**: the spend now binds the spent UTXO's nonce and its generation info to a shared initial nonce (a soundness fix), so the verifier key changed. The SDK's HTTP prover client sends only the proof preimage; the proof server resolves `midnight/dust/spend` from keys compiled into its binary. So the proof server's build — not the wallet's ledger WASM — decides which circuit a wallet proves against, and every rc.4 verifier (node `2.1.0-beta.1`, indexer `25da0487`) rejects an rc.3-circuit proof with `Malformed(InvalidDustSpendProof)`.

Proof-server image tags do not track ledger tags: `9.0.0-rc.4` was published 2026-07-02, five weeks before the circuit change; the `ledger-9.1.0.0-rc.4` tag declares proof-server `9.0.0-rc.7`. With only the prover changed, the smoke e2e goes from 3/3 failing to 3/3 passing with zero node rejections. Nothing in the wallet's dust arithmetic or spend construction was at fault. This is also the root cause of the `25da0487` indexer crash on rc.3 traffic that forced #702's revert.

## Verified locally

- `yarn verify` gates: typecheck, lint, unit (1353 tests), `byteCrossing` codec-identity test (rc.4 kept the zswap-local-state codec), `verify:wasm`.
- Integration: prover-client and wallet-integration-tests suites against proof-server rc.7.
- Smoke e2e (`turbo test-undeployed -- -t @smoke`): 3/3 pass on node `2.1.0-beta.1` + indexer `25da0487` + prover `9.0.0-rc.7`.
- Live hard-fork drill (1.0.1 chain → governance `set_code` → 2.1.0-beta.1 runtime, indexer `25da0487`): all three wallets crossed 0 → 2001000, balances byte-exact, zero indexer errors, and unshielded crossed on a progress frame alone with its watcher deleted.

## Known residuals (not in this PR)

- The in-process `WasmProver`'s default S3 key material (`dust/9/spend.*`) still serves the rc.3 circuit; in-process proving on rc.4 will hit the same rejection until that bucket is refreshed or versioned per ledger line.
- `packages/state-translation` pins `midnight-zswap` at `zswap-9.0.0-rc.3`; `zswap-9.0.0-rc.4` exists and should be bumped (verify:wasm passes today — the chain-state codec is shared).
- `infra/compose/docker-compose-remote-dynamic.yml` stays on prover rc.3: it targets deployed environments whose ledger line is theirs to declare.


agronmurtezi:

PR #710, opened 2026-09-02
feat(shielded,dust,unshielded): hard-fork drill follow-ups — state carry, timeline-start routing, version signals, orphan release, restored-wallet starts, ledger rc.4

agronmurtezi:
Stacked on #688 (`docs/api-hard-fork-snippet`). Everything the live hard-fork drill turned up, fixed, plus the move onto the ledger-v9 rc.4 line the post-fork chain runs.

## What the drill was

The SDK driven through a real ledger 8 → 9 fork: a node 1.0.1 chain with pre-fork history, a governance `set_code` to the 2.1.0-beta.1 runtime, indexer `4.4.0-rc.2-25da0487`, and the wallets synced before, across and after the boundary (`.context/hardfork-drill`, not part of the package). The fork simulation the unit tests use serves what the design assumed; the real indexer does not, and five things surfaced.

## Changes

**Shielded state crosses the fork as bytes** (`packages/shielded-wallet/src/v2/Migration.ts`, `byteCrossing.test.ts`). Both ledgers serialize `ZswapLocalState` under the same codec, so the migration is `v9.deserialize(v8.serialize())`: the commitment tree arrives whole, coins at their Merkle indices, expected outputs preserved. The indexer does **not** replay pre-fork history as new-version events; the previous design waited for that and lost the coins. A characterization test pins the shared codec against both ledger modules so a future move fails loudly. Coin commitments and nullifiers need keys a migration lacks, so they're declared pending and resolved on the first keyed sync update; snapshots in that window round-trip.

**Fresh wallets route on the timeline's start, not the tip** (`packages/capabilities`). The default probe asks block 0's protocol version; a chain that forked over history starts a fresh wallet pre-fork so it can read that history and cross. Renames: `currentChainVersion` → `timelineStartChainVersion`, `LatestBlockAnswer` → `BlockVersionAnswer` (under a `minor` changeset; moot in aggregate, since the beta line already carries a `capabilities` major).

**Quiet wallets notice the fork** (all three wallets, both variants, `indexer-client`). Unshielded reads the tip's protocol version off the `UnshieldedTransactionsProgress` frames it already receives (indexer ≥ `25da0487`, which puts `protocolVersion` next to `highestTransactionId`) and adopts it once caught up to that id — no polling. Shielded and dust have no progress arm on their event streams, so their sync sources re-ask the tip's version on a timer (`versionWatch.intervalMs`, default 30 s), gated by a tip-first / timeline-far-end-second check so a hand-over never outruns unread history. `WalletSyncUpdate` gains a `VersionSignal` arm on all three. Two new id-only subscriptions: `ZswapEventTip`, `DustLedgerEventTip`.

**Unshielded orphaned bookings are released at the crossing.** A pre-fork transaction can never land post-fork (codec moved), so UTxOs booked by still-pending transactions cross as available; the release is exact because the hand-over fires only after the complete pre-fork timeline is applied. Shielded has the same shape of problem and can't yet be fixed the same way — its pending spends live inside the ledger's local state and the ledger's clearing call is a no-op; tracked separately.

**Restored pre-fork wallets can be started.** Shielded and dust forking wallets gain `startWithSeed(seed)` / `startWithKeys({ v8, v9 })` as instance methods; `start(secretKeys)` stays post-fork-only and its refusal names the two that work.

**Ledger-v9 `1.0.0-rc.4`, node `2.1.0-beta.1`, indexer `25da0487`, proof-server `9.0.0-rc.7`.** rc.4's `.d.ts` is byte-identical to rc.3's; the change is inside the WASM. rc.4 changed the dust `spend` circuit (a soundness fix: nonce/generation-info binding), so rc.4 verifiers reject rc.3-circuit proofs with `Malformed(InvalidDustSpendProof)`. The SDK sends only the preimage and the proof server compiles the circuit in, so the prover's build decides — and proof-server image tags don't track ledger tags: `9.0.0-rc.4` predates the circuit change by five weeks; the rc.4 ledger declares `9.0.0-rc.7`. Every local prover pin moves to rc.7; the remote-dynamic compose file stays on rc.3 (deployed environments declare their own line).

## Verified

- `yarn verify` (typecheck, lint, 1353 unit tests incl. `byteCrossing`), `verify:wasm`, `yarn changeset status` aggregate unchanged after consolidating six changesets into one.
- Integration matrix and smoke e2e green in CI on the rc.4 stack (node 2.1.0-beta.1 + indexer 25da0487 + prover 9.0.0-rc.7).
- Live drill on the same stack: all three wallets crossed 0 → 2001000, balances byte-exact, zero indexer errors, unshielded crossed on a progress frame alone with its watcher deleted.

## Known residuals (tracked, not here)

- Shielded orphaned pre-fork spends lock coins post-fork (needs a ledger-side answer).
- In-process `WasmProver` default S3 key material (`dust/9/spend.*`) still serves the rc.3 circuit.
- `packages/state-translation` pins `midnight-zswap` at `zswap-9.0.0-rc.3`; `zswap-9.0.0-rc.4` exists.
- Dust v2 fast-sync has no fork hand-over (bites at the next fork only; indexer piece already shipped in 25da0487).


agronmurtezi:

PR #712, opened 2026-09-02
Hard-fork e2e lane, dust events→projections crossing, and the post-fork dust re-registration fix

agronmurtezi:
Stacked on #710 (`feat/ledger-rc4-alignment`). Review order: #689 → #690 → #688 → #710 → this.

## What this is

The hard-fork drill as a CI lane, a dust wallet crossing from event sync into projections sync on a real fork, and the fix for the SDK bug the lane found on its first full run: after the fork, a native-NIGHT wallet could not re-register for dust generation.

Until now no CI lane crossed a real protocol boundary. PR smoke, the nightly undeployed run and the integration matrix all boot node 2.1.0-beta.1 on its own genesis, so every chain is post-fork from block 1; fork coverage was the in-memory dual simulator plus the WASM translation. The three bugs the hand-run drill found in August (#699, #701, #702) were invisible to all of it. This PR moves that drill into the repo.

## The lane

- **`infra/compose/docker-compose-fork-dynamic.yml`** (new; the smoke compose is untouched). A one-shot container runs `build-spec` on the **old** node image (1.0.1, ledger 8) so genesis carries the pre-fork runtime; the **new** binary (2.1.0-beta.1) runs it; `indexer-standalone` and `proof-server` as in the smoke lane; the node toolkit behind a `tools` profile. Every image tag is `${VAR:-default}`. Ports are ephemeral, so a hand-run stack can coexist.
- **`packages/e2e-tests/src/tests/fork-fixture.ts`** (new). Brings the stack up, streams node/indexer/prover logs to `reports/fork-logs/` (testcontainers tears the stack down in `afterAll`, so nothing else could collect them), and `enactFork()`: runs the toolkit's governance `runtime-upgrade` via `docker compose run` (joins the project network), bisects the code-applying block by spec version, and checks the ledger state is readable at applied ± 1 on both the `midnight_*` RPC and `state_call`. The cNIGHT replay state is logged, not asserted.
- **`packages/e2e-tests/src/tests/hardFork.fork.test.ts`** (new), its own vitest project `fork` (`*.fork.test.ts`, `retry: 0` — a retry would re-run on an already-forked chain), `yarn turbo test-fork`. Eight sequential tests on one facade wallet (dev funded seed): syncs pre-fork on the old ledger (spec 1000000); fork enacted; crosses and settles post-fork with a `Crossing` phase observed; balances deep-equal pre-fork and no stuck pendings; post-fork snapshot restores on the version it declares; **re-registers its NIGHT for dust generation**; **spends an unshielded coin**; **spends a shielded coin that crossed the fork as bytes**.
- **`.github/workflows/e2e-hard-fork.yml`** (new): nightly 02:00 UTC, `workflow_dispatch` with the five image tags as inputs, and `pull_request` filtered to the lane's own files. Read-only packages token, `contents: read` + `checks: write`, junit report and `reports/` uploaded on every outcome. Not in the required `Tests` gate: ~6–10 min, and the toolkit image is Docker Hub only at this tag (anonymous pulls are rate-limited).

## Also: a dust wallet crossing from v8 event sync into v9 projections sync

No test anywhere ran that hand-over. The lane now carries a second facade wallet on the funded seed whose dust wallet is composed with the public `CustomForkingDustWallet`: v1 on event replay, v2 on the projections (fast) sync. It starts pre-fork, crosses on its own, and is compared with the events wallet at four points — pre-fork, after the crossing, after the re-registration, after both spends — on the dust coin set, both tree roots, and the balance at one shared instant. All four agree (12/12, 373 s locally). Two things learned: `withSync` drops `startAux`, so a two-variant projections builder must restate `withStartAuxDefaults()`; and the event path never writes `progress.highestIndex`, so a wallet arriving by migration runs its first projections pass with `dtimeCutoffHeight = 0` — safe (the snapshot is complete at the block hash regardless; only the dtime-update delta is bounded), and now logged at debug level. The runner's sync lock is released on stream completion, so `doSync` after the one-shot pass works; pinned in both `RunningV*Variant` twins.

## The finding, and the fix

The first run of the spend tests failed, for a reason that would have hit every native-NIGHT holder at the real fork.

The fork wipes the ledger's dust state; the chain-side replay restores cNIGHT-backed generation only. The node team's own fork e2e (`util/toolkit/tests/hardfork_e2e.rs`, §5c) re-registers its wallet for the same reason and notes the registration "funds itself from the retroactive DUST its now-generationless NIGHT accrued" — the ledger's `generationless_fee_availability` credits a registration with value × decay rate × time since the UTxO's ctime, over inputs it holds no generation for.

The SDK could not take that path. It decides whether a registration is self-funding from the indexer's `registeredForDustGeneration` flag on the Night UTxO, and post-fork every carried UTxO still said `true`: the indexer of the day (`4.4.0-rc.2-25da0487`) recorded the flag at creation and never revised it, and the wallet carried it across in `carryUtxo`. So the SDK built a zero-fee "re-registration" the node rejected (`Malformed(BalanceCheckOverspend)`), `waitForGeneratedDust` never resolved, and every transfer failed at build with `InsufficientFunds: could not balance dust`. No on-chain escape either: every account crossed with zero dust, so no sponsor existed.

The fix is on both sides:

- **Indexer.** `4.4.0-rc.5` (midnight-indexer #1476) scopes `registeredForDustGeneration` to the chain's current dust epoch at read time. Verified here: an rc.5 indexer pointed at an already-forked node reports `false` on all 20 genesis Night outputs, where rc.2 reports `true`. The lane's default indexer tag is now rc.5.
- **SDK.** The flag stays the one authority on whether a registration is self-funding. The unshielded cross-ledger migration carries it as `false` on every carried Night UTxO, both twins: the indexer never re-emits a UTxO the wallet already synced pre-fork, so the crossing is the only place the wallet's copy can be brought in line with what the indexer now says. `claimableFeePayment(dustState, nightUtxos, now)` is exported for callers that want the amount `waitForGeneratedDust` waits on.

History inside this PR, for the reviewer: while the indexer was still wrong, a wallet-side rule (`isGenerationless`, deciding from the wallet's own dust coins) was added; once rc.5 shipped it was removed again (`c6806987` tests, `a83e0c5e` fix), so the net diff has no such rule. Changeset: patch for dust-wallet, unshielded-wallet, facade.

## Verified

- Hard-fork lane locally, final code, indexer rc.5: **12/12 in 371 s** (governance `runtime-upgrade` ≈ 210 s of it). Fork applied at block ~31, spec 1000000 → 2001000; phases `Settled@0 → Crossing → Settled@2001000`. Re-registration accepted in 65 s; unshielded transfer finalized, receiver credited exactly 1000; **shielded transfer finalized with a genuinely pre-fork input** — the translated Merkle tree accepted the path; the projections twin agrees on the dust the spends paid with.
- Same lane with rc.5 against the previous code also 12/12, so the image follows a 1.0.1-genesis chain across the boundary on its own; rc.2-25da0487 followed it too but kept the stale flag. The drill's split indexer topology was not needed.
- Unit: dust-wallet 229, unshielded-wallet 253, facade 73; whole-repo typecheck, lint, Effect diagnostics, format and changeset check green.

## Worth knowing

- **First CI run should be a manual dispatch** of `E2E - hard fork`: the `runtime-blob` copy's amd64 path has only run on arm64 so far.
- Indexer `4.4.0-rc.5` was cut from an unmerged PR head and has no QA sign-off (its release notes say so). Only the fork lane moves to it; the smoke composes stay on rc.2-25da0487.
- The `indexer-standalone` and `proof-server` images ship no `curl`, so the healthchecks the smoke compose declares on them can never pass (masked there by port waits). The fork compose has none; the fixture polls `/ready`. Separate cleanup for the smoke compose.
- The `Crossing`-phase assertion depends on the three wallets not crossing inside a single state emission; it has held in every run.
- cNIGHT whose generation the chain-side replay restores is also carried as `false`; a registration built over it fails closed on the node (`InsufficientDustForRegistrationFee`) rather than mis-submitting. Documented at `carryUtxo`.
- Open, not in this PR: a fresh dust wallet starting on an already-forked chain replays the whole pre-fork dust timeline through v1 before migrating (correct end state, wasted work); tracked as a follow-up.
- Pre-existing v1/v2 asymmetry in dust `Transacting.ts` (`signatureData` vs `signature` into `addSignatures`), untouched.

### DIFF
--- a/.changeset/dust-registration-follows-the-ledger.md
+++ b/.changeset/dust-registration-follows-the-ledger.md
@@ -24,3 +24,10 @@ reports carried Night UTxOs as generating no dust, which is what the ledger says
 fee at all; `claimableFeePayment(dustState, nightUtxos, now)` is exported from the dust wallet for callers that want
 the amount `waitForGeneratedDust` waits on. Callers of the dust wallet's public API are unaffected — the added state
 argument is internal to the variant capabilities.
+
+Known limitation, deliberately accepted until the indexer reports `registeredForDustGeneration` as of the queried block
+rather than as of the UTxO's creation: Night registered to generate dust for _another_ wallet has no dust coin in the
+registering wallet, so it reads as generationless there, and a later re-registration by that wallet would claim a
+self-funding allowance the ledger does not grant. The node refuses it (`InsufficientDustForRegistrationFee`), so nothing
+is lost, but such a redesignation cannot be built for now. When the indexer flag becomes trustworthy the rule will
+combine both signals: generationless iff no backing dust coin and the flag is false.
--- a/.changeset/dust-registration-follows-the-ledger.md
+++ b/.changeset/dust-registration-follows-the-ledger.md
@@ -13,14 +13,10 @@ rejected with `Malformed(BalanceCheckOverspend)`; `waitForGeneratedDust` never r
 failed at build with `InsufficientFunds: could not balance dust`.
 
 The cross-ledger migration now carries `registeredForDustGeneration` as `false` on every carried Night UTxO. That is
-what the ledger says of them after the fork, and what indexer 4.4.0-rc.5 and later reports for them — it scopes the
-flag to the chain's current dust epoch, so a UTxO whose generation the fork wiped reads `false`. The indexer never
+what the ledger says of them after the fork, and what the indexer reports for them — it scopes the flag to the
+chain's current dust epoch, so a UTxO whose generation the fork wiped reads `false`. The indexer never
 re-emits a UTxO the wallet already synced before the fork, so the crossing is the only place the wallet's copy can be
 brought in line.
 
 `claimableFeePayment(dustState, nightUtxos, now)` is exported from the dust wallet for callers that want the amount
 `waitForGeneratedDust` waits on.
-
-Requires indexer >= 4.4.0-rc.5 for a wallet synced fresh after the fork. Older indexers report
-`registeredForDustGeneration` as of the UTxO's creation and never revise it, so carried Night still reads `true`, the
-SDK builds a zero-fee re-registration, and the node rejects it.
--- a/.changeset/dust-registration-follows-the-ledger.md
+++ b/.changeset/dust-registration-follows-the-ledger.md
@@ -7,27 +7,20 @@
 After the hard fork, a wallet holding native NIGHT can re-register for dust generation and the registration funds its
 own fee from retroactive dust. Previously it could not: the fork wipes the ledger's dust generation state and the
 chain-side replay restores cNIGHT-backed generation only, so such a wallet arrived post-fork with zero dust — while
-every carried Night UTxO still reported `registeredForDustGeneration: true`, because that is a creation-time value the
-indexer never revises. The SDK read that flag to decide whether a registration was self-funding, so it built a
-zero-fee "re-registration" the node rejected with `Malformed(BalanceCheckOverspend)`; `waitForGeneratedDust` never
-resolved, and every subsequent transfer failed at build with `InsufficientFunds: could not balance dust`.
+every carried Night UTxO still reported `registeredForDustGeneration: true`, the value it was created with. The SDK
+reads that flag to decide whether a registration is self-funding, so it built a zero-fee "re-registration" the node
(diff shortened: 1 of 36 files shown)


agronmurtezi:

PR #739, opened 2026-09-08
test(wallets): hard-fork migration, envelope and snapshot-parity coverage

jtsang586:
# Description

Test-only hard-fork coverage against `v2`. Six commits, **1015 insertions across 13 files, no production file
touched**, each with an empty changeset.

Every case here exists because reading the shipped suite showed something it could not fail on. Three of them exist
specifically because a fixture supplies the value a broken implementation would produce, so the assertion above it
holds either way:

- `unshielded v{1,2}/test/migration.test.ts` sets `highestTransactionId` to `params.appliedId`, so a migration that
  derived the source tip from the applied position — or dropped it and let it default — passed every test there. The
  pair is what `SyncProgress.isCompleteWithin` reads, so a collapsed tip makes a wallet report itself caught up at the
  moment it has the whole post-fork replay left to consume.
- The dust migration case asserts the state is built on this ledger version's parameters, and its fixture supplies the
  ledger's initial parameters — which is also what a migration ignoring its configuration reaches for.
- The dust and shielded serialization round trips run over an **empty** wallet and compare snapshot bytes with snapshot
  bytes, so they see neither a change applied to writer and reader together nor anything about value.

# Proposed Solution

**Migration seams.** The whole cursor a cross-ledger migration hands over, not just `appliedId`: the source tip carried
separately, a tip equal to the applied position not inflated, and `isConnected` reset to false at the hand-over. A
cross-ledger migration handed the **wrong** dust parameters — accepted without complaint because there is no validation
seam, a wrong decay rate misvaluing regenerated dust, and a wrong night-dust ratio changing nothing observable, so a
guard on that field alone would fix nothing.

**Snapshot envelopes.** The field set of dust's and shielded's snapshots pinned on both variants, that the two agree,
and that each field is encoded the way the reader expects with the ledger blob left opaque.

**Snapshot parity over wallets that hold something.** Dust: every UTxO with its generation info, both merkle roots and
their first-free indices, the balance valued at one fixed instant, and the sync position. Unshielded: both sides of the
available/pending split, a value larger than a double holds exactly, both dust-registration states, `ctime` as a
`Date`. Shielded: coin hashes asserted by value over a settled wallet holding two coins.

**One case is `it.fails`, deliberately.** `pendingDust` is absent from the dust snapshot schema and the restore path
passes an empty array outright, so a wallet that has spent Dust offers it as available again after a restart. Left
failing rather than fixed: carrying the entry conflicts with pending state that survives serialization and can never be
cleared, so whatever lands has to carry it **and** give it a way to expire — a scope decision, not a test one. Checked
to be a live signal: making the assertion hold reports `Expect test to fail` rather than passing quietly.

**One assertion pins behaviour rather than asking for better.** Unshielded's source tip is rebuilt from the applied
position on restore, so a restored wallet reports a gap of zero and believes itself caught up until its first sync
update lands. Recorded as it is.

# Testing

Every case mutation-verified on this line rather than assumed. The numbers are the argument for each test existing —
they say how much of the suite stays green when the behaviour under test is broken:

| Mutation | Result |
| --- | --- |
| Migration derives the source tip from `appliedId` | **256 of 257** green |
| Migration substitutes the ledger's initial parameters for its configured ones | **235 of 237** green |
| Dust envelope's `offset` renamed to `cursor` on writer **and** reader | **234 of 237** green — every round-trip test passes it |
| Dust snapshot drops the sync position | **234 of 235** green |
| Dust restore rebuilds an empty state instead of carrying the snapshot's | **231 of 235** green |
| Unshielded defaults the dust-registration flag on the way back | **259 of 260** green |
| Shielded derives the nullifier the way the commitment is derived | **253 of 254** green |

That last one is a wallet that cannot tell a spent coin from an unspent one, and it slips past all 276 other shielded
tests.

Stated plainly rather than overclaimed: for the shielded case the existing round-trip property catches more than
expected — the decoded wallet is still typed, so a swapped or renamed field changes the second serialization and shows
up as a byte difference. The new detection there is the semantic one above; the rest is scope, since that property's
generator has a maximum transaction count but no minimum and may assert over an empty wallet.

Suites after: dust 243 + 1 expected fail, shielded 258 + 22 pre-existing skips, unshielded 263. Typecheck, lint and
prettier clean. Rebased onto `v2` at `3ccc26e7` and re-verified.


agronmurtezi:
looks good, the changesets could've been merged in one file and probably written slightly more concisely

PR #741, opened 2026-09-08
fix(dust-wallet): terminate computeBalancingRecipe instead of looping forever

chrispalaskas:
# Description

`TransactingCapabilityImplementation.computeBalancingRecipe` in `@midnightntwrk/wallet-sdk-dust-wallet` can loop forever. When it does, the calling process stops responding and its memory grows without bound (measured: ~16 MB/s for the first two minutes, then ~1.2 MB/s sustained with no plateau — 3.4 GB at 15 minutes on a live devnet, killed by our timeout rather than by the OS), because every pass builds and proof-erases a fresh WASM transaction on the calling thread.

The loop is an `Effect.iterate` whose only exit is `newFee <= coverage`. It has no iteration cap and no check that a pass has made progress, and it has a sign bug:

| Pass | Seeded with | Sign | What `getBalanceRecipe` does |
|---|---|---|---|
| 1 | `transaction.imbalances(0, fee)['dust']` | negative (a deficit) | selects inputs until the deficit is covered |
| 2+ | the `dryRunFee` it just computed | **positive** | reads it as a *surplus*: adds a change output, selects **zero** inputs |

`getBalanceRecipe` is called with `inputFeeOverhead: 0n`, so it cannot know that selecting inputs raises the real fee — that is exactly why the outer loop exists. But once pass 1 under-covers its own fee, pass 2 is seeded with a positive number, selects nothing, and produces a state identical to itself. Nothing can ever change again. Only pass 1 can converge; every pass after a failed pass 1 is dead work.

This is easy to reach. Measured through the real `dryRunFee` on a devnet (node 1.0.1 / ledger 8.1.0), each dust input adds ~3.36e14 to the fee — a dust spend roughly doubles it. A wallet that has paid many fees holds several part-drained coins that are each below fee size; pass 1 spends several of them, the fee grows past what they cover, and the loop never returns.

# Proposed Solution

Each pass seeds the balancer with the **outstanding deficit** — the fee still unpaid by the coins already chosen *and by the dust the transactions already carry* — as a negative imbalance, over the coins not yet selected, then re-prices with everything selected so far via `dryRunFee`. That fixes the sign (every pass is a deficit, so every pass selects). The balancer is also handed the per-input fee measured over the passes so far as `inputFeeOverhead`, so it anticipates the growth the first pass could not see, and a second pass normally completes what the first left short.

Termination is structural, not a counter: a pass that consumes no coin from the pool fails (`SelectionMadeNoProgress`), so every pass shrinks the pool and the recursion is bounded by its size. The loop is a tail-recursive `Either.gen` over `const` bindings — no `let`, no loops, no mutation — with one `Data.TaggedError` per failure mode, mapped to `WalletError` exhaustively at the boundary.

**The largest-first retry, narrowed.** An additive selection is complete — finds a covering set whenever one exists — only when it takes coins from the top, since the k largest coins dominate every other k-subset. Under the default ascending `chooseCoin` it can take a small coin, be forced to add a large one to cover the resulting fee, and find the pair still short — when the large coin alone would have paid the fee for one input (a real devnet wallet holding 6.4e13 and 9.7e14 against a 7.3e14 fee did exactly that). So when the configured order **exhausts the pool**, selection is retried once largest-first. It is *not* retried when the configured selector **declines** coins: a selector returning `undefined` on a non-empty pool is expressing a policy, and that failure is surfaced as its own tagged error and reported as insufficient funds. The two were indistinguishable in the first revision of this PR; they are distinct now.

Alternatives considered: an iteration cap alone would bound the damage but is a heuristic; the no-progress failure is exact. Probing the marginal fee up front (`dryRunFee([anyCoin])`) would cost one WASM dry run on every call, including the common one-coin case; measuring it from the first pass costs nothing extra and collapses multi-coin wallets to two passes in practice.

- `computeBalancingRecipe` is rewritten; its signature and return type are unchanged, and `estimateFee` still routes through it. Dust the transactions already carry (existing spends, registration allowances) now counts toward coverage, so new inputs pay only the shortfall. The reported `fee` remains the **total** fee of the merged result, which is what the facade's `estimateTransactionFee` documents ("the total fee for the given transaction plus the fee of the balancing transaction").
- A transaction that already covers its fee produces **no balancing intent**: `balanceTransactions` returns an empty transaction (which merges as the identity — verified against the ledger) with the state untouched, instead of an intent with empty `DustActions`, which `DustActions::well_formed` rejects. The fee reported is that of the transactions as they are.
- Selected coins keep **selection order**, so `distributeFeeAcrossInputs` drains the coins chosen first (under the default order, the smaller ones) and no coin is spent for a zero `vFee`.
- `dryRunFee` gains an optional trailing `merged` parameter carrying the proof-erased, merged transactions and their segment id, so a caller pricing several input sets merges once; the 7-argument form still works unchanged. With no inputs it now prices the transactions as they are rather than with an extra empty intent. `balanceTransactions` uses the same merge helper instead of its own copy.
- Every ledger call — including the merge and the base-fee computation — runs inside `LedgerOps.ledgerTry`, so the declared `Either<_, WalletError>` is honoured on every path; nothing escapes as a raw throw.
- The `effect` import drops `Effect` and adds `Array as EArray` (the alias the language service enforces), `Data` and `Match`. No public API changes.
- Not touched, deliberately: the default `chooseCoin` selector (ascending). That is a legitimate default for tokens whose small coins are fragments worth consolidating; for dust — whose coins are bound one-to-one to registered NIGHT UTXOs, spend one-in-one-out, and regenerate toward a cap — largest-first is cheaper (7.27e14 with one input vs 1.97e15 with five, same transaction, on our devnet). A separate conversation from this correctness fix.

# Testing

Nine tests in `src/v1/test/transacting.test.ts` drive the real `computeBalancingRecipe` and `balanceTransactions` — including the real `getBalanceRecipe` and `Imbalances` — with `dryRunFee`/`calculateFee` replaced by an analytic model whose one relevant property matches the real one (the fee grows with each input). Where a test's title is a claim about work done, a spy on `dryRunFee` asserts the count:

- the classic shape (twelve part-drained coins beside two full ones) converges on the second pass, **two** prices, where the old loop never returned;
- selection order is kept, so the small coin is drained first and no input pays `0n`;
- the stranded-large-coin case recovers via the retry, two prices in total;
- a selector that declines coins gets insufficient funds and **no** retry — the excluded coin is never spent, zero prices;
- a selector returning a coin it was not offered fails fast with `TransactingError`, zero prices;
- a transaction already carrying 60% of its fee in dust has its new input pay only the shortfall, while the reported fee is the total;
- a fully-covered transaction selects nothing, and `balanceTransactions` returns an empty transaction with the state untouched;
- a genuinely unaffordable pool still reports `InsufficientFundsError`;
- a fee model that can never be covered still terminates, with dry runs bounded by the pool per attempt.

Run against the previous revision of this PR, the order, decline, no-progress, existing-dust and empty-intent tests **fail**; the other four pass on both, as regression guards for behaviour the first revision already had.

All 45 unit tests in the package pass, including the pre-existing `DustWallet.test.ts` spend/revert/deregister cases that exercise the rewritten methods end-to-end against the real ledger. Lint clean (`eslint --max-warnings 0`), `tsc` clean for the touched files.

Downstream this was first shipped as a client-side wrapper and verified against a live devnet with the real WASM fee loop — an A/B on one wallet with identical coin snapshots: SDK loop never returned (killed at 150 s), this loop converged in 51 ms, and with largest-first selection in 13 ms; the memory curve above is from that run. Details, traces and the RSS series are in shieldedtech/moth-wallet#141 and shieldedtech/moth-wallet#142. This PR is the same loop, applied where it belongs.

# Links

- shieldedtech/moth-wallet#142 — the same fix as a downstream wrapper, with the devnet traces and memory measurements
- shieldedtech/moth-wallet#141 — the largest-first selector, and the unit reproduction of the spin

### DIFF
--- a/packages/dust-wallet/src/v1/Transacting.ts
+++ b/packages/dust-wallet/src/v1/Transacting.ts
@@ -10,7 +10,16 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-import { Either, pipe, BigInt as BigIntOps, Iterable as IterableOps, Option } from 'effect';
+import {
+  Array as EArray,
+  BigInt as BigIntOps,
+  Data,
+  Either,
+  Iterable as IterableOps,
+  Match,
+  Option,
+  pipe,
+} from 'effect';
 import {
   DustActions,
   DustRegistration,
@@ -213,10 +222,67 @@ const distributeFeeAcrossInputs = <T extends { value: bigint }>(
     { result: [], remaining: fee },
   ).result;
 
+/** The transactions being balanced, proof-erased and merged once, with the segment a balancing intent may use. */
+type MergedTransactions = {
+  readonly transaction: ProofErasedTransaction;
+  readonly segmentId: number;
+};
+
 /**
- * Picks the largest-value coin from the pool. Used only as a fallback inside `computeBalancingRecipe` when the caller's
- * configured selection order leaves an affordable set of coins unselected (see the comment there) — not exported, since
- * overriding the configured order is this loop's own last resort, not a general-purpose selector.
+ * Proof-erases and merges the transactions being balanced. This does not depend on which dust inputs are being priced,
+ * so callers pricing several input sets against the same transactions do it once, not once per pass. `None` when there
+ * is nothing to balance against. May throw: it is ledger arithmetic, and callers wrap it in `LedgerOps.ledgerTry`.
+ */
+const mergeErasedTransactions = (
+  transactions: ReadonlyArray<FinalizedTransaction | UnprovenTransaction>,
+): Option.Option<MergedTransactions> =>
+  pipe(
+    transactions.map((tx) => tx.eraseProofs()),
+    EArray.match({
+      onEmpty: () => Option.none(),
+      onNonEmpty: ([first, ...rest]) => {
+        const transaction = rest.reduce((acc, tx) => acc.merge(tx), first);
+        return Option.some({
+          transaction,
+          segmentId: Option.getOrElse(findAvailableSegmentId(transaction), () => 1),
+        });
+      },
+    }),
+  );
+
+/** What `computeBalancingRecipe` produces: the total fee of the balanced result and what each selected coin pays. */
+type DustBalancingRecipe = {
+  readonly fee: bigint;
+  readonly recipeInputs: ReadonlyArray<CoinWithValue<Dust>>;
+};
+
+/** The pool ran out under the configured order before the fee was covered — an ordering artefact, retried largest-first. */
+class SelectionExhausted extends Data.TaggedError('DustBalancing.SelectionExhausted')<{
+  readonly message: string;
+}> {}
+
+/** The configured selector declined the coins it was offered — a policy, honoured as-is and never retried. */
+class SelectionRefused extends Data.TaggedError('DustBalancing.SelectionRefused')<{
+  readonly remaining: number;
+}> {}
+
+/** The selector returned coins that were not among those offered, so a pass consumed nothing and could not progress. */
+class SelectionMadeNoProgress extends Data.TaggedError('DustBalancing.SelectionMadeNoProgress')<{
+  readonly returned: number;
+  readonly remaining: number;
+}> {}
+
+/** `getBalanceRecipe` threw something other than its insufficient-funds error. */
+class BalancerFailed extends Data.TaggedError('DustBalancing.BalancerFailed')<{
+  readonly cause: unknown;
+}> {}
+
+type SelectionFailure = SelectionExhausted | SelectionRefused | SelectionMadeNoProgress | BalancerFailed;
+
+/**
+ * Picks the largest-value coin from the pool. Used only for the one retry inside `computeBalancingRecipe`, when the
+ * configured order exhausts the pool with an affordable set of coins left unselected; not exported, since it is that
+ * loop's last resort rather than a general-purpose selector.
  */
 const largestFirst: CoinSelection = (coins) =>
   coins
@@ -225,54 +291,159 @@ const largestFirst: CoinSelection = (coins) =>
     .at(0);
 
 /**
- * Selects dust inputs, under one coin-selection order, until they cover the fee of the transaction they produce. Each
- * pass hands the balancer only the fee still unpaid by the coins already chosen — the deficit, expressed as a negative
- * imbalance — over the coins not yet selected, then re-prices the transaction with everything selected so far via
- * `feeFor`. Adding an input can only raise the fee, so a pass that does not converge has strictly grown the input set;
- * with a finite coin pool that bounds the number of passes by the pool's size. Throws the balancer's own
- * `BalancingInsufficientFundsError` when the pool cannot cover the fee under this order — including when
- * `coinSelection` itself returns nothing from a non-empty pool.
+ * A selector declines by returning `undefined`, and the balancer folds that into the same `InsufficientFundsError` it
+ * throws for an exhausted pool. The two must stay distinguishable — declining is the caller's policy and must never be
+ * overridden by the largest-first retry — so a decline on a non-empty pool is surfaced as its own failure here.
+ * Throwing is the only way out of `getBalanceRecipe`, whose failure contract is exceptions; it is caught by the
+ * `Either.try` in `selectForDeficit` and never leaves this module.
  */
-const runBalancingPasses = (
-  coins: ReadonlyArray<CoinWithValue<Dust>>,
-  initialImbalance: bigint,
-  feeFor: (inputs: ReadonlyArray<CoinWithValue<Dust>>) => bigint,
+const surfacingRefusal =
+  (coinSelection: CoinSelection): CoinSelection =>
+  (coins) => {
+    const chosen = coinSelection(coins);
+    if (chosen === undefined && coins.length > 0) {
+      throw new SelectionRefused({ remaining: coins.length });
+    }
+    return chosen;
+  };
+
+/**
+ * One pass of selection: coins from `remaining`, in the configured order, until their value covers `deficit` plus
+ * `inputFeeOverhead` for each coin taken. Returned in selection order — `distributeFeeAcrossInputs` relies on that to
+ * drain the coins chosen first (under the default order, the smaller ones) before touching the later ones.
+ */
+const selectForDeficit = (
+  remaining: ReadonlyArray<CoinWithValue<Dust>>,
+  deficit: bigint,
+  inputFeeOverhead: bigint,
   coinSelection: CoinSelection,
-): { fee: bigint; recipeInputs: ReadonlyArray<CoinWithValue<Dust>> } => {
-  const inputs: CoinWithValue<Dust>[] = [];
-  let remaining = coins;
-  let fee = -initialImbalance;
+): Either.Either<EArray.NonEmptyReadonlyArray<CoinWithValue<Dust>>, SelectionFailure> =>
+  pipe(
+    Either.try({
+      try: () =>
+        getBalanceRecipe({
+          coins: remaining.map((coin) => ({ type: 'dust', value: coin.value, token: coin.token })),
+          initialImbalances: CapImbalances.fromEntry('dust', -deficit),
+          feeTokenType: 'dust',
+          coinSelection: surfacingRefusal(coinSelection),
+          transactionCostModel: { inputFeeOverhead, outputFeeOverhead: 0n },
+          createOutput: (coin) => coin,
+          isCoinEqual: (a, b) => a.token.nonce === b.token.nonce,
+        }),
+      catch: (err): SelectionFailure =>
+        err instanceof SelectionRefused
+          ? err
+          : err instanceof BalancingInsufficientFundsError
+            ? new SelectionExhausted({ message: err.message })
+            : new BalancerFailed({ cause: err }),
+    }),
+    Either.flatMap((recipe) => {
+      const added = EArray.filterMap(recipe.inputs, (input) =>
+        EArray.findFirst(remaining, (coin) => coin.token.nonce === input.token.nonce),
+      );
+      return EArray.isNonEmptyReadonlyArray(added)
+        ? Either.right(added)
+        : Either.left(new SelectionMadeNoProgress({ returned: recipe.inputs.length, remaining: remaining.length }));
+    }),
+  );
 
-  // Every pass either returns or consumes at least one coin from `remaining`, so this bound
-  // is never actually reached; it turns a broken invariant into a thrown error instead of an
-  // infinite loop.
-  for (let pass = 0; pass <= coins.length; pass++) {
-    const covered = inputs.reduce((sum, input) => sum + input.value, 0n);
-    if (fee <= covered) {
-      return { fee, recipeInputs: distributeFeeAcrossInputs(inputs, fee) };
-    }
+/**
+ * The fee grows with every dust input, and the balancer can anticipate that through `inputFeeOverhead`, which it
+ * subtracts from the outstanding deficit per coin it takes. The ledger's cost model does not expose the figure, so it
+ * is measured: the fee increase over the base fee so far, averaged over the inputs taken, rounded up. Zero until the
+ * first pass has priced something, so that pass selects on raw value and later passes complete what it left short.
+ */
+const marginalFeePerInput = (feeIncrease: bigint, inputs: number): bigint =>
+  inputs === 0 || feeIncrease <= 0n ? 0n : (feeIncrease + BigInt(inputs) - 1n) / BigInt(inputs);
 
-    const recipe = getBalanceRecipe({
-      coins: remaining.map((coin) => ({ type: 'dust', value: coin.value, token: coin.token })),
-      initialImbalances: CapImbalances.fromEntry('dust', covered - fee),
-      feeTokenType: 'dust',
-      coinSelection,
-      transactionCostModel: { inputFeeOverhead: 0n, outputFeeOverhead: 0n },
-      createOutput: (coin) => coin,
-      isCoinEqual: (a, b) => a.token.nonce === b.token.nonce,
+/**
+ * Selects dust inputs until, together with the dust the transactions already carry, they cover the fee of the
+ * transaction they produce. Each pass covers only the outstanding deficit from the coins not yet selected, then
+ * re-prices with everything selected so far. A pass that takes no coin fails as `SelectionMadeNoProgress`, so every
+ * pass shrinks `remaining` and the recursion is bounded by the pool size.
+ */
+const selectDustInputs = (args: {
+  readonly coins: ReadonlyArray<CoinWithValue<Dust>>;
+  readonly baseFee: bigint;
+  readonly existingDust: bigint;
+  readonly feeWith: (
+    inputs: EArray.NonEmptyReadonlyArray<CoinWithValue<Dust>>,
+  ) => Either.Either<bigint, LedgerOps.LedgerError>;
+  readonly coinSelection: CoinSelection;
+}): Either.Either<DustBalancingRecipe, SelectionFailure | LedgerOps.LedgerError> => {
+  const { baseFee, existingDust, feeWith, coinSelection } = args;
+  const go = (
+    remaining: ReadonlyArray<CoinWithValue<Dust>>,
+    inputs: ReadonlyArray<CoinWithValue<Dust>>,
+    fee: bigint,
+  ): Either.Either<DustBalancingRecipe, SelectionFailure | LedgerOps.LedgerError> =>
+    Either.gen(function* () {
+      const covered = BigIntOps.sumAll(inputs.map((input) => input.value));
+      const deficit = fee - existingDust - covered;
+      if (deficit <= 0n) {
+        return { fee, recipeInputs: distributeFeeAcrossInputs(inputs, fee - existingDust) };
+      }
+      const added = yield* selectForDeficit(
+        remaining,
+        deficit,
+        marginalFeePerInput(fee - baseFee, inputs.length),
+        coinSelection,
+      );
+      const nextInputs = EArray.appendAll(inputs, added);
+      const nextFee = yield* feeWith(nextInputs);
+      const rest = remaining.filter((coin) => !added.some((taken) => taken.token.nonce === coin.token.nonce));
+      return yield* go(rest, nextInputs, nextFee);
     });
-
-    const chosen = new Set(recipe.inputs.map((input) => input.token.nonce));
-    for (const coin of remaining) {
-      if (chosen.has(coin.token.nonce)) inputs.push(coin);
-    }
-    remaining = remaining.filter((coin) => !chosen.has(coin.token.nonce));
-
-    fee = feeFor(inputs);
-  }
-
-  throw new Error(`DUST fee balancing failed to converge after ${coins.length + 1} passes`);
+  return go(args.coins, [], baseFee);
 };
+
+/**
+ * Runs selection under the configured order and, only if that exhausts the pool, once more largest-first: an additive
+ * selection is guaranteed to find a covering set only when it takes coins from the top, since the k largest coins
+ * dominate every other k-subset. Any other failure — a declining selector above all — is reported as it is.
+ */
+const selectWithLargestFirstRetry = (
+  select: (
+    coinSelection: CoinSelection,
+  ) => Either.Either<DustBalancingRecipe, SelectionFailure | LedgerOps.LedgerError>,
+  configured: CoinSelection,
+): Either.Either<DustBalancingRecipe, SelectionFailure | LedgerOps.LedgerError> =>
+  pipe(
+    select(configured),
+    Either.orElse((failure) =>
+      failure._tag === 'DustBalancing.SelectionExhausted' ? select(largestFirst) : Either.left(failure),
+    ),
+  );
+
+const toWalletError = (failure: SelectionFailure | LedgerOps.LedgerError): WalletError =>
+  Match.value(failure).pipe(
+    Match.tag(
+      'DustBalancing.SelectionExhausted',
+      ({ message }) => new InsufficientFundsError({ message, tokenType: 'dust' }),
+    ),
+    Match.tag(
+      'DustBalancing.SelectionRefused',
+      ({ remaining }) =>
+        new InsufficientFundsError({
+          message: `Insufficient Funds: the configured coin selection declined the remaining ${remaining} dust coin(s)`,
+          tokenType: 'dust',
+        }),
+    ),
+    Match.tag(
+      'DustBalancing.SelectionMadeNoProgress',
+      ({ returned, remaining }) =>
+        new TransactingError({
+          message: `Coin selection returned ${returned} coin(s), none of them among the ${remaining} offered; a selector must pick from the coins it is given`,
+        }),
+    ),
+    Match.tag(
+      'DustBalancing.BalancerFailed',
+      ({ cause }) =>
+        new OtherWalletError({ message: cause instanceof Error ? cause.message : 'Dust balancing failed', cause }),
+    ),
+    Match.tag('LedgerError', (error) => error),
+    Match.exhaustive,
+  );
 
 export class TransactingCapabilityImplementation<TTransaction extends AnyTransaction> implements TransactingCapability<
   DustSecretKey,
@@ -628,6 +799,12 @@ export class TransactingCapabilityImplementation<TTransaction extends AnyTransac
     );
   }
 
+  /**
+   * The fee the transactions would carry with `recipeInputs` attached as dust spends in a balancing intent. With no
+   * inputs there is no intent to attach — an intent with empty `DustActions` is not well-formed — so it is the fee of
+   * the transactions as they are. Nothing is persisted. `merged` lets a caller pricing several input sets against the
+   * same transactions proof-erase and merge them once; it is derived from `transactions` when omitted.
+   */
   dryRunFee(
     recipeInputs: ReadonlyArray<CoinWithValue<Dust>>,
     transactions: ReadonlyArray<FinalizedTransaction | UnprovenTransaction>,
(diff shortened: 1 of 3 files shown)


iangregsondev:
I will be taking a look at this shortly. @chrispalaskas

packages/dust-wallet/src/v1/Transacting.ts
```diff
@@ -629,76 +702,40 @@ export class TransactingCapabilityImplementation<TTransaction extends AnyTransac
       0n,
     );
 
+    // A non-negative imbalance means the transaction already carries enough dust of its own;
+    // nothing to select. Reported via `feeFor([])`, not `-initialFees`, since the two coincide
+    // only in the (usual) case where the imbalance is exactly the negated fee.
+    const feeFor = (inputs: ReadonlyArray<CoinWithValue<Dust>>): bigint =>
+      this.dryRunFee(inputs, transactions, secretKey, state, ttl, currentTime, ledgerParams);
+    if (initialFees >= 0n) {
```
iangregsondev: **Bug: the `initialFees >= 0n` early return makes `balanceTransactions` emit an intent with empty `DustActions`, which the ledger rejects.**

With `recipeInputs: []`, `balanceTransactions` (below, ~L766-788) still builds `new DustActions(..., [] /* spends */, [] /* registrations */)` inside `Intent.new(ttl)` and returns it; every facade caller (`packages/facade/src/index.ts` ~L832, 895, 966, 1143, 1264) merges that transaction unconditionally. The ledger's `DustActions::well_formed` fails it: `if self.spends.is_empty() && self.registrations.is_empty() { return Err(MalformedTransaction::NotNormalized) }` (`ledger/src/dust.rs` ~L768; spec `dust.md` asserts the same).

This input is reachable: any transaction that already carries a `DustSpend` covering its fee, or a registration with `allowFeePayment >= fee`, yields a non-negative dust imbalance (the ledger credits both toward `dust_bal`). The old code hung here, so this turns a hang into a malformed transaction. Separately, the reported `fee: feeFor([])` prices the merged tx *including* the extra empty intent, but nothing checks the tx's surplus (measured per-tx at its standalone fee) still covers that larger figure.

Suggested fix: when no inputs are needed, do not build a balancing intent at all (return the inputs unchanged / `Option.none`), and report the fee of the transactions as they are.

packages/dust-wallet/src/v1/Transacting.ts
```diff
@@ -213,6 +213,67 @@ const distributeFeeAcrossInputs = <T extends { value: bigint }>(
...
+      initialImbalances: CapImbalances.fromEntry('dust', covered - fee),
+      feeTokenType: 'dust',
+      coinSelection,
+      transactionCostModel: { inputFeeOverhead: 0n, outputFeeOverhead: 0n },
+      createOutput: (coin) => coin,
+      isCoinEqual: (a, b) => a.token.nonce === b.token.nonce,
+    });
+
+    const chosen = new Set(recipe.inputs.map((input) => input.token.nonce));
+    for (const coin of remaining) {
```
iangregsondev: **Bug: selected coins are appended in pool order, not selection order, so `distributeFeeAcrossInputs` no longer drains smaller inputs first and can emit zero-`vFee` dust spends.**

`remaining` comes from `getAvailableCoinsWithGeneratedDust`, which is ledger-state order (no sort). The old code passed `recipe.inputs` (ascending under `chooseCoin`) straight to `distributeFeeAcrossInputs`, whose doc says "draining smaller inputs first" but which simply walks the array.

Example: pool `[big=2e15, small=1e12]`, `fee(0)=1.4e15`, `fee(2)=1.4075e15`. Pass 0 selects small then big (sum >= 1.4e15); `fee(2) <= sum` so it converges; this loop pushes `[big, small]`; the split is `{big: 1.4075e15, small: 0n}`. `balanceTransactions` then calls `CoreWallet.spendCoins` -> `localState.spend(secretKey, small, 0n, ...)`: the ledger accepts `v_fee = 0`, so the small coin is nullified, re-committed and locked in `pendingDust` for nothing, while the big coin is drained instead - the opposite of what a part-drained-coin fix wants. None of the new tests catch it (equal-valued coins, or sums only).

Suggested fix: build `inputs` in `recipe.inputs` order (map each recipe input back to its `remaining` coin by nonce), or sort `inputs` ascending before `distributeFeeAcrossInputs`.

packages/dust-wallet/src/v1/Transacting.ts
```diff
@@ -629,76 +702,40 @@ export class TransactingCapabilityImplementation<TTransaction extends AnyTransac
       0n,
     );
 
+    // A non-negative imbalance means the transaction already carries enough dust of its own;
+    // nothing to select. Reported via `feeFor([])`, not `-initialFees`, since the two coincide
+    // only in the (usual) case where the imbalance is exactly the negated fee.
+    const feeFor = (inputs: ReadonlyArray<CoinWithValue<Dust>>): bigint =>
+      this.dryRunFee(inputs, transactions, secretKey, state, ttl, currentTime, ledgerParams);
+    if (initialFees >= 0n) {
+      return Either.right({ fee: feeFor([]), recipeInputs: [] });
```
iangregsondev: **Bug: `feeFor([])` runs the full WASM `dryRunFee` sequence outside `Either.try`, so a ledger throw escapes as a raw exception instead of `Either.left(WalletError)`.**

`dryRunFee` calls `CoreWallet.spendCoins`, `tx.eraseProofs()`, `acc.merge(tx)`, `Transaction.fromParts(...).addIntent(...)` and `feesWithMargin`; `merge` of two input transactions sharing a fallible segment id is a realistic thrower (`findAvailableSegmentId` only protects the balancing intent). The `Either.try` boundary begins six lines later, so this branch is unguarded, whereas the old code ran every `dryRunFee` inside `Effect.try`.

Downstream, `RunningV1Variant.estimateFee`/`balanceTransactions` call this inside `Effect.flatMap` / `SubscriptionRef.modifyEffect` callbacks, so the throw becomes a fiber defect (`Cause.die`), and `DustWallet.estimateFee`'s `Effect.runPromise` rejects with a `FiberFailure` rather than a typed `WalletError` - the declared `Effect<bigint, WalletError>` is not honoured on this path.

Suggested fix: move the early return inside the `Either.try` body (or wrap `feeFor([])` in `LedgerOps.ledgerTry`).

packages/dust-wallet/src/v1/Transacting.ts
```diff
@@ -629,76 +702,40 @@ export class TransactingCapabilityImplementation<TTransaction extends AnyTransac
...
+        try {
+          return runBalancingPasses(dust, initialFees, feeFor, coinSelection);
+        } catch (err) {
+          if (!(err instanceof BalancingInsufficientFundsError)) throw err;
+          // The additive loop above is complete only when it selects from the top: adding a
+          // coin can only raise the fee, so under an ascending order it can take a small coin,
+          // be forced to add a large one to cover the resulting fee, and find the pair's fee
+          // exceeds the pool — when the large coin alone would have paid the fee for one input.
+          // Retrying once, greedily by value, recovers exactly that case before giving up.
+          return runBalancingPasses(dust, initialFees, feeFor, largestFirst);
```
iangregsondev: **The `largestFirst` retry fires on *any* `BalancingInsufficientFundsError`, so it silently overrides a caller-configured selection *policy*, not just its order.**

`Balancer.doBalance` throws `InsufficientFundsError` whenever the selector returns `undefined` (`Balancer.ts` ~L131), and `CoinSelection` is `(coins) => TCoin | undefined` - returning `undefined` is the selector's way to refuse. The `runBalancingPasses` JSDoc even names that case ("including when `coinSelection` itself returns nothing from a non-empty pool"), yet the guard here is only `instanceof`, so "selector refused" and "pool too small under this order" are indistinguishable and both trigger the override. A `withCoinSelection` policy that reserves nonces or caps coin size (the `capabilities` README's custom-selection example is a *filter*, i.e. a policy) will have its excluded coins spent on the retry with no signal to the caller. The changeset frames this as an ordering difference; it is a set-membership difference too.

Suggested narrowing: only retry when the configured selector never returned `undefined` (i.e. the failure was an ordering artefact), or restrict `largestFirst` to coins the configured selector would itself return (probe it on singletons), or make the retry opt-in.

packages/dust-wallet/src/v1/Transacting.ts
```diff
@@ -213,6 +213,67 @@ const distributeFeeAcrossInputs = <T extends { value: bigint }>(
...
+      isCoinEqual: (a, b) => a.token.nonce === b.token.nonce,
+    });
+
+    const chosen = new Set(recipe.inputs.map((input) => input.token.nonce));
+    for (const coin of remaining) {
+      if (chosen.has(coin.token.nonce)) inputs.push(coin);
+    }
+    remaining = remaining.filter((coin) => !chosen.has(coin.token.nonce));
+
+    fee = feeFor(inputs);
```
iangregsondev: **Dust the transactions already carry is counted at seed time and then forgotten, so from pass 1 the newly selected coins are made to pay the *full* merged fee - overpaying by the pre-existing dust.**

Pass 0 seeds `fee = -initialImbalance`, the netted deficit (the ledger credits existing `v_fee` and `allowFeePayment` toward the dust balance, so `feeImbalance` already subtracts them). This line then replaces `fee` with `feeFor(inputs)`, the full merged fee, and the convergence check `fee <= covered` sums only `inputs`. Example: tx already spends 0.6F of dust (imbalance -0.4F); pass 0 selects ~0.4F; `fee` becomes F; pass 1 seeds `covered - F` and selects another ~0.6F; result: new inputs pay F on top of the 0.6F already in the tx, and `estimateFee` reports F rather than 0.4F. The old code had the same check but hung on this shape, so the PR turns a hang into an overpay.

Related: for two-plus transactions (facade passes `[tx, mergedBalancingTx]`) `initialFees` sums per-tx *standalone* fees while `feeFor` prices the *merged* tx, so pass 0 can over-select and the additive loop never releases coins. Seeding from the merged tx (`-feeFor([])` plus existing dust) and carrying the existing-dust credit through every pass (`deficit = feeFor(inputs) - (covered + existingDust)`) would make both quantities consistent by construction.

packages/dust-wallet/src/v1/Transacting.ts
```diff
@@ -213,6 +213,67 @@ const distributeFeeAcrossInputs = <T extends { value: bigint }>(
...
+  coins: ReadonlyArray<CoinWithValue<Dust>>,
+  initialImbalance: bigint,
+  feeFor: (inputs: ReadonlyArray<CoinWithValue<Dust>>) => bigint,
+  coinSelection: CoinSelection,
+): { fee: bigint; recipeInputs: ReadonlyArray<CoinWithValue<Dust>> } => {
+  const inputs: CoinWithValue<Dust>[] = [];
+  let remaining = coins;
+  let fee = -initialImbalance;
+
+  // Every pass either returns or consumes at least one coin from `remaining`, so this bound
```
iangregsondev: **This comment is false as stated: the bound *is* reachable, and when it is, the failure is misreported and expensive.**

The invariant "every pass consumes at least one coin from `remaining`" only holds if `coinSelection` returns a member of the array it was given. Neither the `CoinSelection` type nor `Balancer.doBalance` enforces that: `doBalance` `addInput`s whatever coin is returned (deficit satisfied) and its `isCoinEqual` filter removes nothing if the coin is absent. Back here, `chosen` then matches nothing in `remaining`, `inputs`/`fee` do not change, and every subsequent pass repeats the identical WASM `dryRunFee` until `pass > coins.length`. The generic `Error` is then rethrown past the `largestFirst` catch (not a `BalancingInsufficientFundsError`) and lands as `OtherWalletError("failed to converge...")`, which names neither the selector nor the coin. A selector that closes over the wallet's full coin list and ignores its `coins` argument is the realistic trigger; no test covers this path.

Suggested fix: express the invariant directly - if no nonce from `recipe.inputs` is in `remaining`, fail immediately on that pass with a tagged error naming the selector - and drop the "never actually reached" claim (or the bound).

packages/dust-wallet/src/v1/test/transacting.test.ts
```diff
@@ -269,3 +272,141 @@ describe('addDustRegistrationSignature', () => {
...
+
+  it('reports InsufficientFundsError, not a hang, when no order of the pool can pay', () => {
+    const coins = [coin(1n, 0), coin(2n, 1)];
+    const error = run(makeTransacting(coins, chooseCoin)).pipe(EitherOps.getOrThrowRight);
+
+    expect(error).toBeInstanceOf(InsufficientFundsError);
+    expect((error as InsufficientFundsError).tokenType).toBe('dust');
+  });
+
+  it('never terminates worse than the coin count: converges on a uniformly tiny wallet too', () => {
```
iangregsondev: **This test does not test its title: nothing observes the pass count, and it is a strict subset of the first test.**

With 12 x 176.25e12 under ascending `chooseCoin`: pass 0 takes 8 coins (1.41e15 >= fee(0) = 1.4e15), fee(8) = 1.43e15 > 1.41e15; pass 1 adds 1 coin; converged. Two `dryRunFee` calls against a bound of 13. The two assertions (`recipeInputs.length <= coins.length`, trivially true since inputs come from the pool; `sum >= fee`) would still pass if the loop bound were removed entirely or set to `coins.length * 100`. The first test (`converges once the deficit-seeded pass covers the fee...`) follows the identical path (the two large coins are never reached) and asserts strictly more.

Suggested fix: spy on `dryRunFee` and assert the call count (`toHaveBeenCalledTimes(2)`, or `<= coins.length + 1` to match the title), ideally on a coin set that forces one coin per pass so the bound is actually approached; otherwise fold this into test 1.

packages/dust-wallet/src/v1/Transacting.ts
```diff
@@ -213,6 +213,67 @@ const distributeFeeAcrossInputs = <T extends { value: bigint }>(
...
+    if (fee <= covered) {
+      return { fee, recipeInputs: distributeFeeAcrossInputs(inputs, fee) };
+    }
+
+    const recipe = getBalanceRecipe({
+      coins: remaining.map((coin) => ({ type: 'dust', value: coin.value, token: coin.token })),
+      initialImbalances: CapImbalances.fromEntry('dust', covered - fee),
+      feeTokenType: 'dust',
+      coinSelection,
+      transactionCostModel: { inputFeeOverhead: 0n, outputFeeOverhead: 0n },
```
iangregsondev: **Efficiency / altitude: the loop rediscovers the fee model one WASM dry-run per extra coin, and each dry-run repeats input-independent work.**

(1) `CounterOffer.addInput` (`capabilities/src/balancer/CounterOffer.ts` L70) already subtracts `inputFeeOverhead` from the fee imbalance on every selected coin, so the shared balancer can anticipate fee growth; with `0n` it stops as soon as the raw sum covers the seeded deficit, and every later pass adds ~1 coin and pays another full `dryRunFee`. Measuring the marginal cost once (`feeFor([anyCoin]) - feeFor([])`, same units as `calculateFee`) and passing it as `inputFeeOverhead` lets pass 0 select a set that already covers fee(k), collapsing many-tiny-coin wallets to 1-2 passes. Note: `ledgerParams.transactionCostModel.inputFeeOverhead` is declared in the ledger d.ts but is `undefined` at runtime (the WASM object only exposes `baselineCost`/`runtimeCostModel`), so the measured variant is the only workable source. This does *not* remove the need for the `largestFirst` retry (the PR's own stranding test still fails under ascending order with the overhead modelled) - it only cuts passes.

(2) Independently of (1): `feeFor` calls `dryRunFee`, whose `transactions.map(tx => tx.eraseProofs())`, the `reduce(... acc.merge(tx))` and `findAvailableSegmentId(mergedExisting)` (L655-658) do not depend on `recipeInputs` and are redone every pass, and `CoreWallet.spendCoins(state, secretKey, recipeInputs)` re-spends the whole cumulative set each pass (O(N^2) `DustLocalState.spend` WASM calls). `merge`/`eraseProofs` return fresh values (no `__destroy_into_raw` in the wasm-bindgen glue), so `mergedExisting`/`segmentId` can be hoisted into `computeBalancingRecipe` and passed in; `dryRunFee` is not on the `TransactingCapability` interface and the test override declares only its first parameter, so the signature change is local. `balanceTransactions` L781-783 duplicates the same block and would use the same helper.

packages/dust-wallet/src/v1/Transacting.ts
```diff
@@ -213,6 +213,67 @@ const distributeFeeAcrossInputs = <T extends { value: bigint }>(
...
+ * `coinSelection` itself returns nothing from a non-empty pool.
+ */
+const runBalancingPasses = (
+  coins: ReadonlyArray<CoinWithValue<Dust>>,
+  initialImbalance: bigint,
+  feeFor: (inputs: ReadonlyArray<CoinWithValue<Dust>>) => bigint,
+  coinSelection: CoinSelection,
+): { fee: bigint; recipeInputs: ReadonlyArray<CoinWithValue<Dust>> } => {
+  const inputs: CoinWithValue<Dust>[] = [];
+  let remaining = coins;
```
iangregsondev: **Conventions: `runBalancingPasses` is imperative (`let`, `for`, `push`, reassignment) and uses thrown exceptions as its failure contract, in SDK code governed by `.claude/rules/functional-style.md`.**

Rule (functional-style.md, `paths: packages/**/*.ts`): "`const` only - never `let`. No `for`/`while` loops - use `map`/`filter`/`reduce`/`flatMap`." "No mutation: no `push`/`pop`/`splice`." "Exceptions (rare, must be isolated and justified): measured performance bottlenecks, inherently mutable external APIs, test setup/teardown." Offending lines: 242 (`const inputs: ...[] = []`), 243-244 (`let remaining`, `let fee`), 249 (`for (let pass ...)`), 266-267 (`for ... inputs.push`), 269, 271. The only comment at the site justifies the loop *bound*, not the departure from functional style; the `Effect.iterate` code this replaces was mutation-free, so this is a regression on that axis, not a carried-over smell.

Also: "Total functions: ... Never throw for expected errors" and "one `Data.TaggedError` per failure mode ... never a thrown exception". The new function's documented contract is to *throw* `BalancingInsufficientFundsError` (an expected outcome, per the new tests), and the caller (~L719-729) dispatches the retry via nested `try/catch` + `instanceof` + rethrow inside an `Either.try` that checks the same class again. The rule-fitting shape is cheap and shorter: a tail-recursive `go(remaining, inputs, covered)` (or `EArray.reduce`) returning `Either<Result, BalancingInsufficientFundsError | NoProgress>`, then `Either.orElse` for the `largestFirst` retry and one error-mapping site. `Arr.partition` (already used in `capabilities/pendingTransactions.ts`) replaces the `chosen` Set + `for...of` + `filter`, and `BigIntOps.sumAll` (used three times in this file) replaces the manual `reduce` at L250.

packages/dust-wallet/src/v1/Transacting.ts
```diff
@@ -614,6 +675,18 @@ export class TransactingCapabilityImplementation<TTransaction extends AnyTransac
...
   }
 
+  /**
+   * Balances a fee-paying dust intent: selects enough dust to cover the transaction's fee, and reports the fee those
+   * inputs must pay.
+   *
+   * The previous implementation of this method (`Effect.iterate`) had no bound on the number of passes and no check
+   * that a pass had made progress. Its first pass was seeded with the transaction's dust imbalance, which is negative
+   * (a deficit); every later pass was seeded with the fee it had just computed, which is positive. `getBalanceRecipe`
+   * treats a non-negative seed as a surplus of the fee token — it adds a change output and selects zero inputs — so any
```
iangregsondev: **Conventions: this JSDoc calls Dust "the fee token" (also `.changeset/dust-fee-balancing-terminates.md` L9, which ships verbatim into the CHANGELOG), and lines 682-688 narrate the previous implementation rather than the current one.**

Repo `CLAUDE.md`: "**Dust** - fee-payment resource generated from Night. **Never call Dust a \"token\".**"; `.claude/rules/transactions.md`: "Dust is a resource generated from Night - never call it a token." The sentence here is describing the transaction's *dust* imbalance, so "a surplus of the fee token" reads as "dust is the fee token". Suggested: "a surplus of dust" in both places.

Separately, "The previous implementation of this method (`Effect.iterate`) had no bound..." documents code that no longer exists and duplicates the changeset. The `runBalancingPasses` JSDoc already states the current invariant (deficit-seeded, monotone fee, pool-size bound); this block can shrink to two sentences of current behaviour plus the one largest-first retry, and leave the history to the changeset.

iangregsondev:
@kapke Would you mind taking a look

FYI @agronmurtezi

packages/dust-wallet/src/v1/Transacting.ts
```diff
@@ -213,6 +213,67 @@ const distributeFeeAcrossInputs = <T extends { value: bigint }>(
...
+  // Every pass either returns or consumes at least one coin from `remaining`, so this bound
+  // is never actually reached; it turns a broken invariant into a thrown error instead of an
+  // infinite loop.
+  for (let pass = 0; pass <= coins.length; pass++) {
+    const covered = inputs.reduce((sum, input) => sum + input.value, 0n);
+    if (fee <= covered) {
+      return { fee, recipeInputs: distributeFeeAcrossInputs(inputs, fee) };
+    }
+
+    const recipe = getBalanceRecipe({
```
kapke: The abstractions `getBalanceRecipe` requires, are simply wrong for the problem at hand, which forces re-creation of the balancing code here. I think the right shape of solution is a `getBalanceRecipeV2`, which gets proper starting point and step functions to converge on the fixed point given an abstract cost model (the overheads are not applicable really).

packages/dust-wallet/src/v1/Transacting.ts
chrispalaskas: Fixed — and thank you, this turned a hang into a malformed transaction, which is worse. `balanceTransactions` now returns `Transaction.fromParts(networkId)` with the state untouched when no inputs are needed; I verified against the ledger that an empty transaction merges into a real one as the identity in both directions (same fee, same intents). The reported fee in that case is `calculateFee` of the merged transactions as they are — no empty intent is ever priced. Covered by the "selects nothing … adds no intent at all" test, which also asserts `intents.size === 0` and state identity.

packages/dust-wallet/src/v1/Transacting.ts
chrispalaskas: Fixed. Selected coins are mapped back from `recipe.inputs` (selection order) via `filterMap`/`findFirst` by nonce, so `distributeFeeAcrossInputs` drains the coins chosen first, as it did before this PR. Your example is now a test verbatim: pool `[big=2e15, small=1e12]`, ascending selection → `[small, big]`, split `[1e12, fee(2) − 1e12]`, and an assertion that no input is charged `0n`. (I did not use `Arr.partition` here because it would lose selection order, which is the property at stake.)

packages/dust-wallet/src/v1/Transacting.ts
chrispalaskas: Fixed. Every ledger call — merge/erase, standalone fees, base fee, each dry run — now runs inside `LedgerOps.ledgerTry`, composed in one `Either.gen`; there is no path on which a ledger throw escapes as a raw exception, so `Effect<bigint, WalletError>` is honoured downstream.

packages/dust-wallet/src/v1/Transacting.ts
chrispalaskas: Agreed, and fixed with your first suggestion: the configured selector is wrapped so that returning `undefined` on a **non-empty** pool surfaces as its own tagged failure (`SelectionRefused`) rather than the balancer's generic `InsufficientFundsError`. Only `SelectionExhausted` — the pool genuinely ran out under the configured order — triggers the largest-first retry; a refusal is reported as insufficient funds with a message saying the configured selection declined the remaining coins. Test: a cap-filter selector on `[allowed, excluded]` gets `InsufficientFundsError` matching `/declined/`, `dryRunFee` never called, the excluded coin never spent. The wrapper throws to escape `getBalanceRecipe` (whose only failure channel is exceptions); it is caught immediately by the `Either.try` around that call and never leaves the module — noted in its JSDoc.

packages/dust-wallet/src/v1/Transacting.ts
chrispalaskas: Fixed as you suggested. `existingDust = Σ feeImbalance(tx, fee_i) + Σ fee_i` is carried through every pass: `deficit = fee(inputs) − existingDust − covered`, and the shortfall the new inputs pay is `fee(inputs) − existingDust`. The seed is now the merged base fee (`calculateFee` of the merged transactions), so the two-transaction case prices consistently. Test: a transaction carrying 60% of its base fee has its one new input pay `fee(1) − 0.6·fee(0)`. One deliberate divergence: the reported `fee` stays the **total** fee of the merged result, not the increment — because the facade's `estimateTransactionFee` documents exactly that ("the total fee for the given transaction plus the fee of the balancing transaction"). Happy to change it if you'd rather the SDK report the increment, but that would be a facade contract change.

packages/dust-wallet/src/v1/Transacting.ts
chrispalaskas: You're right, and the counter is gone. A pass that consumes no coin from `remaining` now fails immediately with `SelectionMadeNoProgress` (`returned` / `remaining` counts), mapped to `TransactingError` with a message saying the selector must pick from the coins it is given. Termination is then structural: every pass shrinks the pool or fails. Test: a selector that ignores its argument and returns a coin of the right shape that is not in the pool → `TransactingError`, `dryRunFee` never called.

packages/dust-wallet/src/v1/test/transacting.test.ts
chrispalaskas: Removed, and folded into the first test as you suggested: that test now spies `dryRunFee` and asserts exactly two prices (pass 1 under-covers, pass 2 converges). Where a title makes a claim about work done, the test now counts it — the retry test asserts two prices in total, the decline/no-progress tests assert zero, and the never-converging model asserts the count is bounded by the pool per attempt.

packages/dust-wallet/src/v1/Transacting.ts
chrispalaskas: (1) Done, with one change to the shape you proposed: rather than probing `dryRunFee([anyCoin]) − dryRunFee([])` up front — which would add one WASM dry run to every call, including the common one-coin case — the marginal fee is **measured** from the first pass (`(fee − baseFee) / inputs`, rounded up) and handed to the balancer as `inputFeeOverhead` for the passes after it. First pass selects on raw value; second pass normally completes it. You're right that this doesn't remove the need for the retry; the stranding test still needs it. I also confirmed your runtime note: `ledgerParams.transactionCostModel.inputFeeOverhead` is `undefined` on the WASM object, so measuring is the only source. (2) Done: `mergeErasedTransactions` proof-erases, merges and picks the segment id once per `computeBalancingRecipe`; `dryRunFee` takes it as an optional trailing parameter so the 7-argument form still works, and `balanceTransactions` uses the same helper instead of its own copy. `spendCoins` still re-spends the cumulative set per pass; with the overhead modelled that is two passes in practice, so I left it.

packages/dust-wallet/src/v1/Transacting.ts
chrispalaskas: Rewritten to the rules: `selectDustInputs` is a tail-recursive `Either.gen` over `const` bindings — no `let`, `for`, `push` or reassignment — returning `Either<Recipe, SelectionFailure | LedgerError>`; failure modes are one `Data.TaggedError` each (`SelectionExhausted`, `SelectionRefused`, `SelectionMadeNoProgress`, `BalancerFailed`), the retry is `Either.orElse` on the exhausted tag only, and the mapping to `WalletError` is a single exhaustive `Match.tag … Match.exhaustive`. `BigIntOps.sumAll` replaces the manual reduce. I used `Array as EArray` rather than `Arr` because that is the alias `tsconfig.base.json`'s language-service config enforces (17 files still use `Arr`; the 4 recent ones use `EArray`). The one remaining `throw` is the refusal wrapper described above, at the boundary of an API whose failure contract is exceptions, caught on the next line up.

packages/dust-wallet/src/v1/Transacting.ts
chrispalaskas: Both fixed: "a surplus of dust" in the JSDoc and the changeset, and the `computeBalancingRecipe` JSDoc now describes current behaviour in two paragraphs — what it selects and what `fee`/`recipeInputs` mean, and when the retry does and does not apply. The history lives in the changeset.

packages/dust-wallet/src/v1/Transacting.ts
chrispalaskas: Agreed on the diagnosis. `getBalanceRecipe` models fee growth as a static per-input overhead, which fits shielded (`ShieldedCostModel`) and is moot for unshielded (`0n/0n`), but dust's fee is a function of the chosen inputs — a `dryRunFee` — so both revisions here had to wrap the balancer in a convergence loop it wasn't built for. Measuring a "marginal" and passing it as `inputFeeOverhead` (the current revision) is the same misfit, just better hidden. And dust uses none of what `CounterOffer` is actually for: multi-token imbalances and change outputs are discarded.

What I'd propose for V2, generic over `TInput extends CoinRecipe`, in `capabilities/src/balancer`:

```ts
balanceToFixedPoint({
  coins,            // the pool
  alreadyCovered,   // starting point: fee already paid by the transactions (for dust: existing spends + registration allowances)
  fee,              // abstract cost model: (inputs: ReadonlyArray<TInput>) => bigint   — dust passes dryRunFee
  coinSelection,    // step: next coin for a deficit
  isCoinEqual,
}): Either<{ fee: bigint; inputs: ReadonlyArray<TInput> }, InsufficientFunds | SelectionRefused | SelectionMadeNoProgress>
```

Fixed point: `fee(inputs) <= alreadyCovered + Σ inputs`. Each step covers only the outstanding deficit from the coins not yet selected, then re-prices; a step that consumes nothing fails, which bounds it by the pool. It is the `selectDustInputs` recursion already in this PR, lifted a package and made generic; no `CounterOffer`, no overheads, no outputs. The dust wallet then reduces to computing `alreadyCovered`, passing `dryRunFee`, and keeping the largest-first retry as its own policy (with the same refusal-vs-exhaustion distinction). Two things fall out for free: the selector-refusal wrapper that has to `throw` to escape `getBalanceRecipe` goes away, since V2 calls the selector itself and returns a tagged error; and the marginal-fee measurement goes away entirely. Shielded and unshielded stay on `getBalanceRecipe` untouched. Roughly: ~120 lines + tests + a `minor` changeset for `capabilities`, and dust-wallet loses about a hundred.

Two questions before I build it: does that interface match what you had in mind (in particular the abstract cost model as `(inputs) => fee` rather than a cost-model object), and do you want it in this PR or as a follow-up on top of it? Happy either way — this PR is a bounded fix that ian has reviewed in detail, so I'd rather not move its goalposts without asking.

kapke: Yes, this is the kind of API I think should work well. 2 things to note:
- effect defines Equivalence API, and it is preferred to a custom check (at the very least, because it provides combinators and docs reference for the invariants)
- it needs to be part of the capabilities/balancing code

Issue #767, opened 2026-09-23
[Bug] WASM prover's default key provider fetches outdated Dust spend keys, so the node rejects every WASM-proved Dust spend (error 170)


jtsang586:
### Version affected

Published `@midnightntwrk/wallet-sdk-prover-client@2.0.0-beta.3` / `@midnightntwrk/wallet-sdk-capabilities@4.0.0-beta.3` (npm `beta`), and `v2` @ `94c69eca`. Introduced 2026-09-04 by the move to `@midnightntwrk/ledger-v9@1.0.0-rc.4` / `proof-server:9.0.0-rc.7` in #715. The ledger-v8 path is not affected (see below).

### Context & setup

- macOS 26.6.2 (Darwin 25.6.0, arm64), Node v22.16.0, Docker 29.8.0
- Local undeployed stack from `infra/compose/docker-compose-dynamic.yml`: `midnight-node:2.1.0-beta.1`, `proof-server:9.0.0-rc.7`, `indexer-standalone:4.4.0-rc.2-25da0487`
- Chain is ledger-v9 from genesis (`forks: ProtocolVersion.V9NativeForkSchedule`)
- In-process proving with the default key material, via either `makeV9WasmProvingService()` or the facade config `provers: { v9: { kind: 'wasm' } }` (both reach `makeDefaultKeyMaterialProvider()` through `versionedProving.ts` `wasmConfigurationOf` → `makeV9WasmProvingServiceEffect({})`)

**Cause (verified).** `makeDefaultKeyMaterialProvider` in `packages/prover-client/src/effect/WasmProver.ts` (L228–231) hard-codes the key source and the key generation:

```ts
const s3 = 'https://midnight-s3-fileshare-dev-eu-west-1.s3.eu-west-1.amazonaws.com';
const ver = config?.circuits ?? 9;
// 'midnight/dust/spend' → `dust/${ver}/spend`
```

The `9` is a circuit-key generation, not the ledger major. Ledger-v9 moved the Dust spend circuit to generation **10**:

- `proof-server:9.0.0-rc.3` embeds `dust/9/spend`; `9.0.0-rc.7` embeds `dust/10/spend`, sourced from `https://srs.midnight.network/` (`MIDNIGHT_PARAM_SOURCE` default), with expected-hash checks.
[...]

PR #768, opened 2026-09-23
fix(prover-client): prove each ledger version with its own key material

agronmurtezi:
# Description

The in-process (WASM) prover proved ledger-v9 Dust spends with outdated keys, so the node rejected every such
transaction with `Custom error: 170`. It read circuit generation 9 from a dev bucket for both ledger versions, but
ledger-v9 rc.4 moved the Dust spend circuit to generation 10. Refs #767.

# Proposed Solution

- One key provider per ledger version: `makeV9KeyMaterialProvider()` (generation 10) and `makeV8KeyMaterialProvider()`
  (generation 9). `makeDefaultKeyMaterialProvider()` stays and returns the v9 one.
- Keys come from `https://srs.midnight.network/`, the host the ledger itself uses, and every file is checked against the
  hash the ledger release publishes. A mismatch or an error response is rejected instead of used.
- `{ source }` lets you host the files yourself, e.g. for a browser app, since neither host sends CORS headers.
- The v8 and v9 in-process backends in capabilities each use their own provider.

- **Breaking (prover-client beta):** the `circuits` option is removed; pick the ledger version instead.
- A test fails when the installed ledger version changes, so the next ledger bump forces a re-check of the pinned hashes
  (the step #715 missed).
- `facadeTransfer`'s shielded transfer test is now `@smoke`. Only a node checks Dust proofs, so this runs one on every PR
  (about 2 minutes).
- The `hard-fork-support` changeset no longer mentions `circuits`.

# Testing

- Local undeployed stack: on `v2` the shielded transfer test fails with error 170. With this PR it passes, as do the other
  two in-process transfer tests. The two "arbitrary transaction" tests in that file still fail on an unrelated e2e helper
  bug (it stamps protocol version 0).
- `yarn test --filter=@midnightntwrk/wallet-sdk-prover-client` (needs network; one suite uses Docker) and
  `yarn test:unit --filter=@midnightntwrk/wallet-sdk-capabilities`.


jtsang586 (approved):
All previous failing tests now pass. Lgtm

